PR #8, opened 2017-09-01
Adding boundary tag to nodes on ways that span boundaries

matthieun:
When a way is not assigned any country code (it is not within a country boundary) but it is still connected to ways that are inside the boundary (long bridges, ferry routes for example), the end nodes of those ways were not assigned any `synthetic_boundary_node` tag. This PR adds the tag to those nodes.
It also refactors the function that decides when an OSM node becomes an Atlas `Point`. That allows to take the new tags into account, and to avoid building more `Point`s than necessary.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
@@ -25,6 +25,7 @@ import org.openstreetmap.atlas.geography.atlas.items.Line;
 import org.openstreetmap.atlas.geography.atlas.items.Point;
 import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas;
 import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlasBuilder;
+import org.openstreetmap.atlas.geography.atlas.pbf.converters.TagMapToTagCollectionConverter;
 import org.openstreetmap.atlas.geography.atlas.pbf.slicing.CountrySlicingProcessor;
 import org.openstreetmap.atlas.geography.atlas.pbf.slicing.WaySectionProcessor;
 import org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier.PaddingIdentifierFactory;
@@ -39,7 +40,9 @@ import org.openstreetmap.atlas.tags.LastEditUserIdentifierTag;
 import org.openstreetmap.atlas.tags.LastEditUserNameTag;
 import org.openstreetmap.atlas.tags.LastEditVersionTag;
 import org.openstreetmap.atlas.tags.RouteTag;
+import org.openstreetmap.atlas.tags.SyntheticBoundaryNodeTag;
 import org.openstreetmap.atlas.tags.Taggable;
+import org.openstreetmap.atlas.utilities.collections.Maps;
 import org.openstreetmap.osmosis.core.container.v0_6.EntityContainer;
 import org.openstreetmap.osmosis.core.domain.v0_6.Bound;
 import org.openstreetmap.osmosis.core.domain.v0_6.Entity;
@@ -72,8 +75,8 @@ public class OsmPbfProcessor implements Sink
     private static final Logger logger = LoggerFactory.getLogger(OsmPbfProcessor.class);
 
     private static final JtsMultiPolygonToMultiPolygonConverter JTS_MULTI_POLYGON_TO_MULTI_POLYGON_CONVERTER = new JtsMultiPolygonToMultiPolygonConverter();
-
     private static final int MAXIMUM_NETWORK_EXTENSION = 100;
+    private static final TagMapToTagCollectionConverter TAG_MAP_TO_TAG_COLLECTION_CONVERTER = new TagMapToTagCollectionConverter();
 
     private final PackedAtlasBuilder builder;
     private final AtlasLoadingOption loadingOption;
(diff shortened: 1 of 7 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderTest.java
```diff
@@ -99,4 +112,22 @@ public void testNoFilter()
         logger.info("{}", atlas);
         Assert.assertEquals(3, atlas.numberOfLines());
     }
+
+    @Test
+    public void testOutsideWayBoundaryNodes()
```
MikeGost: Nice test!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
```diff
@@ -694,11 +699,39 @@ private void keepOutsideWaysThatAreConnected()
                 {
                     if (!this.store.containsWay(way.getId()))
                     {
+                        final List<WayNode> wayNodes = way.getWayNodes();
                         for (final WayNode wayNode : way.getWayNodes())
```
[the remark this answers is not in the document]
matthieun: Good catch, thanks!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
```diff
@@ -694,11 +699,39 @@ private void keepOutsideWaysThatAreConnected()
...
                         {
                             final long identifier = wayNode.getNodeId();
                             if (this.store.containsNodeAtEndOfEdges(identifier))
                             {
+                                final Long startNodeIdentifier = wayNodes.get(0).getNodeId();
```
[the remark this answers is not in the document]
matthieun: This is within the for loop that loops through all the wayNodes, so if it is empty, it would not reach the body of the for loop.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
```diff
@@ -502,9 +503,25 @@ public boolean isAtlasPoint(final long identifier)
 
     public boolean isAtlasPoint(final Node node)
     {
-        // Each node will have a last edit time tag
-        return node.getTags().size() > AtlasTag.TAGS_FROM_OSM.size()
-                || containsNodeInRelations(node.getId());
+        if (containsNodeInRelations(node.getId()))
+        {
+            return true;
```
mertemin: So if an OSM node is part of a relation and it is at the end of an edge, then it would be have one node counterpart in Atlas and one point counterpart. Is that intentional?

mertemin: Plus, what is the reason we make an OSM node Atlas point because it is part of a relation?

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
```diff
@@ -502,9 +503,25 @@ public boolean isAtlasPoint(final long identifier)
...
+        }
+        // Tags from OSM are the tags that all the nodes will have
+        if (node.getTags().size() > AtlasTag.TAGS_FROM_OSM.size())
```
mertemin: This logic behind needs some explanation and seems easy to break. I'd suggest adding some comments and explaining why it makes a node an Atlas point.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
matthieun: I am not sure why this was designed this way. I figured for this PR I would not touch that part of the logic, but you are correct this can be questionable... Created https://github.com/osmlab/atlas/issues/10 to track it.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
matthieun: Created https://github.com/osmlab/atlas/issues/11 as well. Again for this PR I did not want to change the logic here, but that deserves a refactor.
I will add more comments as well in the mean time.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
mertemin: Cool. It can definitely be addressed later.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
robberk: The design was any OSM node with artificial tag will be promoted as an Atlas Point.
So here it is checking if the OSM node has artificial tag other than those default tags from OSM (like LastEditTimeTag etc.).

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
robberk: I don't remember why we need this check, it doesn't seem correct. Agree to try to remove this later. Thanks!

PR #15, opened 2017-09-27
DynamicAtlas to load from multiple initial shards

matthieun:
`DynamicAtlas` had the option to start only with one single shard. This PR allows loading from a set of multiple initial shards, or from a polygon. That gives much more flexibility to `DynamicAtlas`.

Also includes some extensions to `MultiPolygon` for the `overlaps` and `intersects` methods.


matthieun:
I realize this will need a bit more testing on the `DynamicAtlas` part before it gets merged...

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/policy/DynamicAtlasPolicy.java
```diff
@@ -59,9 +95,16 @@ public DynamicAtlasPolicy(final Function<Shard, Optional<Atlas>> atlasFetcher,
         };
     }
 
-    public Shard getInitialShard()
+    public Set<Shard> getInitialShards()
+    {
+        return this.initialShards;
+    }
+
+    public MultiPolygon getInitialShardsBounds()
```
MikeGost: Nit: getInitialShardsBounds --> getInitialShardBounds. (no s)

src/test/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasMultipleInitialShardTest.java
```diff
@@ -0,0 +1,102 @@
...
+import org.openstreetmap.atlas.geography.atlas.Atlas;
+import org.openstreetmap.atlas.geography.atlas.dynamic.policy.DynamicAtlasPolicy;
+import org.openstreetmap.atlas.geography.sharding.Shard;
+import org.openstreetmap.atlas.geography.sharding.SlippyTile;
+import org.openstreetmap.atlas.geography.sharding.SlippyTileSharding;
+
+/**
+ * @author matthieun
+ */
+public class DynamicAtlasMultipleInitialShardTest
```
MikeGost: Nice tests!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/policy/DynamicAtlasPolicy.java
MikeGost: Nit: getInitialShardsBounds --> getInitialShardBounds (no s)

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -289,6 +289,32 @@ public int nextVertexIndex(final int currentVertexIndex)
         }
     }
 
+    public boolean overlaps(final MultiPolygon multiPolygon)
```
MikeGost: Might be worthwhile adding a test for this method in `PolygonTest`.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/policy/DynamicAtlasPolicy.java
matthieun: I added the "s" there because there are multiple shards. Let me know if that reads too weird and I'll remove.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
matthieun: It is tested in the MultiPolygonTest but that is not obvious. I will move it to that test instead. Thanks!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/policy/DynamicAtlasPolicy.java
MikeGost: Sounds better without it. Your call thought :)

Issue #17, opened 2017-09-28
Osmosis PBF to Atlas loader refactor


matthieun:
The current system to load Osmosis PBF files into the Atlas format (and perform boundary slicing and way sectioning) requires multiple passes through the PBF file, which could be faster.

It would be nice to load the PBF files once only, and then do the slicing/sectioning using the Atlas APIs directly, which would provide spatial and geometry access.

matthieun:
This one will help for way sectioning: https://github.com/osmlab/atlas/pull/25

MikeGost:
This will be done in 3 parts:
1. Create Raw Atlas
2. Country Slice Raw Atlas
3. Way-Section Country-Sliced Raw Atlas in a deterministic way using DynamicAtlas functionality.

MikeGost:
@matthieun - closing as completed.

PR #23, opened 2017-10-16
Initial Raw Atlas Implementation

MikeGost:
This is the first portion of the Osmosis PBF to Atlas refactor [task](https://github.com/osmlab/atlas/issues/17). 

**A summary of changes:**
- Created `OsmPbfCounter` (Osmosis `Sink` implementation) - responsible for reading the PBF file and generating a count of all `Point`s, `Line`s and `Relation`s found.
- Created `OsmPbfReader` (Osmosis `Sink` implementation) - responsible for reading the PBF file creating corresponding Atlas Entities (`Point`, `Line`, `Relation`)
- Created `RawAtlasGenerator` - responsible for Raw `Atlas` creation, calls the two new Sink implementations, creates the `AtlasSize` and `AtlasMetaData` and builds the raw `Atlas`.
- Added a raw `Atlas` functionality test to verify successful generation and expected feature count
- Added another test to validate parity between the old generated Atlas and the new raw Atlas, that compares counts of Entities with the same filters to ensure nothing got lost in translation.

**Justification to loop over PBF twice:**
Ideally, I wanted to loop over the PBF just once. However, because we don't know how many items are in a PBF file before processing it, we can't provide a `AtlasSize` to the `PackedAtlasBuilder`. This causes us to continuously resize the underlying `PackedAtlas` arrays used to track identifiers and locations. For a single PBF shard, I'd typically see around 6-7 re-sizes, which would dramatically increase the runtime. Instead, I created the `OsmPbfCounter`, which runs through the PBF file once, gets the counts of all desired entities and uses those counts to create an appropriate `AtlasSize`. Here are some logs with and without setting the size:

Without:
```
2017-10-16 13:48:14 INFO  [main] RawAtlasGenerator:163 - Read PBF in 2.716 seconds, preparing to build Raw Atlas
```

With:
```
2017-10-16 13:47:24 INFO  [main] RawAtlasGenerator:201 - Counted PBF entities in 202 milliseconds
2017-10-16 13:47:24 INFO  [main] RawAtlasGenerator:163 - Read PBF in 691 milliseconds, preparing to build Raw Atlas
```

You'll notice that the time it took to read the PBF _without_ proper AtlasSize is ~3x as long as it is to count the entities and then read the PBF _with_ proper AtlasSize. Feel free to try it out yourself by uncommenting the line that sets `AtlasSize` in RawAtlasGenerator before buildings (should be line 97).

**Next steps**:
1. ~~Addressing some of the TO-DOs - namely around keeping statistics for dropped, skipped and added items.~~ (see commit [here](https://github.com/osmlab/atlas/pull/23/commits/75a3db0a6d909bc13e64c611af4b1228d1efa3f1))
2. Update Atlas Generation to use either flow, specified by configuration: a. existing flow using `OsmPbfLoader` b. The raw `Atlas` flow from above. This will make it easier to test and to upgrade to the refactored version


src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/OsmPbfReader.java
```diff
@@ -0,0 +1,420 @@
...
+     *
+     * @param way
+     *            The {@link Way} for which to construct the {@link PolyLine}
+     * @return the constructed {@link PolyLine}
+     */
+    private PolyLine constructWayPolyLine(final Way way)
+    {
+        final List<WayNode> wayNodes = way.getWayNodes();
+        final List<Location> wayLocations = new ArrayList<>();
+        wayNodes.forEach(node -> wayLocations.add(getNodeLocation(node.getNodeId())));
```
matthieun: In the rare case of a malformed pbf file, is it possible that `getNodeLocation` returns nothing here?

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/OsmPbfReader.java
```diff
@@ -0,0 +1,420 @@
...
+            final RelationBean bean = constructRelationBean(relation);
+            if (!bean.isEmpty())
+            {
+                this.builder.addRelation(relation.getId(), relation.getId(), bean,
+                        populateEntityTags(relation));
+            }
+            else
+            {
+                logger.debug(
+                        "Empty Relation {} cannot be added to the Atlas. We're either filtering out the members that make up the Relation or none of the members are present in this PBF shard.",
```
matthieun: Maybe break that long string into two or three lines?

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/OsmPbfReader.java
```diff
@@ -0,0 +1,420 @@
...
+     * encounter a {@link Relation} that's over {@link #MAXIMUM_RELATION_DEPTH} levels deep, then we
+     * throw a {@link CoreException} and stop building the {@link Atlas}.
+     */
+    private void processStagedRelations()
+    {
+        int relationDepth = 0;
+        List<Relation> stagedRelations = this.stagedRelations;
+        while (!stagedRelations.isEmpty() && relationDepth < MAXIMUM_RELATION_DEPTH)
+        {
+
```
matthieun: Whitespace

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/RawAtlasGenerator.java
```diff
@@ -0,0 +1,229 @@
...
+
+    /**
+     * Constructor.
+     *
+     * @param resource
+     *            The OSM PBF {@link Resource} to use
+     */
+    public RawAtlasGenerator(final Resource resource)
+    {
+        // TODO : Update AtlasLoadingOption to remove country-slicing/way-sectioning configurations
```
matthieun: 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/OsmPbfReader.java
MikeGost: Good call!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/loading/OsmPbfReader.java
MikeGost: It is possible - either we filtered it out or the PBF is malformed. If that's the case, we throw a `CoreException` on line 278 below. Let me add more context to the Exception message there.

MikeGost:
Updated the PR to do the following:
1. Handle staged relations in a cleaner way. Instead of having a loop of 500, we use the previous staged relation size change to see if we made any progress in adding any relations. If there are two successful tries to process staged relations without progress, any further tries won't result in added relations. At this point, try to add the relations to the Atlas in the state they are in. It's possible that the Relations will be missing other Relation members, because they will be in another PBF shard. The assumption we're making is if we try to combine this shard with a neighboring shard, which shares the same relation, their member lists will be combined to find the missing members. Otherwise, if the Relation member list is empty, we won't add the relation.
2. Adding Relations to the Atlas is now a private method. 
3. Added tests around Relations of Relations and Relations with missing members. This significantly increased the code coverage (see graphic below)

screen shot 2017-10-19 at 1 43 08 pm

MikeGost:
One more change, to test for corrupt/invalid PBF input files, more tests around Relations and fixed a bug when looking up Point Relation members. Here are the latest test coverage metrics.

screen shot 2017-10-20 at 3 00 16 pm

PR #25, opened 2017-10-30
DynamicAtlas expansion according to a predicate

matthieun:
This change allows for expanding of a `DynamicAtlas` to be triggered only by the loading of some `AtlasEntity`(ies) that follow a specific `Predicate`. That brings in more flexibility to `DynamicAtlas` if a user is interested only in expanding along one feature type for example.

I added 2 new tests for the new behavior, using the same `DynamicAtlasTestRule `.

Also included: now `DynamicAtlas` can expand on `Relation`s "aggressively". Because Relations might have incomplete member lists in any given shard, this option looks at all the neighboring shards to find if any of those shards contain the same relation and new members. In that case, it does add the new shard to the expansion list.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
@@ -751,6 +751,7 @@ public class DynamicAtlas extends BareAtlas
     {
         final RelationMemberList knownMembers = relation.members();
         boolean result = true;
+        boolean loop = false;
         for (final RelationMember member : knownMembers)
         {
             final AtlasEntity entity = member.getEntity();
@@ -783,6 +784,7 @@ public class DynamicAtlas extends BareAtlas
                     logger.error(
                             "Skipping! Unable to expand on relation which has a loop: {}. Parent tree: {}",
                             relation, parentRelationIdentifierTree);
+                    loop = true;
                     result = true;
                 }
                 else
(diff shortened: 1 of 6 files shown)


MikeGost:
Other than testing suggestions, code looks great!

matthieun:
@MikeGost Thanks good ideas. Let me close while I add some testing to it.

src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
```diff
@@ -385,30 +341,6 @@ public String toString()
...
-     * Add the siblings and parent.
-     *
-     * @param candidates
-     *            The candidate tiles
-     * @param visitedTiles
-     *            The tiles already visited
-     * @param targetTile
-     *            The target tile
-     */
-    private void addNextZoomLevelTilesToCandidateList(final Queue<SlippyTile> candidates,
```
MikeGost: I'd suggest keeping the old method (zoom-level irrespective) around in case it can be useful, perhaps renaming it as `getNeighborsForAllZoomLevels`.

matthieun: ok, let me do that

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -801,6 +805,47 @@ else if (entity instanceof Relation)
...
+            this.loadedShards.keySet().forEach(
+                    shard -> this.sharding.neighbors(shard).forEach(onlyNeighboringShards::add));
+            onlyNeighboringShards.removeAll(this.loadedShards.keySet());
+            // For each of those shards, load the Atlas individually and find the relation and its
+            // members if it is there too.
+            final Set<Shard> neighboringShardsContainingRelation = new HashSet<>();
+            onlyNeighboringShards
+                    .forEach(shard -> this.policy.getAtlasFetcher().apply(shard).ifPresent(atlas ->
+                    {
+                        final Relation newRelation = atlas.relation(relation.getIdentifier());
```
mertemin: Can't you implement this already right now with the current `DynamicAtlasPolicy` definition? It seems like code block from line 821-840 could be implemented into `DynamicAtlasPolicy`'s `atlasFetcher` parameter.

matthieun: So this method here loads the very neighboring shards and looks at the relation members to see if any completely new members are present here or not. If yes, it adds the new shards to the list.

The previous code could load and expand on relations, but only if some members of the relation were partially outside of the current set of shards.

If you think I have misunderstood your question, can you send a link to the code that you think can replace this?

MikeGost: @mertemin - please file a github issue if you see a refactor opportunity and we can address it. Otherwise, merging to make use of this functionality.

matthieun:
Relates to #17

PR #31, opened 2017-11-08
Raw Atlas Country Slicing for Ways and Points

MikeGost:
This is part 2 of the PBF Ingest [issue](https://github.com/osmlab/atlas/issues/17). A few changes here:

1. Cleaning up and removing existing country-slicing code. One example is packaging the country code and nearest neighbor values into a simple POJO class instead of relying on a Tuple to keep track of them.
2. Updating Raw Atlas generation to remove shape points that were used by filtered Ways and to pad identifiers using the existing 6 digit logic (first 3 for country slicing, next 3 for way-sectioning)
3. Created `RawAtlasCountrySlicer`, which takes a Raw Atlas as input and country-slices it. In doing so, introduced the notion of `TemporaryEntity`, which is the bare minimum information required to create an Atlas Entity. These are created as a result of country slicing creating new Points and new Lines when cutting entities along borders. There is also the definition of a `RawAtlasChangeSet`, which keeps track of added, modified and deleted items as a result of country slicing and is the applied by the `RawAtlasChangeSetBuilder` to build the finished product.
4. Corresponding unit tests. Some of the scenarios covered are Lines fully enclosed in a country, Lines spanning multiple countries with and without a shape point on the border, Closed Lines (Areas) spanning a single and multiple countries, various testing for Synthetic tags such as nearest neighbor and boundary tags node tagging.


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,502 @@
...
+                    AbstractIdentifierFactory.IDENTIFIER_SCALE, coordinate);
+            return Optional.empty();
+        }
+        else
+        {
+            // Add the synthetic boundary node tags
+            pointTags.put(SyntheticBoundaryNodeTag.KEY, SyntheticBoundaryNodeTag.YES.toString());
+
+            return Optional.of(new TemporaryPoint(pointIdentifierFactory.nextIdentifier(),
+                    new Location(Latitude.degrees(coordinate.y), Longitude.degrees(coordinate.x)),
```
matthieun: Here you should use the `JtsLocationConverter` to avoid code duplication.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,502 @@
...
+        // If we used nearest neighbor logic to determine the country code, add a tag
+        // to indicate this
+        if (countryDetails.usingNearestNeighbor())
+        {
+            tags.put(SyntheticNearestNeighborCountryCodeTag.KEY,
+                    SyntheticNearestNeighborCountryCodeTag.YES.toString());
+        }
+
+        // For any border nodes, add the existing tag
+        if (fromRawAtlas && countryDetails.inMultipleCountries())
```
matthieun: Here does `fromRawAtlas` mean that the `Point` that generated this `Location` is already in the raw Atlas?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,502 @@
...
+     * Slices all the {@link Line}s in the given raw Atlas.
+     */
+    private void sliceLines()
+    {
+        this.rawAtlas.lines().forEach(this::sliceLine);
+    }
+
+    /**
+     * Updates all stand-alone {@link Point}s that haven't been assigned a country code after
+     * way-slicing.
```
matthieun: Maybe expand the comment here to explain that it means they fell outside of any country boundary?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/TemporaryEntity.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @Override
+    public boolean equals(final Object other)
+    {
+        if (this == other)
+        {
+            return true;
+        }
+        if (other != null && this.getClass() == other.getClass())
```
matthieun: Why not use `instanceof` here?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/TemporaryEntity.java
```diff
@@ -0,0 +1,81 @@
...
+
+import org.apache.commons.lang3.builder.HashCodeBuilder;
+
+/**
+ * Base class for a temporary object, that keeps track of the bare minimum information required to
+ * create some Atlas entity.
+ *
+ * @author mgostintsev
+ */
+public class TemporaryEntity implements Serializable
```
matthieun: Might be nice to make this one abstract, and to add an abstract method `toString` so the sub classes have to implement it.

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/LineAndPointSlicingTestRule.java
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ * {@link RawAtlasLineAndPointSlicingTest} test data.
+ *
+ * @author mgostintsev
+ */
+public class LineAndPointSlicingTestRule extends CoreTestRule
+{
+    // Single crossing across boundary
+    private static final String LIBERIA_END = "7.2,-8.4";
```
matthieun: Nice way of making all this very readable!

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasLineAndPointSlicingTest.java
```diff
@@ -0,0 +1,453 @@
...
+    }
+
+    @Rule
+    public LineAndPointSlicingTestRule setup = new LineAndPointSlicingTestRule();
+
+    @BeforeClass
+    public static void setup()
+    {
+        COUNTRY_BOUNDARY_MAP = new CountryBoundaryMap(
+                new InputStreamResource(RawAtlasLineAndPointSlicingTest.class
```
matthieun: When creating an `InputStreamResource` it is usually better to provide a supplier here instead of the stream directly, so the resource can be reused. It might not impact here, but just in case.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfReader.java
```diff
@@ -446,8 +501,39 @@ private void processStagedRelations()
...
+            this.statistics.recordCreatedLine();
+        }
+    }
+
+    /**
+     * Because of how an Atlas is constructed, we add all PBF {@link Node}s as {@link Point}s.
+     * However, if we filter out an PBF {@link Way}, we want to filter out all the {@link Node}s
+     * that make up that way. We are going to keep track of all filtered PBF {@link Node}/Atlas
+     * {@link Point} identifiers, and use subAtlas functionality to filter them out after the Atlas
+     * is built. For now, ignore Relations.
```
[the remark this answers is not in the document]
MikeGost: Good callout, I'll add a TODO to come back to it. The intention was to fill in Relation logic once they're country-sliced.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -163,7 +163,20 @@ private Atlas buildRawAtlas()
         logger.info("Read PBF in {}, preparing to build Raw Atlas", parseTime.elapsedSince());
 
         final Time buildTime = Time.now();
-        final Atlas atlas = this.builder.get();
+        final Atlas atlas;
```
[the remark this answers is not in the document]
MikeGost: Good idea, will go with your option.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasChangeSet.java
```diff
@@ -0,0 +1,115 @@
...
+    {
+        this.createdPoints = new ArrayList<>();
+        this.updatedPointTags = new HashMap<>();
+        this.createdLines = new ArrayList<>();
+        this.updatedLineTags = new HashMap<>();
+        this.deletedLines = new HashSet<>();
+        this.deletedToCreatedLineMapping = new HashMap<>();
+    }
+
+    public void clear()
```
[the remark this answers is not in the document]
MikeGost: No intention to re-use, was trying to get rid of memory overhead. Playing around with the idea of getting rid of `clear()` and just setting the entire change set to `null` once it's been applied.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasChangeSetBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+                if (this.changeSet.getUpdatedLineTags().containsKey(lineIdentifier))
+                {
+                    final Line originalLine = this.original.line(lineIdentifier);
+                    final Map<String, String> updatedTags = originalLine.getTags();
+                    updatedTags.putAll(this.changeSet.getUpdatedLineTags().get(lineIdentifier));
+                    this.builder.addLine(lineIdentifier, originalLine.asPolyLine(), updatedTags);
+                }
+                else
+                {
+                    logger.error(
```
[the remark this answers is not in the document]
MikeGost: Good idea.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasChangeSetBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+            if (this.changeSet.getUpdatedPointTags().containsKey(pointIdentifier))
+            {
+                final Point originalPoint = this.original.point(pointIdentifier);
+                final Map<String, String> updatedTags = originalPoint.getTags();
+                updatedTags.putAll(this.changeSet.getUpdatedPointTags().get(pointIdentifier));
+                this.builder.addPoint(pointIdentifier, originalPoint.getLocation(), updatedTags);
+            }
+            else
+            {
+                logger.error(
```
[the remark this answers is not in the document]
MikeGost: Done

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasChangeSetBuilder.java
```diff
@@ -0,0 +1,271 @@
...
+                        // Do one last sanity check to make sure the member we're adding
+                        // actually exists
+                        if (this.builder.peek().line(addedLineIdentifier) != null)
+                        {
+                            bean.addItem(addedLineIdentifier, member.getRole(),
+                                    member.getEntity().getType());
+                        }
+                        else
+                        {
+                            logger.error(
```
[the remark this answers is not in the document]
MikeGost: You make a good point. It's better to fail loudly here than log and silently have a corrupt Atlas.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,502 @@
...
+        slicePoints();
+
+        // Apply changes and rebuild the Atlas with the changes before slicing Relations
+        final RawAtlasChangeSetBuilder changeBuilder = new RawAtlasChangeSetBuilder(this.rawAtlas,
+                this.afterSlicingLinesAndPoints);
+        final Atlas withSlicedWaysAndPoints = changeBuilder.applyChanges();
+        this.rawAtlas = withSlicedWaysAndPoints;
+
+        // Clear for GC
+        this.afterSlicingLinesAndPoints.clear();
```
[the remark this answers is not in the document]
MikeGost: Setting to `null` instead.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Good call!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/TemporaryEntity.java
MikeGost: Good idea!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,502 @@
...
+    private Optional<TemporaryPoint> createNewPoint(final Coordinate coordinate,
+            final CountrySlicingIdentifierFactory pointIdentifierFactory,
+            final Map<String, String> pointTags)
+    {
+        if (!pointIdentifierFactory.hasMore())
+        {
+            logger.error(
+                    "Country Slicing exceeded maximum number of supported new points at Coordinate {}",
+                    AbstractIdentifierFactory.IDENTIFIER_SCALE, coordinate);
+            return Optional.empty();
```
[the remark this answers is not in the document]
MikeGost: Good callout! Thought about this for a while and went back and forth on what to do. If we throw an Exception, we'll stop country-slicing for this RawAtlas and essentially cease Atlas creation until either the code is changed (refactor identifiers, filter types of ways) or OSM data changes (we don't want this). Two feature types that are prone to this are boundaries and coastlines. 

I'd like to avoid this and handle it more gracefully. Some ideas: 1. Throw a specific type of error, log and drop this way from the raw Atlas. 2. Cease slicing once we reach the maximum identifier and extend the last segment to be a single long segment until the end. 

Option 1 seems cleaner and Option 2 is error-prone and confusing, since you will have a (potentially long) segment spanning 2 (or more) countries with a single country code. 

Thoughts?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: The logic is that - if we're creating a new `Point`, it's only going to happen at the boundary, since that's where slicing happens. We have two ways of retrieving `Point`s - 1. from the original un-sliced Raw Atlas 2. from the `newPointCoordinates` map which maps a `Coordinate` to any `TemporaryPoint` identifier.

We first check to see if there is already a new `Point` for this `Coordinate` in the map (to avoid creating duplicate points). If there's not, we check the raw Atlas to see if there is already a `Point` at this location. If the atlas has a Point here, we know to add the `existing` boundary tag value, as opposed to the `yes` value. This has some corner cases, such as an existing node at the same coordinate that's not part of that line, but have yet to see this when testing.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Good point

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: I'll add this as a comment

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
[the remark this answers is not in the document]
matthieun: I agree, option 1 is better! Thanks!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
matthieun: Not sure this comment will help, but when two `Node`s are at the exact same location, the Atlas connects everything to one of the 2. The other node is still there, but is connected to nothing.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,544 @@
...
+        slicePoints();
+
+        // Apply changes and rebuild the Atlas with the changes before slicing Relations
+        final RawAtlasChangeSetBuilder changeBuilder = new RawAtlasChangeSetBuilder(this.rawAtlas,
+                this.afterSlicingLinesAndPoints);
+        final Atlas withSlicedWaysAndPoints = changeBuilder.applyChanges();
+        this.rawAtlas = withSlicedWaysAndPoints;
+
+        // Clear for GC
+        this.afterSlicingLinesAndPoints = null;
```
matthieun: Still not sure if this is necessary... When the `RawAtlasCountrySlicer` is dereferenced, this object will be cleaned automatically too.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,535 @@
...
+     * only going to happen at the boundary, since that's where slicing happens. We have two ways of
+     * retrieving Points - 1. from the original un-sliced Raw Atlas 2. from the newPointCoordinates
+     * map which maps a {@link Coordinate} to a {@link TemporaryPoint} identifier. We first check to
+     * see if there is already a {@link TemporaryPoint} for this {@link Coordinate} in the map (to
+     * avoid creating duplicate points). If there's not, we check the raw Atlas to see if there is
+     * already a {@link Point} at this {@link Location}. If the atlas has a {@link Point} there, we
+     * know to assign the {@link SyntheticBoundaryNodeTag#EXISTING} value, as this {@link Point} is
+     * already part of the raw OSM data. If the raw Atlas doesn't contain a {@link Point} at this
+     * {@link Location}, then we create a new {@link TemporaryPoint} and assign the
+     * {@link SyntheticBoundaryNodeTag#YES} value.
```
matthieun: 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -0,0 +1,535 @@
...
+     * @param relation
+     */
+    private void sliceRelation(final Relation relation)
+    {
+        this.statistics.recordProcessedRelation();
+
+        if (Validators.isOfType(relation, RelationTypeTag.class, RelationTypeTag.BOUNDARY,
+                RelationTypeTag.MULTIPOLYGON))
+        {
+            // Process geometry based relations
```
matthieun: This will come in another PR?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/TemporaryEntity.java
```diff
@@ -0,0 +1,80 @@
...
+        this.tags = tags;
+    }
+
+    @Override
+    public boolean equals(final Object other)
+    {
+        if (other instanceof TemporaryEntity)
+        {
+            final TemporaryEntity that = (TemporaryEntity) other;
+            return this.getIdentifier() == that.getIdentifier();
```
matthieun: There is a possible issue here that I hadn't seen before. It is possible in OSM that a line and a point have the same ID but are different objects.
I would change this one to abstract and implement in the sub-classes where the `instanceof` will take care of it.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Correct, coming in the next PR, currently in progress.

MikeGost:
I'll update the tests to rely on a single PBF file instead of 3 separate ones. The sizes are around 4gb each, with exception of one which is .3GB

PR #32, opened 2017-11-08
Country slicing: force-slice ferries and piers

matthieun:
Ferries need to be country sliced properly, even when jumping from one country to no country (international waters). Current behavior is to not slice any feature that is crossing such a boundary. Also, in some cases, when the country boundary is admin_level 3 and hugs the coastline, we also want the piers connected to ferry lines to be sliced, even though they go from a country to no country.

That PR brings multiple things:
 - Ferries are always country-sliced, even when going from country to no country
 - Piers are always country-sliced, even when going from country to no country
 - All other features going from country to no country are not sliced, as before.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java
@@ -22,8 +22,12 @@ import org.openstreetmap.atlas.geography.atlas.pbf.store.TagMap;
 import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMap;
 import org.openstreetmap.atlas.geography.converters.jts.JtsUtility;
 import org.openstreetmap.atlas.tags.ISOCountryTag;
+import org.openstreetmap.atlas.tags.ManMadeTag;
+import org.openstreetmap.atlas.tags.RouteTag;
 import org.openstreetmap.atlas.tags.SyntheticBoundaryNodeTag;
 import org.openstreetmap.atlas.tags.SyntheticNearestNeighborCountryCodeTag;
+import org.openstreetmap.atlas.tags.Taggable;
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java
```diff
@@ -341,6 +345,13 @@ else if (piece1.getStartNode().equals(piece2.getEndNode()))
         return results;
     }
 
+    private boolean canSkipSlicingIfSingleCountry(final Way way)
```
MikeGost: Nit: maybe a quick comment saying we can skip slicing for a single country unless the way is a Pier or a Ferry, since both of these features can extend into the surrounding water.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -900,7 +903,7 @@ public int size()
 
         // Performance improvement, if only one polygon returned no need to do any further
         // evaluation.
```
MikeGost: Should we update the comments here and below?

src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessorTest.java
```diff
@@ -117,6 +133,38 @@ public void testFilterBasedOnBound()
...
+
+        final MultiPolygon bound = boundaryMap.countryBoundary("CIV").get(0).getBoundary();
+        final CountrySlicingProcessor processor = new CountrySlicingProcessor(this.store,
+                boundaryMap, bound, null);
+        processor.run();
+        Assert.assertEquals("should be 2 piece after slicing", 2, this.store.wayCount());
+    }
+
+    @Test
+    public void testForceSlicingForPiers()
```
MikeGost: Nice tests!

PR #34, opened 2017-11-15
Remove forced slicing due to decreased performance

matthieun:
This is rolling back #32 as the performance is negatively impacted due to too many slicing operations.


MikeGost:
Thanks for the quick revert @matthieun. Merging to get this into the other PR #31

PR #38, opened 2017-12-08
Raw Atlas Relation Slicing

MikeGost:
This fills in the missing piece ofpart 2 of the PBF Ingest [issue](https://github.com/osmlab/atlas/issues/17). A few changes here:

1. Separating out and cleaning up the definition of a change set. The initial slicing of lines/points will have on change set and relation slicing will yield a separate one. Each change set will have a corresponding builder as well.
2. Updating Raw Atlas generation to remove shape points that were used by filtered Ways and to pad identifiers using the existing 6 digit logic (first 3 for country slicing, next 3 for way-sectioning)
3. Updated `RawAtlasCountrySlicer` to handle Relation slicing. 
4. Corresponding unit tests. There are still some TO-DOs in this area
5. Adding synthetic tags to keep track of added relation members and role changes

Still left to do:
1. Add more unit tests
2. Improve readability by separating Relation slicing from line/point slicing
3. Update statistics to track changes more accurately


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -452,9 +1223,12 @@ private void sliceLine(final Line line)
     {
         this.statistics.recordProcessedLine();
         final List<Geometry> slices = convertToJtsGeometryAndSlice(line);
-        processSlices(line, slices);
+        processLineSlices(line, slices);
     }
 
+    // TODO come back and verify we're keeping track of all required statistics
```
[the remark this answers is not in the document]
MikeGost: Yep, I have more changes incoming.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -79,7 +126,108 @@ public RawAtlasCountrySlicer(final Atlas rawAtlas, final Set<IsoCountry> countri
...
+     * Creates a mapping of all inner rings enclosed by the outer rings.
+     *
+     * @param outerToInnerMap
+     *            The mapping to update any time we find an outer ring that contains an inner ring
+     * @param outerRings
+     *            The list of outer {@link LinearRing}s to use
+     * @param innerRings
+     *            The list of inner {@link LinearRing}s to use
+     */
+    final void findEnclosedRings(final MultiMap<Integer, Integer> outerToInnerMap,
```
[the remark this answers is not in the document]
MikeGost: Good catch, they shouldn't be.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -91,20 +239,116 @@ public Atlas slice()
...
+        final List<PolygonPiece> pieces = new ArrayList<>(members.size());
+        final Deque<PolygonPiece> stack = new ArrayDeque<>(members.size());
 
-        return this.rawAtlas;
+        for (final RelationMember member : members)
+        {
+            final Line line = this.rawAtlas.line(member.getEntity().getIdentifier());
+            if (line == null)
+            {
+                logger.error("Relation {} has a member {} that's not in the Atlas",
```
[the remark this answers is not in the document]
MikeGost: Good point.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -91,20 +239,116 @@ public Atlas slice()
...
+
+        while (!stack.isEmpty())
+        {
+            final PolygonPiece currentPiece = stack.pop();
+            while (!currentPiece.isClosed())
+            {
+                boolean foundConnection = false;
+
+                // Try all the members first
+                for (final PolygonPiece piece : stack)
```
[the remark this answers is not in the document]
MikeGost: Looks doable, let me play with the idea.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -79,7 +126,108 @@ public RawAtlasCountrySlicer(final Atlas rawAtlas, final Set<IsoCountry> countri
...
+                {
+                    foundEnclosingOuter = true;
+                    outerToInnerMap.add(outerIndex, innerIndex);
+                }
+            }
+
+            if (!foundEnclosingOuter)
+            {
+                // Isolated inner ring, invalid multipolygon
+                logger.error("Found isolated inner member for Multipolygon Relation geometry: {}",
```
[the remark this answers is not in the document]
matthieun: Sounds good

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -475,43 +1249,217 @@ private void slicePoints()
...
+            final Map<String, String> tags = new HashMap<>();
+            tags.put(ISOCountryTag.KEY, ISOCountryTag.COUNTRY_MISSING);
+            this.slicedRelationChanges.updateRelationTags(relation.getIdentifier(), tags);
+        }
+        else if (countryCount == 1)
+        {
+            // Assign the single country
+            final Map<String, String> tags = new HashMap<>();
+            tags.put(ISOCountryTag.KEY, countryEntityMap.keySet().iterator().next());
+            this.slicedRelationChanges.updateRelationTags(relation.getIdentifier(), tags);
```
jwpgage: If it's possible there's a missing country tag, is it possible that the countryCount could really be 2 here? As in the relation crosses a boundary but one of the country tags is missing so the countryCount is 1.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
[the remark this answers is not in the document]
MikeGost: It worked! Good idea!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +619,118 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+                        // Entity is in the Raw Atlas
+                        final Optional<IsoCountry> possibleCountryCode = ISOCountryTag
+                                .first(entity);
+                        if (possibleCountryCode.isPresent())
+                        {
+                            return possibleCountryCode.get().getIso3CountryCode();
+                        }
+                        else
+                        {
+                            logger.error(
```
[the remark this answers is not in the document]
MikeGost: No, we would return a Missing ISO country code on line 728. Then, when we group members by countries, the members that don't have a country code would be added to both groups. For example, using the test cases scenarios - if there was a multipolygon relation straddling Liberia and the Ivory Coast, and some member did not get a country code (for whatever reason, likely a bug though), that member would be added to both the created Liberia relation and the Ivory Coast relation. See line 1436.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -475,43 +1249,217 @@ private void slicePoints()
...
+                                            memberIdentifier, member.getRole(), memberType)));
+                        }
+                        else
+                        {
+                            members.add(temporaryMember);
+                        }
+                        break;
+                    default:
+                        // Raw Atlas should not have any other types than the ones here
+                        throw new CoreException("Unsupported {} Member for Relation {}", memberType,
```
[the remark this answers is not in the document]
MikeGost: Good suggestion. I've done this in a couple places.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RawAtlasRelationChangeSet.java
```diff
@@ -0,0 +1,152 @@
...
+ * more
+ * <li>4. Relation deletion - relations are deleted when a relation is split into two or more and we
+ * need to get rid of the original one
+ * <li>5. Relation tag updates - a relation country code change or synthetic tag addition will be
+ * recorded here *
+ * <li>6. Relation member addition - a new line member could have been added to a relation to make
+ * an open mulitpolygon closed again
+ * <li>7. Relation member deletion - a relation member could have been deleted if two or more
+ * members were merged into one
+ * <li>8. Line deletion - in the rare case, we might delete a line that was added during way-slicing
```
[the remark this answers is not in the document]
MikeGost: Country slicing of ways :) - will update to clarify

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RawAtlasSimpleChangeSetBuilder.java
```diff
@@ -0,0 +1,216 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * The {@link RawAtlasSimpleChangeSetBuilder} is responsible for applying a
+ * {@link RawAtlasSimpleChangeSet} to a given {@link Atlas}.
+ *
+ * @author mgostintsev
+ */
+public class RawAtlasSimpleChangeSetBuilder extends RawAtlasChangeSetBuilder
```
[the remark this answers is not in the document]
MikeGost: Good callout, this could use some clarification.

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicingTest.java
```diff
@@ -1,60 +1,171 @@
...
+    {
+        COUNTRY_BOUNDARY_MAP = new CountryBoundaryMap(
+                new InputStreamResource(() -> RawAtlasLineAndPointSlicingTest.class
+                        .getResourceAsStream("CIV_GIN_LBR_osm_boundaries.txt.gz"))
+                                .withDecompressor(Decompressor.GZIP));
     }
 
+    // TODO - test with multiple inners across boundary
+    // TODO - test with relations of relations
+
```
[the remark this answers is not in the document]
MikeGost: Great suggestions!

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RelationSlicingTestRule.java
```diff
@@ -0,0 +1,248 @@
...
+    private static final String LOCATION_31 = "6.88017051964,-8.33197681515";
+    private static final String LOCATION_32 = "6.87441616889,-8.32877230817";
+    private static final String LOCATION_33 = "6.8837848,-8.3389237";
+    private static final String LOCATION_34 = "6.87142836787,-8.33437857895";
+    private static final String LOCATION_35 = "6.87193436775,-8.33149050006";
+    private static final String LOCATION_36 = "6.883466,-8.328271";
+    private static final String LOCATION_37 = "6.87407883746,-8.34039743242";
+    private static final String LOCATION_38 = "6.87822317845,-8.33993631058";
+
+    @TestAtlas(
```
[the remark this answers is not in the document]
MikeGost: See [BigNodeFinderTestCaseRule](https://github.com/osmlab/atlas/blob/898a83500ae35fd0c0f791d05ff7f8a1f707820b/src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/bignode/BigNodeFinderTestCaseRule.java) for an ever bigger one :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/CoordinateToNewPointMapping.java
```diff
@@ -0,0 +1,64 @@
...
+ * However, JTS often has precision problems when slicing that causes us to create duplicate Points.
+ * <p>
+ * For example, we'll have a slice that has a Coordinate of (-8.3261734, 6.8663838) and another one
+ * that is really close, but differs at the 7th digit of significance: (-8.3261735, 6.8663838) -
+ * note the 4 and 5 for the latitude value. This will cause us to create two separate Points at what
+ * should be a single Location. To prevent this from happening, we're storing all Coordinates with 6
+ * digits of precision. This class takes care of doing any conversions and lookup for the caller.
+ *
+ * @author mgostintsev
+ */
```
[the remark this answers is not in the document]
MikeGost: Good question! Possibly, I've not seen this happen yet, have only seen the case outlined above, but worth considering. More testing will reveal this. Planning on running this on a few shards to test parity between old and new code.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -60,16 +90,33 @@
...
-    private final Map<Coordinate, Long> newPointCoordinates = new HashMap<>();
+    private final CoordinateToNewPointMapping newPointCoordinates = new CoordinateToNewPointMapping();
 
     private final RawAtlasSlicingStatistic statistics = new RawAtlasSlicingStatistic(logger);
 
+    /**
+     * JTS has trouble dealing with high-precision double values. For this reason, we round all
+     * coordinates to 7 degrees of precision. See {@link Coordinate} java doc for more detailed
+     * explanation of scaling.
+     *
```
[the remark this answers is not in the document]
MikeGost: We are storing all Latitude/Longitude values with dm7 precision, to be consistent with the rest of the code base. The class above, that references 6 degrees of precision, is only used to store a mapping between Coordinate to Point identifier. This is only used for quick lookup to verify we don't have duplicate points, otherwise everything is dm7.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -292,32 +771,319 @@ private boolean lineBelongsToSingleCountry(final Line line, final List<Geometry>
...
+                    markRemovedMemberLineForDeletion(inner, relationIdentifier);
+                }
+
+                // Create points (if necessary) and lines for the merged member
+                final com.vividsolutions.jts.geom.Polygon merged = (com.vividsolutions.jts.geom.Polygon) mergedMembers;
+                final LineString exterior = merged.getExteriorRing();
+
+                // Set the proper country code
+                CountryBoundaryMap.setGeometryProperty(exterior, ISOCountryTag.KEY,
+                        ISOCountryTag.first(outer).get().getIso3CountryCode());
```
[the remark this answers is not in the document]
MikeGost: Good callout! I wouldn't put it past the data :) - in which case this would be an issue. I'll add it to the non-happy path testing TO-DO list.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: We remove all missing country code tags on line 1397, otherwise it would definitely be possible.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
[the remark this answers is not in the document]
MikeGost: Took a closer look at this - actually this case is already handled. If you look at `findIntersectingMembers` method - it checks that the intersecting features must be from the same country before marking them as a mergeable intersection. Either way - good to add a test case that handles this!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/identifier/ReverseIdentifierFactory.java
```diff
@@ -17,6 +17,13 @@ public long getCountryOsmIdentifier(final long countryCodeAndWaySectionedIdentif
                 countryCodeAndWaySectionedIdentifier / AbstractIdentifierFactory.IDENTIFIER_SCALE);
     }
 
+    public long getFirstAtlasIdentifier(final long countryCodeAndWaySectionedIdentifier)
```
mgcuthbert: Method comments would be useful to include here.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfReader.java
```diff
@@ -526,8 +526,7 @@ private void processWay(final Entity entity)
      * However, if we filter out a PBF {@link Way}, we want to keep track of all {@link Node}s that
      * make up that {@link Way}, in case we need to remove them as well. We are going to keep track
      * of all filtered PBF {@link Node}/Atlas {@link Point} identifiers, and use subAtlas
-     * functionality to filter them out after the Atlas is built. For now, ignore Relations. TODO
-     * Come back to Relations!
+     * functionality to filter them out after the Atlas is built. For now, ignore Relations.
```
mgcuthbert: Having removed the TODO would indicate that it is not required anymore, however you still have "For now..." which indicates that you plan on introducing it later. Is the plan still to introduce code to not ignore Relations in the future, if not then we should just update the comments to read "This ignores relations" or something to that effect.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+            if (!foundEnclosingOuter)
+            {
+                // Isolated inner ring, invalid multipolygon
+                logger.error("Found isolated inner member for Multipolygon Relation geometry: {}",
+                        inner);
+            }
+        }
+    }
+
+    /**
```
mgcuthbert: This is more a personal stylistic thing for me, but when you add java docs for a method, I think it is good to include a description of the method above. I mean in this particular instance the method name is fairly self explanatory, but I think it should include a brief description for clarity and style.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+                        outerToInnerIntersectionMap.add(outerIndex, innerIndex);
+                    }
+                }
+            }
+        }
+
+        return outerToInnerIntersectionMap;
+    }
+
+    /**
```
mgcuthbert: Same as above comment.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+        final Optional<IsoCountry> countryOne = ISOCountryTag.first(one);
+        final Optional<IsoCountry> countryTwo = ISOCountryTag.first(two);
+        if (countryOne.isPresent() && countryTwo.isPresent())
+        {
+            return countryOne.get().equals(countryTwo.get());
+        }
+        return false;
+    }
+
+    /**
```
mgcuthbert: Same as previous comment.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -527,9 +1430,52 @@ private void updateLineShapePoints(final Line line)
...
+            else
+            {
+                // Synthetic key doesn't exist, need to insert a new value
+                updatedTags.put(SyntheticRelationMemberAdded.KEY.toString(),
+                        String.valueOf(newLineIdentifier));
+            }
+            this.slicedRelationChanges.updateRelationTags(relationIdentifier, updatedTags);
+        }
+    }
 }
```
mgcuthbert: This file is now 1500 lines long, which is getting to the point of considering why it is so long and possibly if there are any changes that can be made to clean it up or break it up. Readability, testability and understanding classes of this size start breaking down fairly quickly once you start getting up to that many lines.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/SimpleChangeSet.java
```diff
@@ -0,0 +1,112 @@
...
+ * {@link TemporaryEntity} objects to keep track of the bare minimum needed to create Atlas
+ * entities. For tag updates, we store a mapping between the identifier of the object and the added
+ * tags. For deletions, we store the set of identifiers to delete. We also keep a mapping between
+ * deleted and created lines to properly update any relations. No relation slicing is done with this
+ * class. We only update existing relations if a member line was deleted and split into two or more
+ * lines.
+ *
+ * @author mgostintsev
+ */
+public class SimpleChangeSet
```
mgcuthbert: There are now 2 classes called SimpleChangeSet https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/geography/atlas/changeset/SimpleChangeSet.java 
It doesn't seem like the one I have referenced is actually used, so might considering deleting that one.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: I totally agree with you! See line 1185 for a TO-DO calling out this exact item. I'm working on adding a few more test cases and then separating this behavior into something more concise.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Yeah, I see what you mean. I just thought it might be a bit repetitive to add javadoc description as well as the return field. Will update it though, to be consistent with the other method javadocs.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Will update!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Updated!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
mgcuthbert: Nice, I obviously skipped over the TODO.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfReader.java
MikeGost: Good callout! The intent here is to filter out any OSM Node/Atlas Point identifiers that make up the filtered Way. If we were to start removing Nodes from Filtered Relations - that logic may get very tricky, since the filtered Relation may not use the Node, but some underlying way might. We want to avoid this, and push this logic to way-sectioning code. I've updated the comment.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -141,18 +243,50 @@ public Atlas slice()
...
+    /**
+     * @param relation
+     *            The {@link Relation} for which to create seeds for
+     * @return an array of identifiers corresponding to the first unused identifier for each
+     *         relation member in the given Relation
+     */
+    private long[] createIdentifierSeeds(final Relation relation)
+    {
+        // Assume any member can be split, so create a seed for each one
+        final long[] seeds = new long[relation.members().size()];
```
cstaylor: Dumb question: is it possible to have a `relation` with no members?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+     * @return a mapping of intersecting outer to inner relation members
+     */
+    private MultiMap<Integer, Integer> findIntersectingMembers(final List<Line> closedOuterLines,
+            final List<Line> closedInnerLines)
+    {
+        final MultiMap<Integer, Integer> outerToInnerIntersectionMap = new MultiMap<>();
+
+        if (!closedOuterLines.isEmpty() && !closedInnerLines.isEmpty())
+        {
+            for (int innerIndex = 0; innerIndex < closedInnerLines.size(); innerIndex++)
```
cstaylor: Maybe hoist `closedInnerLines.size()` out as a temporary variable since it's used a few times in the loop and looks to be invariant at first glance.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+     */
+    private boolean fromSameCountry(final Line one, final Line two)
+    {
+        final Optional<IsoCountry> countryOne = ISOCountryTag.first(one);
+        final Optional<IsoCountry> countryTwo = ISOCountryTag.first(two);
+        if (countryOne.isPresent() && countryTwo.isPresent())
+        {
+            return countryOne.get().equals(countryTwo.get());
+        }
+        return false;
```
cstaylor: So what happens if we neither `one` or `two` don't have ISO country codes?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,215 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+                final Line line = this.rawAtlas.line(member.getEntity().getIdentifier());
+                if (line == null)
+                {
+                    logger.error("Line member {} for Relation {} is not the in raw Atlas.",
+                            member.getEntity().getIdentifier(), relationIdentifier);
+                }
+
+                if (!closed)
+                {
+                    return line == null || !line.isClosed();
```
cstaylor: So `null` `Line` objects should be permitted?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -292,32 +726,319 @@ private boolean lineBelongsToSingleCountry(final Line line, final List<Geometry>
...
+     * @param line
+     *            The relation member {@link Line}
+     * @param relationIdentifier
+     *            The parent {@link Relation} identifier
+     */
+    private void markRemovedMemberLineForDeletion(final Line line, final long relationIdentifier)
+    {
+        // Only mark this line for deletion if this line isn't part of any other relations
+        final boolean isPartOfOtherRelations = line.relations().stream()
+                .filter(partOf -> partOf.getIdentifier() != relationIdentifier).count() > 0;
```
cstaylor: Could possibly abort this sooner by `.findFirst().isPresent()`?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -292,32 +726,319 @@ private boolean lineBelongsToSingleCountry(final Line line, final List<Geometry>
...
+
+                    final List<LineString> borderLines;
+                    try
+                    {
+                        borderLines = this.countryBoundaryMap.clipBoundary(relationIdentifier,
+                                polygon);
+                    }
+                    catch (final Exception e)
+                    {
+                        logger.error("Error processing relation {}, message: {}, geometry: {}",
```
cstaylor: Do we want to lose the cause of the exception here?

MikeGost: Not at all, good catch!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: No such thing as a dumb question! It's not possible - one of the Atlas relation constraints is that we cannot have a Relation without any members. See [here](https://github.com/osmlab/atlas/blob/1dbcbe5818d018f7c919b51d8ea82902ec1a5fa7/src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java#L923) for reference code. So at this point, we can safely assume there will be at least one member.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: I only see it used once. Do you mean pull out the sizes of the outer and inner lines? If I remember correctly, ArrayList does this under the covers for us already.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Impossible, at this point, there's only 3 Atlas entities - Points, Lines and Relations. Relations are the only ones that haven't been assigned a country code yet, everything else has. I'll add an error condition just in case!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/identifier/ReverseIdentifierFactory.java
MikeGost: Good idea, let me add them for the entire class.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/SimpleChangeSet.java
MikeGost: Good callout! There are a number of change sets in the Atlas codebase, some no longer applicable. I didn't want to use the one above since it had fields I would never use. In this case, I thought that creating a country-slicing specific change-set to make it specific and light-weight was the right way to go. Hopefully being in the ***/slicing package makes this more clear. Having said that, there are a number of classes that can and will be removed after this refactor is done. SimpleChangeSet above is a candidate. Thanks Mike!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Good idea!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Good catch! The original intent was to allow to process members that may have been added on the fly (part of the relation, but not in the raw Atlas - only in the temporary relation member change set version). However, looking again, I've refactored this in a such a way this can never happen. Will not allow null members here.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -91,20 +144,69 @@ public Atlas slice()
...
-        // Apply changes and rebuild the Atlas with the changes before slicing Relations
-        final RawAtlasChangeSetBuilder changeBuilder = new RawAtlasChangeSetBuilder(this.rawAtlas,
-                this.afterSlicingLinesAndPoints);
-        final Atlas withSlicedWaysAndPoints = changeBuilder.applyChanges();
-        this.rawAtlas = withSlicedWaysAndPoints;
+        // Apply changes and rebuild the atlas with the changes before slicing relations
+        final ChangeSetHandler simpleChangeBuilder = new SimpleChangeSetHandler(this.rawAtlas,
+                this.slicedPointAndLineChanges);
+        final Atlas atlasWithSlicedWaysAndPoints = simpleChangeBuilder.applyChanges();
+        this.rawAtlas = atlasWithSlicedWaysAndPoints;
```
AlexHsieh22: is there a reason why rawAtlas is not final (it seems to have lost its value after the first round of slicing)?  A follow up question is, why is rawAtlas kept in an intermediate state of having only slicedWaysAndPoints and not relations?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,226 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+                }
+
+                if (!closed)
+                {
+                    return line != null && !line.isClosed();
+                }
+                else
+                {
+                    return line != null && line.isClosed();
+                }
```
AlexHsieh22: what about:
```
if(line == null)
{
    // log error
    return false;
}

boolean isClosed = line.isClosed();
return closed ? isClosed : !isClosed;
```

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -252,6 +477,226 @@ private TemporaryPoint createNewPoint(final Coordinate coordinate,
...
+     *            The members we wish to group
+     */
+    private Map<String, List<TemporaryRelationMember>> groupRelationMembersByCountry(
+            final List<TemporaryRelationMember> members)
+    {
+        final Map<String, List<TemporaryRelationMember>> countryEntityMap = members.stream()
+                .collect(Collectors.groupingBy(member ->
+                {
+                    final AtlasEntity entity;
+                    if (member.getType() == ItemType.POINT)
```
AlexHsieh22: consider using a switch statement instead?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -292,32 +737,329 @@ private boolean lineBelongsToSingleCountry(final Line line, final List<Geometry>
...
+     * @param line
+     *            The relation member {@link Line}
+     * @param relationIdentifier
+     *            The parent {@link Relation} identifier
+     */
+    private void markRemovedMemberLineForDeletion(final Line line, final long relationIdentifier)
+    {
+        // Only mark this line for deletion if this line isn't part of any other relations
+        final boolean isPartOfOtherRelations = line.relations().stream()
+                .filter(partOf -> partOf.getIdentifier() != relationIdentifier).findFirst()
```
AlexHsieh22: anyMatch might be easier than a combination of findFirst and isPresent

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: It's not final because we want to rebuild the atlas after applying the changes from the way/point slicing. The reason why it's separated into two is because slicing ways has the potential to create many new ways and has the implications on relations (deleted the original member and added 2 or more sliced lines that used to make up the deleted member). Dealing with this many number of changes without them being in the Atlas was 1. cumbersome 2. not very efficient when we wanted to make use of Atlas geospatial queries for ways that were sliced but not part of the atlas. Currently in the process of splitting this class up some to separate behavior between the point/way and relation slicing. This may address your concern of non-final.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/identifier/ReverseIdentifierFactory.java
```diff
@@ -1,40 +1,109 @@
 package org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier;
 
 /**
+ * Common methods to extract information from an Atlas identifier. Atlas identifier is defined as
+ * {OSM identifier}{3 digits for country slicing id}{3 digits for way-section id}.
```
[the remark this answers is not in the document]
MikeGost: I don't think it's worth calling out explicitly here, there is consistency and comments around the terminology. Going to leave as is.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Nice, I like it!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Good call!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
MikeGost: Yeah, that will look a little cleaner!

PR #40, opened 2017-12-19
DynamicAtlas performance improvement for 3rd party extension shards

matthieun:
This PR solves an issue where in some cases, the `preemptiveLoad()` function on `DynamicAtlas` was building the same underlying `MultiAtlas` twice for no reason. The only time a re-build of the underlying `MultiAtlas` is useful is when a feature intersects the initial shard(s) without any shapepoint inside the initial shard(s).

- Improved documentation and comments
- Added a test for the specific use case above
- Renaming the `loadShards()` method to `buildUnderlyingMultiAtlas()` which is more descriptive.


src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -138,6 +138,33 @@ public Rectangle bounds()
         return this.current.bounds();
     }
 
+    public void buildUnderlyingMultiAtlas()
+    {
+        logger.debug("{}: Loading MultiAtlas with {}", this.getName(),
```
AlexHsieh22: should this be wrapped in a ```logger.isDebugEnabled()``` block?  No reason to build the shard names if this log message won't get emitted

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -374,20 +376,50 @@ public Point point(final long identifier)
...
+        Set<Shard> currentShards = new HashSet<>(this.loadedShards.keySet());
+        // Loop through the entities again to find potential shards to add. This can still happen if
+        // a way intersects the initial shard without shapepoints inside the initial shards, and was
+        // revealed by loading a new neighboring shard. At that point, if that way also intersects a
+        // third shard which was not loaded before, that third shard might become now eligible.
+        this.entities();
+        // Repeat the same process as long as we find some of those third party shards.
         while (!this.loadedShards.keySet().equals(currentShards))
         {
+            final Set<Shard> missingShards = new HashSet<>(this.loadedShards.keySet());
```
AlexHsieh22: consider wrapper line 408 - 415 in a ```logger.isInfoEnabled()``` block.  This code is only used for the logger call.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -374,20 +376,50 @@ public Point point(final long identifier)
...
+            for (final Shard currentShard : currentShards)
+            {
+                missingShards.remove(currentShard);
```
AlexHsieh22: what about doing ```missingShards.removeAll(currentShards)``` instead of this explicit for loop?

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -374,20 +376,50 @@ public Point point(final long identifier)
...
     {
-        Set<Shard> currentShards = null;
+        // Loop through the entities to find potential shards to add
+        this.entities();
+        // Load all the shards into a multiAtlas
+        this.buildUnderlyingMultiAtlas();
+        // Record the current list of shards
+        Set<Shard> currentShards = new HashSet<>(this.loadedShards.keySet());
+        // Loop through the entities again to find potential shards to add. This can still happen if
+        // a way intersects the initial shard without shapepoints inside the initial shards, and was
```
AlexHsieh22: stupid question: when would a way intersect a shard without shape points inside the shard?  Would that be when the way ends exactly on the boundary of the shard?  And the case of a third shard, would that be a way ends exactly on the boundary of where multiple shards meet?

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -374,20 +376,50 @@ public Point point(final long identifier)
...
+            logger.info("{}: Preemptive load found new unexpected 2nd degree shard(s): {}",
+                    this.getName(),
+                    missingShards.stream().map(Shard::getName).collect(Collectors.toList()));
+
+            // Load all the shards into a multiAtlas
+            this.buildUnderlyingMultiAtlas();
+            // Record the current list of shards
             currentShards = new HashSet<>(this.loadedShards.keySet());
+            // Loop through the entities again to find potential shards to add.
             this.entities();
```
AlexHsieh22: doesn't ```this.entities``` internally call ```this.buildUnderlyingMultiAtlas()```? Why do we need an explicit call?

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -138,6 +138,33 @@ public Rectangle bounds()
...
+        final List<Atlas> nonNullAtlasShards = getNonNullAtlasShards();
+        if (!nonNullAtlasShards.isEmpty())
+        {
+            if (nonNullAtlasShards.size() == 1)
+            {
+                this.current = nonNullAtlasShards.get(0);
+            }
+            else
+            {
+                this.current = new MultiAtlas(nonNullAtlasShards);
```
AlexHsieh22: instead of always creating a new multi atlas with all shards, is there a way we can create a multi atlas with the current DynamicAtlas and the new shards?  Would that help in performance?

src/test/resources/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasPreemptiveLoadTest.osm
```diff
@@ -0,0 +1,41 @@
+<?xml version='1.0' encoding='UTF-8'?>
```
MikeGost: Is the .osm file an intended addition?

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
MikeGost: +1

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: That sounds good!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: Every MultiAtlas creates some overhead and does some indexing of all the features in the sub-atlas it contains. So as this would not improve performance that much, it would increase memory consumption.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: Here is the fake data I created for the test in that PR:
screen shot 2017-12-20 at 9 48 35 am
You see that the lower left shard is intersected by the red way, but none of its shapepoints are inside.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: Good idea!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: oh yes!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: This is a very good question. When the policy is set to deferLoading = true, then the entities call will identify shards to load but not load them. That makes me think that preemptive loading is useful only when defer loading is set to true, so I will make sure it does nothing when the deferLoading is false.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -374,20 +376,50 @@ public Point point(final long identifier)
...
     /**
-     * In case of an Atlas shard with neighbors containing data that might intersect it (example is
-     * a way from a neighboring shard intersecting the initial shard, but without any shapepoints in
-     * the initial shard) this method allows dual loading of neighboring shards. Once to load the
-     * neighboring shards if any, and to discover such features if any, then to load the other
-     * shards those features might intersect.
+     * Do a preemptive load of the {@link DynamicAtlas} as far as the {@link DynamicAtlasPolicy}
+     * allows.
+     * <p>
+     * In some very specific case, where the {@link DynamicAtlasPolicy} allows expansion only if new
```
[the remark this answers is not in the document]
matthieun: I actually thought "case" as "one type of case". Do you think it sounds correct?

src/test/resources/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasPreemptiveLoadTest.osm
matthieun: Yes, I used JOSM to create it, and this way the test data is easy to open in JOSM.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
matthieun: And if created from an osmosis PBF, the lower left shard will not contain the red way. It will be discovered only when expanding to the upper left or lower right shards. Then that uncovers a new way that intersects the initial shard, which in turn makes the upper right shard eligible for loading on the second pass, when it was not initially.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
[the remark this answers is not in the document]
MikeGost: Some implied multiple to me, thought the correct approach was "in some very specific cases...". Feel free to ignore!

src/test/resources/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasPreemptiveLoadTest.osm
MikeGost: Yeah, I saw that. Just this is a first for the Atlas project. We might consider adding a "Test best practices" section - explaining that we have TestAtlas files, .osm files, .txt files and .atlas.gz in some rare cases and when it's best to use each one.

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
AlexHsieh22: Oh I see! Thanks for the explanation

Issue #35, opened 2017-11-16
Warnings to fix in the code


matthieun:
There are some warnings that need to be fixed (crossed out means fixed):

Link|Warning|Code
-----|-----|-----
~~[AtlasPrimitiveBigNode.java#L131](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/store/AtlasPrimitiveBigNode.java#L131)~~|Unlikely argument type `Location` for `contains(Object)` on a `Collection<AtlasPrimitiveLocationItem>`|`!nodes().contains(edge.end())`
~~[GeoJsonObject.java#L53](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonObject.java#L53)~~|Unlikely argument type for `equals()`: `JsonElement` seems to be unrelated to `String`|`!"FeatureCollection".equals(this.jsonObject.get("type"))`
~~[Tagger.java#L50](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/main/java/org/openstreetmap/atlas/tags/cache/Tagger.java#L50)~~|Unlikely argument type `Optional<String>` for `containsKey(Object)` on a `Map<String,Optional<T>>`|`!this.storedTags.containsKey(tagValue)`
~~[ComplexWaterEntityTest.java#L86](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/waters/ComplexWaterEntityTest.java#L86)~~|Unlikely argument type `ItemType` for `get(Object)` on a `Map<WaterType,Integer>`|`waterTypeCount.get(type)`
~~[TemporaryEntityTest.java#L37](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/TemporaryEntityTest.java#L37)~~|Unlikely argument type for `equals()`: `TemporaryLine` seems to be unrelated to `TemporaryPoint`|`pointOne.equals(lineOne)`

MikeGost:
[...]

PR #43, opened 2018-01-02
Fix warnings

matthieun:
This fixes the remaining warnings in #35.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/store/AtlasPrimitiveBigNode.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/store/AtlasPrimitiveBigNode.java
@@ -116,6 +116,12 @@ public class AtlasPrimitiveBigNode extends AtlasPrimitiveEntity
                 .collect(Collectors.toSet());
     }
 
+    public Set<Location> nodeLocations()
+    {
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/builder/store/AtlasPrimitiveBigNode.java
```diff
@@ -116,6 +116,12 @@ public Rectangle bounds()
                 .collect(Collectors.toSet());
     }
 
+    public Set<Location> nodeLocations()
+    {
+        return this.nodes.stream().map(AtlasPrimitiveLocationItem::getLocation)
```
MikeGost: Nit: should we use `this.nodes()`?

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/store/AtlasPrimitiveBigNode.java
```diff
@@ -128,7 +134,7 @@ public boolean nodesContain(final Location location)
 
     public Set<AtlasPrimitiveEdge> outEdges()
     {
-        return edges().stream().filter(edge -> !nodes().contains(edge.end()))
+        return edges().stream().filter(edge -> !nodeLocations().contains(edge.end()))
```
MikeGost: `this.nodeLocations()`?

PR #46, opened 2018-01-04
Raw Atlas Relation Slicing Improvements

MikeGost:
Summary of changes:

1. For raw atlas lines that could not be sliced (due to running out of point or line identifiers - think OSM Way intertwining the country boundary, requiring an individual slice on each side of the border) - we used to assign an N/A country code. Instead, we now collect the country codes from all generated slices and assign the cumulative value to the un-cut Line. 
2. Updated relation handling. Namely two changes:

- Removed recursive relation parent handling. This wasn't needed, since we are relying on the Atlas `relationsLowerOrderFirst` method and are always guaranteed to get the leaf relations first. 

- A relation can either be updated (country code added) or deleted (and replaced by two or more new relations). This logic is updated and simplified in the `RelationChangeSet` and `RelationChangeSetHandler`. We now keep a mapping between deleted to created relations for easier lookup. This simplified other pieces across the code.

3. Moved an existing unit test to an integration test to show the full OSM PBF to Raw Atlas to Sliced Raw Atlas flow. The PBF used had some interesting use cases - (a) Complex multipolygon spanning 3 countries (b) Nested relations. This effort uncovered a few minor issues which were fixed above

4. Minor fixes, typos and TO-DO completions across the codebase

5. I was playing around with replicating new Java 9 Collectors functionality per a stack overflow question I found. Originally, was planning on using this to group Lines with multiple country codes into country buckets, but found an easier way. Decided to keep the Collector code in place in case it can be used elsewhere. Read more on the change [here](http://iteratrlearning.com/java9/2016/08/16/java9-collectors.html).


MikeGost:
Closing and re-opening to trigger Travis build.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -808,7 +818,7 @@ private void preProcessMultiPolygonRelation(final Relation relation)
      * @param relation
      *            The {@link Relation} to slice
      */
-    private List<TemporaryRelation> sliceRelation(final Relation relation, final List<Long> parents)
+    private List<TemporaryRelation> sliceRelation(final Relation relation)
```
matthieun: Parents were not used before here?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RelationChangeSet.java
```diff
@@ -57,16 +57,23 @@
     private final Set<Long> deletedLines;
     private final Set<Long> deletedRelations;
 
+    // Relations that were sliced will be deleted and replaced by two or more new Relations. We need
```
matthieun: Nice comments! Really easy to read and understand!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RelationChangeSetHandler.java
```diff
@@ -54,7 +55,7 @@ public RelationChangeSetHandler(final Atlas atlas, final RelationChangeSet chang
     public Atlas applyChanges()
     {
         // Log original Atlas statistics
-        logger.info(atlasStatistics(super.getAtlas()));
+        logger.info("Before Slicing Relations: " + atlasStatistics(super.getAtlas()));
```
matthieun: If those logger calls generate a lot of processing, it might be nice to wrap them in `if (logger.isInfoEnabled())` so it does not trigger when the logging level is at WARN or higher?

src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
```diff
@@ -0,0 +1,109 @@
...
+    public static void setup()
+    {
+        COUNTRY_BOUNDARY_MAP = new CountryBoundaryMap(
+                new InputStreamResource(() -> OsmPbfToSlicedRawAtlasTest.class
+                        .getResourceAsStream("CIV_GIN_LBR_osm_boundaries.txt.gz"))
+                                .withDecompressor(Decompressor.GZIP));
+    }
+
+    @Test
+    @Ignore("Taking too long, but still valuable to have")
```
matthieun: Might be worth making smaller?

MikeGost: Yeah, possibly. I have an optimization in mind for assigning country codes for Points - that's the part that's taking the most right now. Instead of assigning each individual Point a country code - only assign ones to existing ones on the boundary or newly created ones. All other ones are guaranteed to be points in lines that are contained within one country. Toying with this idea now!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
MikeGost: Take a look at the PR description - we're removing recursive relation slicing because we're looking at them in order of lowest to highest.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RelationChangeSet.java
MikeGost: Thanks!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/RelationChangeSetHandler.java
MikeGost: Only a single line. Here's an example. Before:
2018-01-05 11:52:12 INFO  [main] RelationChangeSetHandler:58 - Points: 648709 Lines: 57643 Relations: 36
2018-01-05 11:52:16 INFO  [main] RelationChangeSetHandler:77 - Points: 648709 Lines: 57643 Relations: 44

After:
2018-01-05 11:52:12 INFO  [main] RelationChangeSetHandler:58 - Before Slicing Relations: Points: 648709 Lines: 57643 Relations: 36
2018-01-05 11:52:16 INFO  [main] RelationChangeSetHandler:77 - After Slicing Relations: Points: 648709 Lines: 57643 Relations: 44

MikeGost: Wanted to add something to different the two.

matthieun: I meant more if `atlasStatistics(super.getAtlas())` takes some time to process, only process it when the logger will actually print it!

MikeGost: Good point, will add this.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -407,4 +415,28 @@ private void updateLineShapePoints(final Line line)
...
+     *            The {@link Line} in question
+     * @param slices
+     *            The {@link Geometry} slices for the given {@link Line}
+     */
+    private void updateLineToHaveCountryCodesFromAllSlices(final Line line,
+            final List<Geometry> slices)
+    {
+        final Map<String, String> tags = new HashMap<>();
+        final Set<String> allCountries = new HashSet<>();
+        slices.forEach(geometry -> allCountries
```
AlexHsieh22: This could result in varied ordering of the isoCountryCodes which could result in different values for ```IsoCountryTag.first```.  Do we want to consider enforcing some order here?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -407,4 +415,28 @@ private void updateLineShapePoints(final Line line)
...
+                .add(CountryBoundaryMap.getGeometryProperty(geometry, ISOCountryTag.KEY)));
```
AlexHsieh22: An alternative here would be:
```
slices.stream().map(geometry -> CountryBoundaryMap.getGeometryProperty(geometry, ISOCountryTag.KEY)).collect(Collectors.toSet());
```
which would prevent you from having to specifically initialize the hashSet.

MikeGost: Good call!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
MikeGost: Good catch, updated!

matthieun:
Relates to #17

PR #47, opened 2018-01-09
DynamicAtlas violating policy when using edgesIntersecting and similar methods

matthieun:
When loading a DynamicAtlas with deferred loading (true) and extend indefinitely (false), and the user calls the `edgesIntersecting(Polygon)` method with a polygon that covers previously un-loaded shards, it would trigger the loading of those shards, violating the extend indefinitely (false) policy. This is a fix for that issue. Removing the `newLocation`, `newPolygon` etc. calls is not an issue when the policy is different, as the `this::areaCovered`, `this::lineItemCovered` etc. methods are still there to take care of expansions.

Added one test, and some package reorganization for tests as well.


src/test/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasMovingTooFastTest.java
```diff
@@ -0,0 +1,84 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.sharding.Shard;
+import org.openstreetmap.atlas.geography.sharding.SlippyTile;
+import org.openstreetmap.atlas.geography.sharding.SlippyTileSharding;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
+
+/**
+ * @author matthieun
+ */
+public class DynamicAtlasMovingTooFastTest
```
MikeGost: You gave a really good explanation of what moving too fast usually means. Can you add it here? It would give more clarity to what's being tested.

matthieun:
Correct! Let me add the explanation to the test comments.

Issue #10, opened 2017-09-01
OSM Nodes in Relations get duplicated to Point


matthieun:
From https://github.com/osmlab/atlas/pull/8#discussion_r136627217

OSM Nodes that are part of a Relation are also duplicated as an Atlas Point. That should probably not be the case if they do not have any OSM tags.

mertemin:
I think nodes and points should be exclusive though (OSM node either becomes a node or point). Logic could be simplified as **any OSM node that is not converted to Atlas node becomes an Atlas point**. OSM nodes at the end of edges, OSM nodes that are part of relations (most likely these would be turn restrictions) and OSM nodes on boundaries could be converted to Atlas nodes.

MikeGost:
Related PR #48

PR #48, opened 2018-01-09
Removing superfluous Atlas Point

MikeGost:
**Description**
1. We have a specific circumstance, where an OSM Node is translated as both an Atlas `Point` and an Atlas `Node` during `Atlas` creation. This happens when the OSM Node is part of a Relation, doesn't contain any explicit OSM tags and is at an intersection. This is extra data that we don't need, since the created Atlas `Point` doesn't give any additional information that is not already contained in the created Atlas `Node`. This change makes it so the `Point` is never created.
2. Tests for all OSM Node scenarios
3. Fix javadoc warning during build

As part of the exercise and to test for completeness, I've created a chart to show how an OSM Node will get translated to Atlas format, depending on the circumstance. The bolded row is what was fixed in this PR (used to produce both Point and Node)

| Relation Member | Has Tags | At an intersection  |          Atlas Item          |
|:---------------:|:--------:|:-------------------:|:----------------------------:|
|       Yes       |    Yes   |         Yes         |  Atlas Node and Atlas Point  |
|       Yes       |    Yes   |          No         |          Atlas Point         |
|       **Yes**       |    **No**    |         **Yes**         |          **Atlas Node**          |
|       Yes       |    No    |          No         |          Atlas Point         |
|        No       |    Yes   |         Yes         |  Atlas Node and Atlas Point  |
|        No       |    Yes   |          No         |          Atlas Point         |
|        No       |    No    |         Yes         |          Atlas Node          |
|        No       |    No    |          No         | Neither - simple shape point |

**Testing**
1. Verified all the cases in the table above 
2. Validated the change on Dominican Republic 8-77-113 shard. Specifically, there were 5 cases that ran into this case there:
Case 1: http://www.openstreetmap.org/relation/6088504 (target [Node](http://www.openstreetmap.org/node/350685878)) 
Case 2: http://www.openstreetmap.org/relation/6138961 (target [Node](http://www.openstreetmap.org/node/2033937407#map=19/19.44991/-70.68681)) 
Case 3: http://www.openstreetmap.org/relation/6117258 (target [Node](http://www.openstreetmap.org/node/249651017#map=19/19.41281/-70.64795)) 
Case 4: http://www.openstreetmap.org/relation/6138349 (target [Node](http://www.openstreetmap.org/node/4121662786#map=19/19.44963/-70.68679)) 
Case 5: http://www.openstreetmap.org/relation/6138962 (target [Node](http://www.openstreetmap.org/node/4122105419)). I ran Atlas Generation for this shard pre-change and post-change and verified that after the changes, there were 5 less Points in the Atlas, and everything else was identical. 

Pre: 
`[PackedAtlas: Nodes = 42,544, Edges = 111,640, Areas = 19,114, Lines = 818, Points = 11,268, Relations = 62]`

Post:
`[PackedAtlas: Nodes = 42,544, Edges = 111,640, Areas = 19,114, Lines = 818, Points = 11,263, Relations = 62]`

where the 5 removed Points were the Nodes listed above. 

3. Verified that `testOsmNodePartOfRelationNoTagsAtIntersection()` fails with the old code and passes with the new.
4. This fixed a small issue in today's code and was a good exercise for clarifying requirements and creating re-usable tests for Part 3 of issue #17. This is also tied to issue #10. Regarding #10 - we still have two cases where we will have both Points and Nodes, per discussion with @matthieun. This can be easy to fix and verify if we want to change OSM Node to a one-to-one mapping to Atlas Nodes/Points.


src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfNodeToAtlasItemTestRule.java
```diff
@@ -0,0 +1,77 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas;
+
+/**
+ * {@link OsmPbfNodeToAtlasItemTest} test cases.
+ *
+ * @author mgostintsev
+ */
+public class OsmPbfNodeToAtlasItemTestRule extends CoreTestRule
+{
+    @TestAtlas(loadFromJosmOsmResource = "noRelationNoTagsNoIntersectionAtlas.osm")
```
matthieun: As the `loadFromJosmOsmResource` uses the PBF loading to generate the test Atlas every time, it cannot be used to test the PBF loading... I would recommend saving all those Atlas files to textAtlas one time, and then loading them using `loadFromTextAtlas`. Like this the pbf loading is out of the equation, and you keep the same test set!

MikeGost: Fair point, let me update it!

MikeGost: Not sure I fully agree here. If we use text atlases, then we're just loading an Atlas using PackedAtlasBuilder under the covers. This doesn't test PBF entity to Atlas Item conversion in any way. I think it actually makes sense to use .osm files here to test that code flow. Thoughts?

matthieun: Oh I did not realize that the test depended on having something loaded using the PBF loader. In that case this is fine.
Mind adding a comment explaining why `loadFromJosmOsmResource` is actually right here?

MikeGost: Sure, let me add one.

PR #52, opened 2018-01-11
Use fastclasspath to get around reflections bug

matthieun:
Reflections has [this issue](https://github.com/ronmamo/reflections/issues/81).

This is a thread-safety issue in the reflections framework that hasn’t been resolved for over a year, so moving to fastclasspath allows getting around it.


dependencies.gradle
```diff
@@ -15,7 +15,7 @@ project.ext.versions = [
     cli: '1.3.1',
     csv: '1.2',
     lang: '3.4',
-    reflections: '0.9.10',
+    fastclasspath: '2.8.0',
```
MikeGost: Can we upgrade to the latest - 2.9.4? See https://github.com/lukehutch/fast-classpath-scanner/releases for release notes.

PR #58, opened 2018-01-12
Replaced reflections with fast-classpath-scanner

cstaylor:
After experiencing [this bug](https://github.com/ronmamo/reflections/issues/81) when using multiple `Validators` in a multithreaded application, I fixed the problem by replacing it with `fast-classpath-scanner`.


cstaylor:
Duplicate of PR 52.  Closing.

PR #59, opened 2018-01-12
Thread safety on the AtlasFeatureCountsSubCommand

cstaylor:
If `AtlasReader` is used with multiple shards it will use multiple threads in the `handle` method, meaning we need to be careful about how we store values in the non-concurrent hashmap.

I also wrapped the `FileOutputStream` in a `BufferedOutputStream` for smoother I/O performance.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFeatureCountsSubCommandTestCase.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFeatureCountsSubCommandTestCase.java
@@ -8,8 +8,7 @@ import org.junit.BeforeClass;
 import org.junit.ClassRule;
 import org.junit.Test;
 import org.openstreetmap.atlas.streaming.resource.File;
-
-import com.google.common.collect.Iterables;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
 
 /**
  * Test case verifying that protecting against multiple concurrent threads in the AtlasReader report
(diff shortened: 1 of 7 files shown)


cstaylor:
@matthieun do you mean for feature counts in general or for the thread synchronization?

matthieun:
Just for the thread synchronization, which is what you changed in the PR! Thanks!

src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFeatureCountsSubCommandTestCase.java
```diff
@@ -0,0 +1,74 @@
...
+                throw new RuntimeException("Error when cleaning up", oops);
+            }
+        });
+        Files.deleteIfExists(temporaryPath);
+    }
+
+    @BeforeClass
+    public static void prepareAtlasesOnDisk() throws IOException
+    {
+        final Path path = Files.createTempDirectory("atlastest");
```
MikeGost: Can we use the File.temporaryFolder() here, so we are using our internal File code without switching between ours and nio?

src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFeatureCountsSubCommandTestCase.java
```diff
@@ -0,0 +1,68 @@
...
+import java.io.IOException;
+
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.Test;
+import org.openstreetmap.atlas.streaming.resource.File;
+
+import com.google.common.collect.Iterables;
```
MikeGost: Sorry, one more nit, let's use our `Iterables` class instead.

MikeGost:
Thanks Chris!

PR #65, opened 2018-01-23
Fixing Edge Definition Inconsistency

MikeGost:
Most important change here: From a previous PR (https://github.com/osmlab/atlas/pull/45) - tag criteria for what becomes an `Edge` when converting from PBF was updated. However, when building relations, we were looking at the Edge definition as provided by the `atlas-edge.json` file. Essentially - two definitions of `Edge` (see `TagMap` > `matchAtlasEdge` > `PbfOneWay` and the configuration file) This is an inconsistency that will be addressed by Part 3 of the PBF Ingest Refactor (see Raw Atlas PRs). However, in the meantime, the fix is to bring these two definitions in sync. This PR does just that and creates a test for `Relation` [4451979](https://www.openstreetmap.org/relation/4451979), which exhibits some of the complex tagging that was affected. 

Other changes include:
1. Cleaning up logging in `CountryBoundaryMap`. This has the potential to flood Spark logs with information about the time it took to slice each Way. Most slice operations take tens or at most hundreds of milliseconds and the logs would be overly verbose if we log each and every slice. Reducing the logging to only log in case of abnormal (greater than 1 minute) slicing time. 
2. Adding new `PublicServiceVehiclesTag` - encountered during investigation. It may prove to be relevant for those interested in specific road designation.
3. Fixing integration test that updates the members of a `Relation`. Specifically, Way [83642691](https://www.openstreetmap.org/way/83642691) became a bi-directional Edge instead of a Line. It was way-sectioned 3 times, and before had 3 Lines representing it, whereas now it has 6 edges (2 in each direction) representing it. The atlas excerpt:

Before:
```
[Relation: id=1251624000000, [Members: 
	{Member: ID = 83642691000001, Type = LINE, Role = }, 
	{Member: ID = 83642691000002, Type = LINE, Role = }, 
	{Member: ID = 83642691000003, Type = LINE, Role = }, ...
```

After:
```
[Relation: id=1251624000000, [Members: 
	{Member: ID = -83642691000003, Type = EDGE, Role = }, 
	{Member: ID = -83642691000002, Type = EDGE, Role = }, 
	{Member: ID = -83642691000001, Type = EDGE, Role = }, 
	{Member: ID = 83642691000001, Type = EDGE, Role = }, 
	{Member: ID = 83642691000002, Type = EDGE, Role = }, 
	{Member: ID = 83642691000003, Type = EDGE, Role = }, ...
```


src/main/resources/org/openstreetmap/atlas/geography/atlas/pbf/atlas-edge.json
```diff
@@ -1,6 +1,9 @@
 {
     "filters": [
-        "access->!no",
+        "access->!no|motor_vehicle->yes|motorcar->yes|vehicle->yes",
```
mertemin: It'd be nice to have a comment in `PbfOneWay`'s referencing this file.

MikeGost: Good call!

src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderIntegrationTest.java
```diff
@@ -228,7 +228,7 @@ public void testRoute()
         Assert.assertEquals("route", relation1.getTags().get("type"));
         Assert.assertEquals("bus", relation1.getTags().get("route"));
         Assert.assertEquals("Adult Tram", relation1.getTags().get("name"));
-        Assert.assertEquals(24, relation1.members().size());
+        Assert.assertEquals(27, relation1.members().size());
```
matthieun: It is always better to have a test that is not wrong ;)

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1027,7 +1028,10 @@ public int size()
             }
         }
 
-        logger.info("Took {} to slice way {}", time.untilNow(), identifier);
+        if (logger.isDebugEnabled() && time.untilNow().isMoreThan(Duration.ONE_MINUTE))
+        {
+            logger.info("Took {} to slice way {}", time.untilNow(), identifier);
```
matthieun: Might want to do `logger.debug` here.

MikeGost: Oops, had this and it changed back to info after fixing the branch, nice catch!

MikeGost:
@mertemin - updated slightly to remove some of the functionality you added in your last [PR](https://github.com/osmlab/atlas/pull/45/files#diff-30b1963a2c456670fed8cf1fbd1bb370R80). Take a look at this one more time please. cc. @matthieun for updates per our last chat.

PR #66, opened 2018-01-23
Remove fail-fast condition from areasIntersecting

mertemin:
This PR addresses one more issue with fail-fast condition in spatial queries (following https://github.com/osmlab/atlas/pull/62 from @MikeGost). In the case of `areasIntersecting` if given polygon was a rectangle, then results were returned without further filtering. However, spatial index is returning results based on feature bounds. So actual feature geometries don't have to intersect.

Fail-fast condition is removed and a test case is created. Test uses the following two triangles.
screen shot 2018-01-23 at 2 05 42 pm

When you call `areasIntersecting` with orange triangle's bounds, the method was returning the purple triangle, which is a false positive. After the update, it doesn't return it.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/items/AtlasItemIntersectionTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/items/AtlasItemIntersectionTest.java
@@ -54,6 +54,22 @@ public class AtlasItemIntersectionTest
     }
 
     @Test
+    public void testAreasNoIntersectingPolygon()
+    {
+        final Atlas atlas = this.rule.getNoIntersectionAtlas();
+
+        final Polygon triangle = new Polygon(Location.forString("47.6263, -122.209198"),
+                Location.forString("47.628685, -122.209305"),
+                Location.forString("47.628704, -122.211761"));
+
+        Assert.assertEquals("There should be no intersecting area", 0,
+                Iterables.size(atlas.areasIntersecting(triangle)));
+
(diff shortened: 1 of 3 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/items/AtlasItemIntersectionTest.java
```diff
@@ -53,6 +53,22 @@ public void testAreasIntersectingPolygon()
...
+
+        final Polygon triangle = new Polygon(Location.forString("47.6263, -122.209198"),
+                Location.forString("47.628685, -122.209305"),
+                Location.forString("47.628704, -122.211761"));
+
+        Assert.assertEquals("There should be no intersecting area", 0,
+                Iterables.size(atlas.areasIntersecting(triangle)));
+
+        Assert.assertEquals("There should be no intersecting area", 0,
+                Iterables.size(atlas.areasIntersecting(triangle.bounds())));
```
matthieun: If I follow your diagram correctly, shouldn't this return an area? The triangle does not intersect anything, but it seems that its bounds do.

matthieun: Oh. Is that initial `triangle` the orange one?

mertemin: Yes

PR #67, opened 2018-01-24
Raw Atlas Improvements

MikeGost:
Couple of changes here:
1. Creating grid index (JTS `STRtree`) for improved performance when slicing. This boosted performance quite a bit, allowing  the previously ignored integration test to be enabled.
**Before**: `Took 11.611 minutes to generate raw Atlas`
**After**: `Took 2.144 minutes to generate raw Atlas`
2. Some unit tests ran into an interesting corner case. When slicing was run with grid index being built, we were receiving duplicate `LineString`s when clipping a `Polygon` to a boundary (see CountryBoundaryMap#clipBoundary). This is because the feature was large enough that it spanned multiple index cells, and the returned multiple of the same polygons to slice against, hence we needed to de-dupe the returned `LineString`s. 
3. Enforced an alphabetical ordering of country codes for Lines and Points spanning the boundary. This creates more deterministic slicing across all features. Updated tests accordingly.
4. Updated the `RawAtlasCountrySlicer` to remove any intermediate slicing artifacts from being member variables. These are only required in the context of a slice and don't need to persist once the final atlas is built. This is also a cleaner design since it allows the class to be reused across multiple slice operations, without having to create a new one or rebuild the index each time.
5. Cleaned up some tests, verbose logging and better stream usage.


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -744,11 +745,15 @@ private void patchNonClosedMembers(final Relation relation, final List<RelationM
                         return;
                     }
 
-                    final List<LineString> borderLines;
+                    final Set<LineString> borderLines;
                     try
                     {
-                        borderLines = getCountryBoundaryMap().clipBoundary(relationIdentifier,
-                                polygon);
+                        // The intent is to have the all the unique slices in the same order that
```
matthieun: `is to have the all the unique`: `the` is repeated twice

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfMemoryStore.java
```diff
@@ -769,9 +765,6 @@ private boolean nodeHasExplicitOsmTags(final Node node)
         }
         else if (nodeTagSize < AtlasTag.TAGS_FROM_OSM.size())
         {
-            logger.error(
```
AlexHsieh22: was this removed because it would never hit? Fine with the removal in general.

MikeGost: Removed because it was hit in the case of new node creation, which wasn't padded with the new tags yet, so a lot of junk can possibly appear in the logs.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -48,14 +80,35 @@ public RawAtlasCountrySlicer(final Set<IsoCountry> countries,
...
+                slicedRelationChanges, newPointCoordinates);
         return relationSlicer.slice();
     }
+
+    /**
+     * Creates the grid index, for optimal country-slicing performance
+     */
+    private void initializeGridIndex()
+    {
+        if (this.multiPolygon != null && this.countryBoundaryMap != null)
```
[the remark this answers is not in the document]
MikeGost: Good question. It doesn't take into account whether the CountryBoundaryMap has already had a created grid index. I can't think of a case where that would happen. Do you think it's worth keeping state and tracking?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
MikeGost: Thanks, will remove.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -204,8 +207,16 @@ public RawAtlasSlicer(final Set<IsoCountry> countries,
         final CountryCodeProperties countryDetails = getCountryBoundaryMap()
                 .getCountryCodeISO3(location);
 
-        // Store the country code
-        tags.put(ISOCountryTag.KEY, countryDetails.getIso3CountryCode());
+        // Store the country code, enforce alphabetical order if there are multiple
+        if (countryDetails.inMultipleCountries())
+        {
+            tags.put(ISOCountryTag.KEY, Joiner.on(",").join(Sets.newTreeSet(Arrays.asList(
```
snb3300: Quick Questions: 

1. Have you seen country codes getting duplicated before or its just an extra piece of validation before we add the tag?
2. What kind of gains do we get from ordering the country codes? As adding to tree set from an ArrayList will still be  ```O(n * log(n))```.

MikeGost: Great questions! 

1. No duplication. For synthetic nodes created for features that span the boundary, we will have two or more country codes. Same goes for polygons that are cut across a boundary. The new line that will connect them is on top of the country boundary and will have two country codes.
2. Key gain isn't performance, it's consistency and deterministic ordering. This ensures that running slicing will give us the same LineString with the same metadata each time. I figured this is pretty important in debugging various issues as they come up. Testing this, I've only seen a maximum of 3 country codes at an intersection where there are 3 countries, so this has pretty low performance impact.

snb3300: Thanks for the explanation. Having deterministic ordering makes more sense now. I kind of suspected that the country list won't be that long to have a performance impact but just wanted to be sure for my understanding. This makes much more sense  now

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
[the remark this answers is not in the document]
matthieun: So it would make sense to keep using the grid index that ships with the boundary map object, if it is already there.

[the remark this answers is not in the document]
MikeGost: The chances of us re-creating the grid index in existing flow are slim to none. Furthermore, the existing Atlas generation flow makes a very similar grid index generation call, so it is not a new issue. I agree with you though, that we can add a simple check in the CountryBoundaryMap class to not build grid index if it's already built. We have a task coming up in the near future that has some grid index work. I will add this request in that task, as that fits in more nicely with the refactor that will be done. Sound good?

PR #68, opened 2018-01-26
Ingest building min_height OSM tag

arpanaj:
If a building does not start from ground level, then OSM min_height tag is used to indicate the height of bottom part of the building. Adding method to read this tag and set it in the Building entity.


src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,17 @@
+package org.openstreetmap.atlas.tags;
+
+import org.openstreetmap.atlas.tags.annotations.Tag;
+import org.openstreetmap.atlas.tags.annotations.Tag.Validation;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+
+/**
+ * OSM min_height tag: https://taginfo.openstreetmap.org/keys/min_height#values
```
MikeGost: Nit: the values link can be removed, it's already available in the taginfo parameter below :)

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * @author ajayaswal
+ */
+@Tag(value = Validation.LENGTH, taginfo = "https://taginfo.openstreetmap.org/keys/min_height#values", osm = "https://wiki.openstreetmap.org/wiki/Key:min_height")
```
MikeGost: Is `LENGTH` validator the correct one to use here? TagInfo shows a lot of negative values for this tag (-3, -6, etc.) If we want to consider these, maybe `DOUBLE` is the correct one to use.

MikeGost: Also, we may want to add a few unit test cases for this to reflect OSM data (negative values, 0, positives values, decimal values to make sure they're handled properly)

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/ComplexBuilding.java
```diff
@@ -58,6 +59,34 @@ protected ComplexBuilding(final AtlasEntity source)
...
+            minHeightTag = tags.get(MinHeightTag.KEY);
+            if (minHeightTag != null)
+            {
+                return Optional.of(HEIGHT_CONVERTER.convert(minHeightTag));
+            }
+        }
+        catch (final Exception e)
+        {
+            logger.warn("Invalid base height {} for building id {}", minHeightTag,
+                    getSource().getIdentifier());
```
MikeGost: Do we care about keeping the exception in the log?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/ComplexBuilding.java
```diff
@@ -58,6 +59,34 @@ protected ComplexBuilding(final AtlasEntity source)
...
+    {
+        Map<String, String> tags = getSource().getTags();
+        String minHeightTag = tags.get(MinHeightTag.KEY);
+        try
+        {
+            if (minHeightTag != null)
+            {
+                return Optional.of(HEIGHT_CONVERTER.convert(minHeightTag));
+            }
+            tags = this.outlineSource.getTags();
```
MikeGost: Consider using `Validators.hasValuesFor(...)` here instead of grabbing allTags - this can be somewhat expensive.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
matthieun: Oh also it might be a problem using `Distance` which cannot be negative...

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/ComplexBuilding.java
```diff
@@ -58,6 +59,34 @@ protected ComplexBuilding(final AtlasEntity source)
...
+     */
+    public Optional<Distance> baseHeight()
+    {
+        Map<String, String> tags = getSource().getTags();
+        String minHeightTag = tags.get(MinHeightTag.KEY);
+        try
+        {
+            if (minHeightTag != null)
+            {
+                return Optional.of(HEIGHT_CONVERTER.convert(minHeightTag));
```
AlexHsieh22: Would it be possible to create a "getMinimumHeight" function that does:
```
try
{
    String minHeight = tags.get(MinHeightTag.KEY);

    if (minHeightTag != null)
    {
        return Optional.of(HEIGHT_CONVERTER.convert(minHeightTag));
    }
}
catch(...)
{
...
}
```

Then we could make 2 separate calls here and if the first call throw an exception, the second one can still be checked!

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
AlexHsieh22: nit: Maybe a href for the link?

arpanaj: Removed the ink, keeping the one in the tag info parameter.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
arpanaj: Good point, will use a DoubleValidator and add more test cases. Will replace Distance with a new Height scalar.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/ComplexBuilding.java
arpanaj: Will cleanup this code, looks like its more relevant for height, not for min_height.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,38 @@
...
+    static Optional<Height> get(final Taggable taggable)
+    {
+        final Optional<String> tagValue = taggable.getTag(KEY);
+
+        if (tagValue.isPresent())
+        {
+            if (validator.isValid(tagValue.get()))
+            {
+                return Optional.of(new Height(Double.valueOf(tagValue.get())));
+            }
```
MikeGost: Nit: possible improvement could be combining if statements:
`if (tagValue.isPresent() && validator.isValid(tagValue.get())) {...}`

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
```diff
@@ -0,0 +1,45 @@
+package org.openstreetmap.atlas.utilities.scalars;
+
+import java.io.Serializable;
+
+/**
+ * Distance wrapper to denote height above or below ground
+ *
+ * @author ajayaswal
+ */
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
```diff
@@ -0,0 +1,45 @@
...
+    private final Distance distance;
+    private final boolean aboveGround;
+
+    public Height()
+    {
+        this.distance = Distance.ZERO;
+        this.aboveGround = true;
+    }
+
+    public Height(final Double rawHeight)
```
MikeGost: double instead of Double?

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
```diff
@@ -0,0 +1,45 @@
...
+    public double getAbsoluteHeightAsMeters()
+    {
+        return this.distance.asMeters();
+    }
+
+    public boolean isAboveGround()
+    {
+        return this.aboveGround;
+    }
+
```
AlexHsieh22: would a ```getHeighAsMeters()``` method be useful here?

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
```diff
@@ -0,0 +1,45 @@
...
+    @Override
+    public String toString()
+    {
+        return String.format("%.1f meters ", this.distance, this.aboveGround ? "" : "underground");
```
AlexHsieh22: are we missing a %s in the string format?

MikeGost: Slim chance of this happening, but it looks like distance can be printed in kilometers in some cases, do we want to rely on the toString for distance to dictate units? Also, do we need a second string.format parameter for the "" or "underground" string?

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
```diff
@@ -0,0 +1,45 @@
...
+
+import java.io.Serializable;
+
+/**
+ * Distance wrapper to denote height above or below ground
+ *
+ * @author ajayaswal
+ */
+
+public class Height implements Serializable
```
AlexHsieh22: Thinking about this class a bit more, and after a brief discussion with @MikeGost , could we potentially use or extend the ```Altitude``` class?

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
arpanaj: Removing this class, existing Altitude class works.

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
arpanaj: Removing this class, existing Altitude class works.

src/main/java/org/openstreetmap/atlas/utilities/scalars/Height.java
arpanaj: Thanks, Altitude class works well here, removing Height.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,35 @@
...
+import org.openstreetmap.atlas.tags.annotations.Tag.Validation;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+import org.openstreetmap.atlas.tags.annotations.validation.DoubleValidator;
+
+/**
+ * OSM min_height tag
+ *
+ * @author ajayaswal
+ */
+@Tag(value = Validation.LENGTH, taginfo = "https://taginfo.openstreetmap.org/keys/min_height#values", osm = "https://wiki.openstreetmap.org/wiki/Key:min_height")
```
MikeGost: Still think we should change this to Validation.DOUBLE - since `LengthValidator` would fail on a large number of values.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,35 @@
...
+
+    DoubleValidator validator = new DoubleValidator();
+
+    static Optional<Altitude> get(final Taggable taggable)
+    {
+        final Optional<String> tagValue = taggable.getTag(KEY);
+
+        if (tagValue.isPresent() && validator.isValid(tagValue.get()))
+        {
+            return Optional.of(Altitude.meters(Double.valueOf(tagValue.get())));
```
MikeGost: One minor thing to consider, there's a pattern of `validateAndExtract` found in other tag classes that contain numerical values (see `LayerTag` and `MaxSpeedForwardTag`). As an optional improvement, we should consider adding a `DoubleExtractor` and following the same pattern. This also works though!

src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/BuildingHeightTestCaseRule.java
```diff
@@ -0,0 +1,69 @@
...
+ * Test fixture for the {@link BuildingHeightTestCase}
+ *
+ * @author ajayaswal
+ */
+public class BuildingHeightTestCaseRule extends CoreTestRule
+{
+    public static final long BUILDING_WITH_ZERO_MINHEIGHT_ID = 100000000L;
+    public static final long BUILDING_WITH_NEGATIVE_MINHEIGHT_ID = 200000000L;
+    public static final long BUILDING_WITH_BOTH_HEIGHT_ID = 300000000L;
+    public static final long BUILDING_WITH_NO_MINHEIGHTT_ID = 400000000L;
```
MikeGost: Nit: two `TT_ID` --> `T_ID`. Also, consider adding `_` between `MIN` and `HEIGHT` for readability.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
arpanaj: Agree, missed in the previous checkin, will fix now.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
arpanaj: Since the extraction code is very minimal, thinking of leaving it as is, instead of creating a new class.

MikeGost: Sounds good!

arpanaj:
@matthieun , added a separate MinHeightTag class, all the tag value validations for MaxHeightTag do not apply to MinHeight.

src/main/java/org/openstreetmap/atlas/tags/MinHeightTag.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author ajayaswal
+ */
+@Tag(value = Validation.DOUBLE, taginfo = "https://taginfo.openstreetmap.org/keys/min_height#values", osm = "https://wiki.openstreetmap.org/wiki/Key:min_height")
+public interface MinHeightTag
+{
+    @TagKey
+    String KEY = "min_height";
+
+    DoubleValidator validator = new DoubleValidator();
```
snb3300: Because this is getting used only in a static method below do you think it will be a good idea to have it as
``` private static final DoubleValidator doubleValidator = new DoubleValidator()``` ?

arpanaj: Its an interface, so can't do private.

PR #70, opened 2018-01-31
Adding test for bridge in water case

MikeGost:
Two changes here:

1. Adding a test case that maintains parity with the previous slicing logic for bridges crossing water.
2. Updating the TestAtlasHandler to refer loading from a generic osm resource. The boolean flag will identify whether it's a JOSM-based file or not, the method name didn't make sense before.

### DIFF
--- a/src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
+++ b/src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
@@ -5,17 +5,22 @@ import java.util.Set;
 
 import org.junit.Assert;
 import org.junit.Test;
+import org.openstreetmap.atlas.geography.MultiPolygon;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.items.Edge;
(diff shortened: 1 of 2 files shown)


src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
```diff
@@ -99,4 +104,31 @@ public void testPbfToSlicedRawAtlas()
...
+        final Atlas rawAtlas = rawAtlasGenerator.build();
+
+        final Set<IsoCountry> countries = new HashSet<>();
+        countries.add(IsoCountry.forCountryCode("CUB").get());
+        final RawAtlasCountrySlicer slicer = new RawAtlasCountrySlicer(countries, map, loadingArea);
+
+        final Atlas slicedAtlas = slicer.slice(rawAtlas);
+
+        // Make sure that the big bridge over water made it to the Atlas
+        Assert.assertNotNull(slicedAtlas.line(308541861000000L));
```
matthieun: I believe that way is already inside the CUB boundary supplied for the test case, which will succeed, no matter if the code handles outside ways or not.

MikeGost:
On second thought, it makes more sense to isolate this to a single test case that will test various boundary scenarios. Closing for now and reopening when updated.

PR #71, opened 2018-02-01
Remove false positive existing boundary tag

matthieun:
In some cases, the `EXISTING` synthetic boundary node tag was applied to some nodes inside the country boundary, even though it should be on nodes only outside or on the country boundary.

This was caused in corner cases where the `OsmPbfLoader`'s `loadingArea` is smaller than the full country boundary, and the pbf file being loaded is larger than the loading area and contains a way that seems fully outside the loading area but is still inside the country boundary.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
@@ -32,8 +32,11 @@ import org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier.PaddingIde
 import org.openstreetmap.atlas.geography.atlas.pbf.store.PbfMemoryStore;
 import org.openstreetmap.atlas.geography.atlas.pbf.store.PbfOneWay;
 import org.openstreetmap.atlas.geography.atlas.pbf.store.TagMap;
+import org.openstreetmap.atlas.geography.atlas.raw.slicing.CountryCodeProperties;
 import org.openstreetmap.atlas.geography.converters.jts.JtsMultiPolygonToMultiPolygonConverter;
+import org.openstreetmap.atlas.geography.converters.jts.JtsPointConverter;
 import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.ISOCountryTag;
 import org.openstreetmap.atlas.tags.LastEditChangesetTag;
 import org.openstreetmap.atlas.tags.LastEditTimeTag;
 import org.openstreetmap.atlas.tags.LastEditUserIdentifierTag;
@@ -57,6 +60,8 @@ import org.openstreetmap.osmosis.core.task.v0_6.Sink;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.vividsolutions.jts.geom.Geometry;
+
 /**
  * This {@link OsmPbfProcessor} loads four different OSM entities (Node, Way, Relation and Bound),
  * and build a complete atlas with a multipolygon or bounding box. 'Complete' means if part of an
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
```diff
@@ -625,6 +631,23 @@ private boolean isHighwayOrFerry(final Way way)
...
+     *         neighbor logic.
+     */
+    private boolean isOutsideBoundary(final Node node)
+    {
+        final Geometry nodeLocation = JTS_POINT_CONVERTER.convert(new Location(
+                Latitude.degrees(node.getLatitude()), Longitude.degrees(node.getLongitude())));
+        final CountryCodeProperties countryCodeProperties = this.loadingOption
+                .getCountryBoundaryMap().getCountryCodeISO3(nodeLocation, false);
+        return ISOCountryTag.COUNTRY_MISSING.equals(countryCodeProperties.getIso3CountryCode())
+                || countryCodeProperties.usingNearestNeighbor();
```
snb3300: I am not sure if this is the right place to ask question about nearest neighbor logic but how does it work? 
Is it something like 

if current node does not have ISO country tag then 
get the country code for the nearest neighbor and assign it to current node?

src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderTest.java
```diff
@@ -101,6 +107,29 @@ public void init()
                 Rectangle.forLocated(line4, line5)));
     }
 
+    /**
+     * In some cases, ways can be outside the loading area (still inside the osm PBF) and connected
+     * to some other ways that are inside the loading area, but still inside the country boundary.
+     * That happens at shard boundaries, which do a soft cut.
+     */
+    @Test
+    public void testBoundaryNodesForWayOutsideLoadingAreaButInsideCountryBoundary()
```
snb3300: Nice Test Case 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
MikeGost: @snb3300 - that's right. If a feature is close to a country, but not quite within the border, then we will try to find the nearest country and assign its country code. We will add the nearest neighbor tag to make this case a clear distinction.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
```diff
@@ -904,7 +927,7 @@ else if (entity instanceof Relation)
 
     private void tagOutsideBoundaryNodes(final Node node)
     {
-        if (this.nodeIdentifiersAtNetworkBoundary.contains(node.getId()))
+        if (this.nodeIdentifiersAtNetworkBoundary.contains(node.getId()) && isOutsideBoundary(node))
```
AlexHsieh22: what is the expectation for a node that falls directly on a boundary during country slicing?  Will it get a country code for the country it is on the boundary of?  Will nearest neighbor be false?  If that's the case, will a SyntheticBoundaryNodeTag get applied since the condition is that the node must be outside the boundary?

MikeGost: For a `Node` directly on the boundary, it will get a country code for all countries it is on the boundary of. So it will have 2 or more country codes. Nearest neighbor will be `false` and `SyntheticBoundaryNodeTag` will be applied [here](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java#L940) or [here](https://github.com/osmlab/atlas/blob/8d6defc17ae02e318fe7c68a4209af5d197c817e/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java#L969), depending on the case (if it is an existing node or a new node created by country slicing a way crossing the boundary).

AlexHsieh22: Ah, nice! Thanks @MikeGost .

MikeGost: Forgot to mention, the first case that's linked above, where the existing OSM Node is on the boundary, is not handled in today's flow (see `if false` statement). It is handled in the Raw Atlas refactor [flow](https://github.com/osmlab/atlas/blob/a46f9baf6acd0240731468b89e2a87bb5925486d/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java#L237).

PR #73, opened 2018-02-01
Add utility isApproximatelyNSided method to Polygon

mertemin:
This PR adds a utility method called `isApproximatelyTriangle` to `Polygon` class. Given a threshold, method goes over the segments and the heading differences between segments. If number of heading changes is equal to 2, that means shape is triangular like. Higher thresholds can treat square like shapes as triangulars as well (expected though). See unit tests for examples.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
@@ -303,7 +303,7 @@ public class Polygon extends PolyLine
         // Make sure we start with some heading
         if (!previousHeading.isPresent())
         {
-            logger.trace("{} doesn't have a heading to calculate number of sides.");
+            logger.trace("{} doesn't have a heading to calculate number of sides.", this);
             return false;
         }
 
--- a/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
@@ -18,6 +18,8 @@ import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.MultiIterable;
 import org.openstreetmap.atlas.utilities.scalars.Angle;
 import org.openstreetmap.atlas.utilities.scalars.Surface;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Lists;
 import com.vividsolutions.jts.geom.GeometryCollection;
@@ -50,6 +52,8 @@ public class Polygon extends PolyLine
             Location.forString("37.390234491673446,-122.03111171722412"));
 
     private static final JtsPolygonConverter JTS_POLYGON_CONVERTER = new JtsPolygonConverter();
+
+    private static final Logger logger = LoggerFactory.getLogger(Polygon.class);
     private static final long serialVersionUID = 2877026648358594354L;
 
     // Calculate sides starting from triangles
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -248,6 +256,69 @@ public Location interiorCenter()
...
 
+    /**
+     * @param threshold
+     *            {@link Angle} threshold that decides whether a {@link Heading} difference between
+     *            segments should be counted towards heading change count or not
+     * @return true if this {@link Polygon} has an approximately triangular shape while ignoring
+     *         {@link Heading} differences between inner segments that are below given threshold. A
+     *         {@link Polygon} might have more than 3 shape points with slight heading differences.
+     */
+    public boolean isApproximatelyTriangle(final Angle threshold)
```
mertemin: This could possibly be updated to handle more than triangles and other shapes by giving the number of sides as a parameter.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -248,6 +253,73 @@ public Location interiorCenter()
...
+    /**
+     * @param nCount
+     *            Number of sides
+     * @param threshold
+     *            {@link Angle} threshold that decides whether a {@link Heading} difference between
+     *            segments should be counted towards heading change count or not
+     * @return true if this {@link Polygon} has approximately n sides while ignoring {@link Heading}
+     *         differences between inner segments that are below given threshold.
+     */
+    public boolean isApproximatelyNSided(final int nCount, final Angle threshold)
```
matthieun: How about changing `nCount` to `expectedNumberOfSides` or something similar to make it more explicit? The same way you have `expectedHeadingChangeCount` below, which is really easy to read.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -248,6 +253,73 @@ public Location interiorCenter()
...
+            if (previousHeading.isPresent())
+            {
+                break;
+            }
+        }
+
+        // Make sure we start with some heading
+        if (!previousHeading.isPresent())
+        {
+            return false;
```
matthieun: Maybe have a TRACE log here to say what polygon was rejected?

src/test/java/org/openstreetmap/atlas/geography/PolygonTest.java
```diff
@@ -223,6 +224,196 @@ public void testNoAnglesMatchingOurCriteria()
         Assert.assertTrue(result.size() == 0);
     }
 
+    @Test
+    public void testNSidedness()
+    {
+        // Test point (has no sides)
```
matthieun: Really nice set of tests! 

- How about adding one more where the first one or two segments are points, and then the rest are not?
- How about one where the polygon is self intersecting? To make sure that the method does not throw exceptions?

Finally if you don't mind, it would be nice to break each use case in a different test method; that would make debugging easier down the line.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -248,6 +257,75 @@ public Location interiorCenter()
...
+            if (previousHeading.isPresent())
+            {
+                break;
+            }
+        }
+
+        // Make sure we start with some heading
+        if (!previousHeading.isPresent())
+        {
+            logger.trace("{} doesn't have a heading to calculate number of sides.");
```
matthieun: Simple omission here: you need to add what will be replaced within `{}`.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -248,6 +257,75 @@ public Location interiorCenter()
...
+        final List<Segment> segments = this.segments();
+        final int segmentSize = segments.size();
+
+        // Index to keep track of segment to work on
+        int segmentIndex = 0;
+
+        // Keep track of heading changes
+        int headingChangeCount = 0;
+
+        // Find initial heading
```
AlexHsieh22: Is this a candidate for a helper method (not sure where it'd go, maybe a static method in Segment?)?  Finding the initial heading from a list of segments?

mertemin: I'd note this and extract it out when there are more clients needing this functionality.

AlexHsieh22: 👍

PR #74, opened 2018-02-02
 Restrain DynamicAtlas in case of initial polygon

matthieun:
In DynamicAtlas, when the initial shards are supplied with a Polygon or MultiPolygon, and the expansion policy is withExtendIndefinitely=false, then the expansion needs not consider the full initial shards for intersection, but the initial polygon.

This PR fixes that issue. Also added one test.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasRestrainedExpansionWithPolygonTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasRestrainedExpansionWithPolygonTest.java
@@ -76,7 +76,7 @@ public class DynamicAtlasRestrainedExpansionWithPolygonTest
     }
 
     @Test
-    public void testNotMovingTooFast()
+    public void testRestrainedExpansionWithPolygon()
     {
         Assert.assertEquals(2, this.dynamicAtlas.numberOfEdges());
     }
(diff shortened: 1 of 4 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasRestrainedExpansionWithPolygonTest.java
```diff
@@ -0,0 +1,83 @@
...
+        this.store.put(new SlippyTile(6, 34, 6),
+                this.rule.getAtlas().subAtlas(new SlippyTile(6, 34, 6).bounds()).get());
+        this.store.put(new SlippyTile(4, 34, 6),
+                this.rule.getAtlas().subAtlas(new SlippyTile(4, 34, 6).bounds()).get());
+        this.dynamicAtlas = new DynamicAtlas(policy);
+        this.dynamicAtlas.preemptiveLoad();
+    }
+
+    @Test
+    public void testNotMovingTooFast()
```
MikeGost: Nit: I'm probably not familiar with the terminology, the test name is a little hard to follow. It sounds like we're testing DynamicAtlas expansion based off the features that extend out of the starting (Multi)Polygon instead of features that extend out of all the shards that intersect with the starting Polygon. Maybe rename it to "testFeatureExpansionWithPolygon" or "testRestrainedExpansion"?

matthieun: Oh! That is a copy paste error! Thanks!

PR #75, opened 2018-02-03
Fixing OptionalIterable Bug

jklamer:
OptionalIterable's iterator's hasNext() method called next() on the underlying iterable, such that multiple hasNext Method calls in a row would result in lost item(s) depending on how many hasNext() calls were made. Added multiple hasNext() calls to the test and used the cached item as a hasNext() indicator.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/collections/OptionalIterable.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/collections/OptionalIterable.java
@@ -9,6 +9,7 @@ import java.util.Optional;
  * @param <T>
  *            The type of the {@link Optional} iterator
  * @author cuthbertm
+ * @author jklamer
  */
 public class OptionalIterable<T> implements Iterable<T>
 {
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/utilities/collections/OptionalIterableTest.java
```diff
@@ -31,6 +31,8 @@ public void iterationTest()
         Assert.assertTrue(iterator.hasNext());
         Assert.assertEquals(Integer.valueOf(1), iterator.next());
```
MikeGost: Since we're updating tests, do you mind changing `optIt` here to be non-abbreviated to follow general repo conventions? :)

src/main/java/org/openstreetmap/atlas/utilities/collections/OptionalIterable.java
```diff
@@ -31,17 +32,24 @@ public OptionalIterable(final Iterable<Optional<T>> iterable)
...
                     {
-                        this.previousElement = current;
-                        return true;
+                        final Optional<T> current = this.iterator.next();
+                        if (current.isPresent())
+                        {
+                            this.previousElement = current;
+                            return true;
+                        }
+                        this.previousElement = Optional.empty();
```
MikeGost: Nit: Do we need to set `previousElement` to empty here? It's already initialized to empty and if it's not present, we'd just be setting it again.

jklamer: logically this assignment is pointless so I'll remove it.

src/test/java/org/openstreetmap/atlas/utilities/collections/OptionalIterableTest.java
jklamer: Done

PR #76, opened 2018-02-03
Changed AWT Polygon to JTS and added unit test

jwpgage:
This change was necessary because the size of some bounds were causing an int overflow when using the awt Polygon.

An existing unit test was adjusted due to the difference in definitions of containment between awt and jts. Here is the definition of contains (insideness) of awt point:

A point is considered to lie inside a Shape if and only if: it lies completely inside the Shape boundary or it lies exactly on the Shape boundary and the space immediately adjacent to the point in the increasing X direction is entirely inside the boundary or it lies exactly on a horizontal boundary segment and the space immediately adjacent to the point in the increasing Y direction is inside the boundary.

The definition in JTS is much more simple, `polygon.covers(point)` considers a point to be covered if it is within the border of the polygon, or lies anywhere on the boundary. This definition seems to make more sense.


src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -162,7 +160,9 @@ public Location center()
      */
     public boolean fullyGeometricallyEncloses(final Location location)
     {
-        return awtPolygon().contains(location.asAwtPoint());
+        final com.vividsolutions.jts.geom.Polygon polygon = new JtsPolygonConverter().convert(this);
```
MikeGost: It looks like we have a static converter already initialized above. We can use it instead of creating a new one.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -207,7 +207,8 @@ public boolean fullyGeometricallyEncloses(final Rectangle rectangle)
             return false;
         }
         // The item is within the bounds of this Polygon
-        return awtArea().contains(rectangle.asAwtRectangle());
+        final com.vividsolutions.jts.geom.Polygon polygon = new JtsPolygonConverter().convert(this);
```
MikeGost: Same comment here about using the static converter for both of these calls.

MikeGost: Nit: comment needs an update, references awt definition.

src/test/java/org/openstreetmap/atlas/geography/PolygonTest.java
```diff
@@ -89,7 +113,7 @@ public void testConcave()
 
         Assert.assertTrue(concave.fullyGeometricallyEncloses(inside1));
         Assert.assertTrue(concave.fullyGeometricallyEncloses(onBoundary1));
-        Assert.assertFalse(concave.fullyGeometricallyEncloses(onBoundary2));
+        Assert.assertTrue(concave.fullyGeometricallyEncloses(onBoundary2));
```
MikeGost: As you said in the PR description, this makes a lot more sense!

MikeGost:
Since you added the osm files, what do you think about loading atlas files from them to test the modified calls, instead of creating the Polygons inline during the test. Might be a bit cleaner. This was if someone has to modify, they can modify in a single location.

jwpgage:
@MikeGost As per a discussion with @matthieun if the .osm files are used directly, then the pbf creation would be in the loop. Loading from wkt makes the tests more lightweight, but definitely will make them a bit harder to change in future if necessary.

MikeGost:
@jwpgage - that's true, PBF is in play at that point. I'd argue that's not a huge issue and would help us detect any end to end changes. As a compromise, if we are going to keep two separate representations, maybe put a note in the test javadoc saying that any changes to these tests require a tweak to the osm file as well :)

matthieun:
Good point, that comment would be useful!

jwpgage:
Closing due to performance concerns raised by integration tests. Will be reopened after issue is investigated.

PR #77, opened 2018-02-06
Keyword Configuration Overrides

jklamer:
This PR adds functionality to the configuration framework by allowing flexible single or multiple element overriding per keyword in the same file.

To implement: construct a configuration the normal way. Then for every keyword specific configurable operation, call `Configuration::configurationForKeyword(String)`. This returns a configuration with a lightweight keyword specific configuration as a configuration layer over the original in a `MergedConfiguration`.

To do this, it looks at every key in the JSON other than 'override' and looks in the same context for data with key/data path of "override.key". If override data is present it populates a `configurationData` map that is then used to construct a `StandardConfiguration` for the top layer in the returned `MergedConfiguration`.

Downside: 'override' is no longer a viable configuration key.

Example Configurations:
This configuration will override the feature.range.max value for the Keyword ABC

```
{
  "feature.range": {
    "max": 35.0,
    "override.ABC.max":70.0
  }
}
```
This configuration will overwrite the feature.list for ABC and feature.range.max for XYZ.

```
{
  "feature":{
    "enable":true,
    "list":[
      "nodes",
      "edges"
    ],
    "range":{
      "max":20.0,
      "min":10.0
    },
    "override": {
      "XYZ":{
        "range":{
          "max": 30.0
        }
      },
      "ABC.list":["nodes"]
    }
  }
}
```

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
@@ -1,7 +1,9 @@
 package org.openstreetmap.atlas.utilities.configuration;
 
 import java.util.Arrays;
+import java.util.HashMap;
 import java.util.LinkedList;
+import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Function;
@@ -23,6 +25,7 @@ import com.google.common.base.Joiner;
  *
  * @author cstaylor
  * @author brian_l_davis
+ * @author jklamer
  */
 public class StandardConfiguration implements Configuration
 {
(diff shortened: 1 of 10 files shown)


src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
```diff
@@ -75,6 +78,7 @@ private StandardConfigurable(final String key, final Raw defaultValue,
         }
     }
 
+    private static final String OVERRIDE_STRING = "override";
```
MikeGost: Maybe a quick note that this can no longer be used as a key. It says that in the PR description, might also help users in the code base.

src/test/java/org/openstreetmap/atlas/utilities/configuration/MergedConfigurationTest.java
```diff
@@ -57,15 +63,54 @@ public void testLayered() throws IOException
...
 
             Assert.assertFalse(configuration.get("foo").valueOption().isPresent());
         }
     }
 
+    @Test
+    public void testMergedOverriddenConfigurations() throws IOException
+    {
+        final ClassLoader loader = Thread.currentThread().getContextClassLoader();
+        try (InputStream countryOverriddenBaseConfiguration = loader
```
MikeGost: Nit; countryOverriddenBaseConfiguration --> keywordOverridenBaseConfiguration

PR #78, opened 2018-02-13
Raw Atlas Way Sectioning

MikeGost:
There is still a lot of formatting, TO-DO's and minor things to fix. Wanted to get some eyes on the changes while it was still fresh and easy to modify!

Part 3 of issue #17 

**Summary of changes:**
1. All Raw Atlas `Point`s at the same `Location` are consolidated into one `Point` with a newly created `SyntheticDuplicateOsmNodeTag` added to it. This tag represents bad OSM data that needs to be cleaned up. This drastically simplifies way-sectioning logic (avoiding choosing which nodes to use and which to ignore). To be deterministic, we always keep the `Point` with the lowest OSM identifier. This can result in some data loss if the `Point`(s) tossed contain valuable OSM tagging. However, an atlas-check becomes very trivial to implement, that would spot such cases.
2. Way-sectioning - first cut at way-sectioning. Was able to section an entire atlas shard with many complex features (roundabouts with/without intersections, isolated loops, reversed and one/two way roads, etc.)
3. Minor functionality improvements for `PolyLine`, `LineItem`, `StreamIterable` functionality - added ability to append, prepen, calculate the size of a line and have an `anyMatch` function
4. Tests - integration and unit testing for way-sectioning
5. Minor bug fixes along the way

**Remaining work items:**
1. Clean up some of the TO-DOs (namely the spatial index issue, `PolyLine` to `Polygon` conversion, etc.)
2. Consolidate logic for closed/open edge sectioning
3. Finish unit testing specific cases
4. Add `DynamicAtlas` logic for sectioning across multiple shards (likely in a separate PR)
5. Performance and Memory usage statistics

**Testing artifacts:**
Sharing some interesting cases encountered during sectioning - these will be captured as unit tests.

Roundabout with a single intersection (node at intersection without further sectioning):
screen shot 2018-02-12 at 4 54 58 pm

Isolated closed loop (sectioned at start of `PolyLine` and at middle point):
screen shot 2018-02-12 at 4 54 23 pm

Standard roundabout (sectioned at each intersection):
screen shot 2018-02-12 at 4 53 51 pm

Multi-intersection roundabout (sectioned at each intersection):
screen shot 2018-02-12 at 4 52 37 pm


src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
```diff
@@ -43,6 +44,34 @@
...
+
+        final Atlas slicedRawAtlas = new RawAtlasCountrySlicer(COUNTRIES, COUNTRY_BOUNDARY_MAP)
+                .slice(rawAtlas);
+
+        final Atlas finalAtlas = new WaySectionProcessor(slicedRawAtlas,
+                AtlasLoadingOption.createOptionWithAllEnabled(COUNTRY_BOUNDARY_MAP)).run();
+
+        System.out.println(finalAtlas.summary());
+
+        finalAtlas.save(
```
MikeGost: Will be removed upon completion of testing.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -300,6 +296,120 @@ private void populateAtlasMetadata()
...
+     * <li>1. A {@link Point} was a shape point for an OSM {@link Way} that was removed. This point
+     * doesn't have any tags, isn't a part of a {@link Relation} and doesn't intersect with any
+     * other features in the Atlas.
+     * <li>2. A {@link Point} has multiple points at its {@link Location}. In this case, we sort all
+     * the Points, keep the one with the smallest identifier, add a
+     * {@link SyntheticDuplicateOsmNodeTag} and remove the rest of the duplicate Points. Note: we
+     * can potentially be tossing out OSM Nodes with non-empty tags. However, this is the most
+     * deterministic and simple way to handle this. The presence of the synthetic tag will make it
+     * easy to write an Atlas Check to resolve the data error.
+     * </ul>
```
matthieun: Very nice comments!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
```diff
@@ -0,0 +1,94 @@
...
+import java.util.Optional;
+import java.util.Set;
+
+import org.openstreetmap.atlas.geography.Location;
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
+import org.openstreetmap.atlas.geography.atlas.raw.slicing.temporary.TemporaryNode;
+
+/**
+ * Keeps track of all {@link TemporaryNode}s created for an atlas {@link Edge} and the occurrence of
+ * it in the underlying polyline.
```
matthieun: Maybe add a comment here saying this is specifically used for way-sectioning

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+     * files obtained by using the given fetcher policy. If the sharding tree and atlas fetcher
+     * function is not provided, then no expansion will be done and only the given atlas will be
+     * way-sectioned. Note: the sharding tree must the same as the one used to generate the input
+     * raw Atlas.
+     *
+     * @param rawAtlas
+     *            The raw {@link Atlas} to section
+     * @param loadingOption
+     *            The {@link AtlasLoadingOption} to use
+     * @param shardingTree
```
matthieun: Naming nit here. A sharding object does not have to be a tree. The most common one is the quad tree, but for example the `SlippyTileSharding` is just all tiles at the same zoom level. I would call it just `sharding`!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+     *            The {@link Sharding} tree to use to know which neighboring atlas files to get
+     * @param atlasFetcher
```
matthieun: I would call this one the `rawAtlasFetcher` to emphasize that what it needs to return is files at the rawAtlas level.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+     *
+     * @return the way-sectioned {@link Atlas}
+     */
+    public Atlas run()
+    {
+        final Time time = Time.now();
+
+        // Create a changeset to keep track of any intermediate state transitions (points becomes
+        // nodes, lines becoming areas, etc.)
+        final WaySectionChangeSet changeSet = new WaySectionChangeSet();
```
matthieun: I like the fact that all is encapsulated in this method!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+     * OSM forces closed polygons. If we have only 3 nodes (a -> b -> c), then it's not a polygon.
+     * If we have (a -> b -> a), then we have a flat line. At minimum, we are looking for (a -> b ->
+     * c -> a).
+     *
+     * @param line
+     *            The {@link Line} to check
+     * @return {@code true} if given {@link Line} contains more than
+     *         {@link #MINIMUM_SHAPE_POINTS_TO_QUALIFY_AS_AREA} shape points
+     */
+    private boolean containsMoreThanThreeShapePoints(final Line line)
```
matthieun: I would call this `qualifiesAsArea`...

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+     * must not contain highway tags, be a ferry, pier or railway.
+     *
+     * @param line
+     *            The {@link Line} to test
+     * @return {@code true} if the given {@link Line} qualifies as an {@link Area}
+     */
+    private boolean isAtlasArea(final Line line)
+    {
+        return line.isClosed() && containsMoreThanThreeShapePoints(line)
+                && !Validators.hasValuesFor(line, HighwayTag.class) && !RouteTag.isFerry(line)
```
matthieun: It would be nice to create an areaFilter for the tag part of that statement.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+    {
+        return this.loadingOption.getWaySectionFilter().test(point);
+    }
+
+    /**
+     * TODO
+     *
+     * @param line
+     * @param nodes
+     * @return
```
matthieun: Checkstyle should fail for that...

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,879 @@
...
+        {
+            if (oneWay != PbfOneWay.REVERSED)
+            {
+                // Find the first node
+                startingIndex = 0;
+                startNode = nodesToSectionAt.getNode(polyline.first());
+                if (!startNode.isPresent())
+                {
+                    // Weren't able to find the starting node, abort slicing
+                    logger.error(
```
matthieun: How often is that error expected to happen?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LineItem.java
```diff
@@ -89,6 +89,14 @@ public Distance length()
         return this.asPolyLine().overallHeading();
     }
 
+    /**
+     * @return the number of shape-points for this item.
+     */
+    public int size()
```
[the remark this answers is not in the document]
MikeGost: Fair point, let me update. Will not use the word nodes as that's specific to Edges, just say "including start and end points".

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfOneWay.java
```diff
@@ -80,7 +80,7 @@ else if (OneWayTag.isOneWayReversible(taggable))
         }
     }
 
-    private static boolean isNotAccessibleToVehicles(final Taggable taggable)
+    public static boolean isNotAccessibleToVehicles(final Taggable taggable)
```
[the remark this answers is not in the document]
MikeGost: After the previous change to fix edge definition consistency (see #65), this method became unused. To avoid warnings, I made it public. I still think it serves a purpose for anyone looking for these tag combinations, but it may not be the place to put it. Candidate for removal.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
MikeGost: Good point!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good point!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Okay!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Heh, I'll fix this :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good point, I'll add this!!

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/store/PbfOneWay.java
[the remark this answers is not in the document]
matthieun: I would just remove it :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
```diff
@@ -0,0 +1,94 @@
...
+    public Optional<TemporaryNode> getNode(final Location location)
+    {
+        for (final TemporaryNode node : this.nodes.keySet())
+        {
+            if (node.getLocation().equals(location))
+            {
+                return Optional.of(node);
+            }
+        }
+
```
snb3300: Nit: Extra Line

src/main/java/org/openstreetmap/atlas/utilities/collections/StreamIterable.java
```diff
@@ -30,6 +30,30 @@ protected StreamIterable(final Iterable<T> source)
...
+    /**
+     * Test whether all elements from iterable matches given predicate or not
+     *
+     * @param predicate
+     *            Predicate to test
+     * @return True when given predicate is true for all entities in iterable, else false
+     */
+    public boolean allMatch(final Predicate<T> predicate)
+    {
+        return StreamSupport.stream(this.source.spliterator(), false).allMatch(predicate);
```
snb3300: Any reason not to use parallel stream here?
```
return StreamSupport.stream(this.source.spliterator(), true).allMatch(predicate)
```

src/main/java/org/openstreetmap/atlas/utilities/collections/StreamIterable.java
```diff
@@ -30,6 +30,30 @@ protected StreamIterable(final Iterable<T> source)
...
+    /**
+     * Test whether any of the elements from iterable matches given predicate
+     *
+     * @param predicate
+     *            Predicate to test
+     * @return True when given predicate is true for any one entity in iterable, else false
+     */
+    public boolean anyMatch(final Predicate<T> predicate)
+    {
+        return StreamSupport.stream(this.source.spliterator(), false).anyMatch(predicate);
```
snb3300: Any reason not to use parallel stream here?
```
return StreamSupport.stream(this.source.spliterator(), true).allMatch(predicate)
```

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -204,6 +205,27 @@ public boolean addAll(final Collection<? extends Location> collection)
...
     }
 
+    /**
+     * Append the given {@link PolyLine} to this one, if possible.
+     *
+     * @param other
+     *            The {@link PolyLine} to append
+     * @return the new, combined {@link PolyLine}
+     */
+    public PolyLine append(final PolyLine other)
```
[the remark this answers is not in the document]
MikeGost: Good question. I thought it'd be easier to determine whether you can append/prepend right away and fail rather than having the client check if an append/prepend was done. Also, was trying to follow the pattern across this class (see get at index or offsetFromStart methods). Open to changing this if anyone feels strongly.

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -826,6 +848,27 @@ public boolean overlapsShapeOf(final PolyLine other)
...
     }
 
+    /**
+     * Prepends the given {@link PolyLine} to this one, if possible.
+     *
+     * @param other
+     *            The {@link PolyLine} to prepend
+     * @return the new, combined {@link PolyLine}
+     */
+    public PolyLine prepend(final PolyLine other)
```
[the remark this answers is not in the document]
MikeGost: Answered above

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -300,6 +296,120 @@ private void populateAtlasMetadata()
...
+        final Set<Long> duplicatePointsToKeep = new HashSet<>();
+
+        for (final Point point : atlas.points())
+        {
+            final Set<Long> duplicatePoints = Iterables.stream(atlas.pointsAt(point.getLocation()))
+                    .map(Point::getIdentifier).collectToSet();
+            if (!duplicatePoints.isEmpty() && duplicatePoints.size() > 1)
+            {
+                // Sort the points
+                final Set<Long> sortedDuplicates = Iterables.asSortedSet(duplicatePoints);
```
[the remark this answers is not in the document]
MikeGost: I haven't seen more than 2-3 OSM Nodes stacked on top of each other. It could be possible if they were created programmatically and uploaded to OSM, but then I think we have bigger problems.

src/main/java/org/openstreetmap/atlas/utilities/collections/StreamIterable.java
MikeGost: We need to add parallel method to Iterables' stream, and then we can add parallel method here as well. Some things the client may not want or can have parallelized - so this would give us more flexibility to do that. Defaulting to non-parallel until that functionality is there.

src/main/java/org/openstreetmap/atlas/utilities/collections/StreamIterable.java
MikeGost: See above

MikeGost:
@matthieun - I have one more change I'd like to make and then merge away. Also, @AlexHsieh22 was going to review as well.

matthieun:
Sg, will wait for @AlexHsieh22's review!

MikeGost:
@matthieun - ready to merge after weekend's changes. Can follow up any changes in the next PR.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
```diff
@@ -0,0 +1,93 @@
...
+     *
+     * @param location
+     *            The target {@link Location}
+     * @return an {@link Optional} containing the {@link TemporaryNode}
+     */
+    public Optional<TemporaryNode> getNode(final Location location)
+    {
+        for (final TemporaryNode node : this.nodes.keySet())
+        {
+            if (node.getLocation().equals(location))
```
AlexHsieh22: do we know how many nodes are typically expected in a changeset?  Would it more performant to track this as <Location,Int> instead and use atlas.itemsAt() calls to get the node?

AlexHsieh22: if we convert this to use streams, would parallelism help here?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionChangeSet.java
```diff
@@ -0,0 +1,113 @@
...
+
+/**
+ * This class tracks all way-section updates that result in changes to raw atlas entities. The
+ * possible changes are:
+ * <ul>
+ * <li>Raw atlas point becoming a {@link Node}.
+ * <li>Raw atlas point staying a {@link Point} (vs. becoming a simple shape point).
+ * <li>Raw atlas line becoming an {@link Edge}.
+ * <li>Raw atlas line becoming an {@link Area}.
+ * <li>{@link Relation} member updates. If a raw atlas line becomes a series of {@link Edge}s, we
```
AlexHsieh22: great comment!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+                {
+                    // Handle isolated rings by creating a node at the start and the middle
+                    // shape point, effectively splitting it in half.
+                    addPointToNodeList(polyLine.first(), nodesForEdge);
+                    addPointToNodeList(polyLine.get(polyLine.size() / 2), nodesForEdge);
+                }
+                else if (!line.isClosed())
+                {
+                    // For non-closed loops, the first and last point of the edge are always nodes.
+                    // We don't care about first/last points for rings because these can be
```
AlexHsieh22: stupid question: what's a non-closed loop?  It looks like this is guaranteeing the first and last node of a line are always added as nodes?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+     * use later for splitting {@link Line}s into {@link Edge}s.
+     *
+     * @param changeSet
+     *            The {@link WaySectionChangeSet} to track any updates
+     */
+    private void identifyEdgesNodesAndAreasFromLines(final WaySectionChangeSet changeSet)
+    {
+        this.rawAtlas.lines().forEach(line ->
+        {
+            if (isAtlasEdge(line))
```
AlexHsieh22: stupid question: are there any atlas lines that might be both an edge and an area? It looks like we always prioritize edge in that case.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+                }
+
+                // Record the edge and all its nodes
+                changeSet.createEdgeToNodeMapping(line.getIdentifier(), nodesForEdge);
+            }
+            else if (isAtlasArea(line))
+            {
+                changeSet.recordArea(line);
+            }
+            else if (isAtlasLine(line))
```
AlexHsieh22: is there a large performance hit for keeping this code around?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+                    // start node occurrence, since the end node will either be used as a future
+                    // start node or be the end of the way, in which case we don't care.
+                    final int startOccurrence = nodesToSectionAt.getOccurrence(startNode.get()) - 1;
+                    nodesToSectionAt.incrementOccurrence(startNode.get());
+                    final int endOccurrence = nodesToSectionAt.getOccurrence(endNode.get()) - 1;
+
+                    // Build the underlying polyline and reverse it, if necessary
+                    final PolyLine rawPolyline = polyline.between(polyline.get(startIndex),
+                            startOccurrence, polyline.get(index), endOccurrence);
+                    final PolyLine edgePolyline = isReversed ? rawPolyline.reversed() : rawPolyline;
```
AlexHsieh22: in the case of a reversed feature, would the resulting identifiers be reversed - and is it expected?  (poorly phrased so an example below).  NOTE: I maybe have confused myself on this, so let me know if I made any sense here.

Let's say line A->B->C is a reversed line to be sectioned at every point.
1. rawPolyline 1 = A->B, and it will be given identifier 1 and the geometry will be B->A
2. rawPolyline 2 = B->C, and it will be given identifier 2 and the geometry will be C->B.

the resulting edge will have section 001 being B->A and then 002 being C->B.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+
+                    // Increment starting pointers
+                    startIndex = index;
+                    startNode = endNode;
+                }
+            }
+        }
+        catch (final Exception e)
+        {
+            logger.error("Failed to way-section line {}", line.getIdentifier(), e);
```
AlexHsieh22: This could return a half populated map of newEdgesForLine.  Is it better to return nothing in this case? Same could happen to the ring case below.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -0,0 +1,906 @@
...
+
+        // Determines if we need to reverse the polyline and if a reverse edge is needed
+        final PbfOneWay oneWay = PbfOneWay.forTag(line);
+        final boolean hasReverseEdge = oneWay == PbfOneWay.NO;
+        final boolean isReversed = oneWay == PbfOneWay.REVERSED;
+
+        // Keep track of the starting index, start/end nodes and relevant flags
+        int startIndex = 0;
+        PolyLine polyLineUpToFirstNode = null;
+        boolean firstLocationNotNode = true;
```
AlexHsieh22: suggest a rename to something like isFirstNode?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryEdge.java
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    public PolyLine getPolyLine()
+    {
+        return this.polyLine;
+    }
+
+    public long getReversedIdentifier()
+    {
+        return -1 * Math.abs(getIdentifier());
```
AlexHsieh22: If the identifier passed in is negative to start with, we could end up always returning the negative identifier.  Should this be return -1 * getIdentifier()?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryNode.java
```diff
@@ -0,0 +1,62 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Node;
+
+/**
+ * The {@link TemporaryNode} object, keeps track of the bare minimum information required to create
+ * an Atlas {@link Node}. It is meant to be as light-weight as possible, keeping track of only the
+ * identifier and location.
+ *
+ * @author mgostintsev
+ */
+public class TemporaryNode implements Serializable
```
AlexHsieh22: Is there a reason why this needs to be Serializable and the TemporaryEdge not?

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -1074,6 +1117,27 @@ public String toWkt()
...
+    /**
+     * Truncates this {@link PolyLine} at the given start and end index
+     *
+     * @param startIndex
+     *            The index before which to truncate from the start
+     * @param endIndex
+     *            The index after which to truncate from the end
+     * @return all the locations in this {@link PolyLine} after truncation.
+     */
+    public Iterable<Location> truncate(final int startIndex, final int endIndex)
```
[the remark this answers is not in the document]
MikeGost: I didn't like the naming of "indexFromEnd" here, but the description of each parameter made sense, so I used that verbatim. Can add a comment here, but not sure how helpful that would be, given that the descriptions already match up.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
```diff
@@ -0,0 +1,93 @@
...
+
+    /**
+     * Adds the given {@link TemporaryNode} to the store
+     *
+     * @param node
+     *            The {@link TemporaryNode} to add
+     */
+    public void addNode(final TemporaryNode node)
+    {
+        this.nodes.put(node, 1);
```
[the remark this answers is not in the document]
MikeGost: At this point, we just care about keeping track of the node's existence, the occurrences don't get decremented, only incremented when used as a start node, so no check is required here.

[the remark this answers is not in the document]
MikeGost: Will add a comment to clarify

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
MikeGost: 1. This code is for Nodes per Line-To-Become-Edge, so number of nodes will be significantly less than if looking at an entire changeset. Number of nodes is [1-999]. A closed-loop and a way with 1000 intersections. Technically an OSM Way can have 2000 shape-points, but since we're sectioning, we're limited to 999 for the identifiers to work with. Most ways don't have nearly this much.
2. Can you explain more on tracking as <Location, Int>? That's essentially what a TemporaryNode is - it has the node identifier and location (bare minimum to identify a specific node). The goal is to query by Location. I thought about using the atlas.pointsAt(Location) similar to addPointToNodeList call - https://github.com/osmlab/atlas/blob/18929714472a6c35fea9e046254ca810cb0415f1/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java#L137. This would work, even though this returns all Points at that location, since we built in the restriction that a raw Atlas will have a single point at each location. However, given the small number of Nodes per Edge and to avoid having logic to handle the case of "what if we have two nodes that are very similar or happen to be on top of each other", I thought it was better to have an explicit equality check for each key.
3. I'll try parallelism here, thanks!

MikeGost: Did some testing with parallel stream, it actually took longer than the code above. Had more instances of greater than 1ms calls to the method than prior version. Here's the code I used for testing.

return this.getNodes().parallelStream()
                .filter(node -> node.getLocation().equals(location)).findFirst();

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: This should read non-closed lines, good catch!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good question. I kept parity with the old slicing logic to have a 1:1 relationship between OSM Way to Atlas Edge/Line/Area. If this needs to be reconsidered, we can create an issue and it would be a pretty straightforward change.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good catch! I originally intended to call out lines that didn't qualify as an Edge, Line or Area and handle them in the else block - log and remove from Atlas, but ultimately made the call to have everything be in the Atlas as a Line if it's not an Area or Edge. Looks like I got a little too cute here and making an unnecessary call to check for Line properties.

MikeGost: Reconsidering the decision to keep everything as a Line, there are a few cases that make no sense. Examples:
1. Line folding on to itself: https://www.openstreetmap.org/way/320353475#map=19/7.26214/-8.14084
2. JTS country-slicing elements which are points.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/NodeOccurrenceCounter.java
AlexHsieh22: hmm good point - I was trying to see if iterating through all the nodes was time consuming enough that we might want to save the location -> atlas identifier map to quickly get the node from a given location.

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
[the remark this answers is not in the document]
AlexHsieh22: ah, with endIndex, I was thinking of String's substring and List's split api that typically use the startIndex (index from start) and endIndex (also an index from start).  Whereas in this case, the startIndex is indexFromStart and endIndex is index from end.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: The identifiers would not be reversed, they would occur in the order of the shape point occurrence. Only the geometry is affected during the reverse. To summarize, your example above is exactly what would happen :) 

This is expected. I compared it with today's code and atlas files for this way (https://www.openstreetmap.org/way/333112568#map=19/5.31774/-4.01493) and there is exact parity.

Before:
screen shot 2018-02-21 at 12 44 38 pm

After: 
screen shot 2018-02-21 at 12 44 48 pm

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good point!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
AlexHsieh22: nice! Thanks for the detailed pictures as well!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryEdge.java
MikeGost: We will never start off with a negative identifier, will add a comment to clarify.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryNode.java
MikeGost: TemporaryEdge should be as well.

MikeGost: It is - TemporaryEdge extends TemporaryEntity, which is Serializable.

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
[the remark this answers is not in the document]
MikeGost: Let me rename it!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Updating the area filter to bring in railway and highway platforms and filtering out anything that doesn't fit this definition!

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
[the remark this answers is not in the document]
AlexHsieh22: thanks!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
AlexHsieh22: awesome! sounds good.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryEdge.java
AlexHsieh22: Ah, then maybe we don't need to wrap it in the Math.abs() as well, that was what made me think of the negative number case :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryNode.java
AlexHsieh22: doh! good point!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/temporary/TemporaryEdge.java
MikeGost: Yep! Ended up removing that call.

PR #80, opened 2018-02-15
Allow updating AtlasMetaData tags during PackedAtlasCloning

AlexHsieh22:
- Adding functionality to update/add extra metadata tags during PackedAtlasCloning.
- Add tests to verify the new functionality.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasCloner.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasCloner.java
@@ -1,5 +1,9 @@
 package org.openstreetmap.atlas.geography.atlas.packed;
 
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
 import org.openstreetmap.atlas.geography.atlas.Atlas;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasCloner.java
```diff
@@ -36,11 +41,24 @@ public PackedAtlas cloneFrom(final Atlas atlas)
...
-        final AtlasMetaData metaData = atlas.metaData();
+        AtlasMetaData metaData = atlas.metaData();
         if (this.shardName != null)
         {
             metaData.copyWithNewShardName(this.shardName);
         }
+
+        if (this.additionalMetaDataTags.isPresent())
+        {
+            // Create a new map to avoid modifying the old atlas tags
```
MikeGost: Nit: this comment no longer needed, since the new map is implicit in the getTags() call with the latest changes.

AlexHsieh22: great catch!

PR #83, opened 2018-02-21
OsmPbfReader to close the pbf InputStreams once Atlas is produced

matthieun:
The osmosis reader does not close the included input stream once the passes are done, which causes resource leaks.

This updates the `OsmPbfLoader` to keep track of those input streams and close them when the Atlas has been produced.

I am not sure how to unit-test this change directly, so suggestions are welcome! On the other side, it is already covered by all test classes using the `OsmosisReaderMock`.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
@@ -42,7 +42,6 @@ public class OsmPbfLoader
         {
             super(input);
             this.input = input;
-
         }
 
         @Override
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
@@ -1,5 +1,8 @@
 package org.openstreetmap.atlas.geography.atlas.pbf;
 
+import java.io.Closeable;
+import java.io.IOException;
+import java.io.InputStream;
 import java.util.function.Supplier;
 
 import org.openstreetmap.atlas.geography.MultiPolygon;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
```diff
@@ -24,11 +28,35 @@
...
+     */
+    public static class CloseableOsmosisReader extends OsmosisReader implements Closeable
+    {
+        private final InputStream input;
+
+        public CloseableOsmosisReader(final InputStream input)
+        {
+            super(input);
+            this.input = input;
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoader.java
```diff
@@ -24,11 +28,34 @@
...
+        }
+    }
+
     private static final Logger logger = LoggerFactory.getLogger(OsmPbfLoader.class);
     private final PackedAtlasBuilder builder = new PackedAtlasBuilder();
     private final OsmPbfProcessor processor;
-    private final Supplier<OsmosisReader> osmosisReaderSupplier;
-    private OsmosisReader reader;
+    private final Supplier<CloseableOsmosisReader> osmosisReaderSupplier;
+    private CloseableOsmosisReader reader;
```
AlexHsieh22: is there a reason why reader is a class level variable?  If ```makeNewReader``` returns a ```CloseableOsmosisReader```, then we can use try-with-resources in the ```read()``` method and leverage the ```Closeable``` interface! I believe that will also remove the need of the ```closeReader()``` method.

matthieun: I agree! This would be nice. I just wanted to do minimal changes to fix the bug, since #17 should replace this code all together at some point.

PR #85, opened 2018-02-22
Configurable Atlas entity filter based on include or exclude polygons

jklamer:
This PR contains a configurable AtlasEntity filter based on either include or exclude polygons. If include polygons are present then AtlasEntity::intersects must return true for one or more include polygons. If no include polygons are present, AtlasEntity::intersects must return false for all exclude polygons. If there are not filter polygons in the configuration then every entity is passed through. Polygons are not allowed to overlap and they are dropped with a warning as they are read in if they overlap with any previously read polygon. 
The supported polygon string formats are
<li>Atlas's native format</li>
<li>wkt</li>
<li>wkb</li>
<li>geojson*</li>

*This is the only format where more that one polygon per string is permitted

Example Configurations:
```
{
    "filter.polygons":{
        "include.wkt":["POLYGON ((-81.1523656 22.5165725, -81.1140974 22.5202238, -81.0787037 22.5072781, -81.0769071 22.470924, -81.1119414 22.4493395, -81.1732065 22.4481772, -81.1971017 22.4740784, -81.1924305 22.5091038, -81.1577555 22.5288536, -81.1523656 22.5165725))"]
    }
}

```

```
{
    "filter.polygons":{
        "exclude.geojson":["{\"type\":\"FeatureCollection\",\"features\":[{\"type\":\"Feature\",\"geometry\":{\"type\":\"Polygon\",\"coordinates\":[[[-81.1523656,22.5165725],[-81.1140974,22.5202238],[-81.0787037,22.5072781],[-81.0769071,22.470924],[-81.1119414,22.4493395],[-81.1732065,22.4481772],[-81.1971017,22.4740784],[-81.1924305,22.5091038],[-81.1577555,22.5288536]]]},\"properties\":{}}]}"]
    }
}
```

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
@@ -1,0 +1,168 @@
+package org.openstreetmap.atlas.utilities.filters;
+
+import java.io.Serializable;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.openstreetmap.atlas.geography.Polygon;
+import org.openstreetmap.atlas.geography.atlas.items.AtlasEntity;
+import org.openstreetmap.atlas.geography.converters.PolygonStringConverter;
+import org.openstreetmap.atlas.geography.converters.WkbPolygonConverter;
+import org.openstreetmap.atlas.geography.converters.WktPolygonConverter;
+import org.openstreetmap.atlas.geography.index.QuadTree;
+import org.openstreetmap.atlas.streaming.readers.GeoJsonReader;
+import org.openstreetmap.atlas.streaming.readers.json.serializers.PropertiesLocated;
+import org.openstreetmap.atlas.streaming.resource.StringResource;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.conversion.StringConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.vividsolutions.jts.io.WKBReader;
+
+/**
+ * Configurable {@link AtlasEntity} filter that passes through anything that intersects with include
+ * polygons or anything that doesn't intersect with exclude polygons. Exclude polygons are looked at
+ * only if no include polygons exist. Include takes precedence and polygons intersecting with
+ * previously read polygons are dropped with a warning.
+ * 
+ * @author jklamer
+ */
+public class AtlasEntityPolygonsFilter implements Predicate<AtlasEntity>, Serializable
+{
+
+    public static final String EXCLUDED_POLYGONS_KEY = "filter.polygons.exclude";
+    public static final String INCLUDED_POLYGONS_KEY = "filter.polygons.include";
+    private static final PolygonStringConverter POLYGON_STRING_CONVERTER = new PolygonStringConverter();
+    private static final List<String> POLYGON_STRING_FORMATS = Arrays.asList("atlas", "geojson",
+            "wkt", "wkb");
+    private static final WkbPolygonConverter WKB_POLYGON_CONVERTER = new WkbPolygonConverter();
+    private static final WktPolygonConverter WKT_POLYGON_CONVERTER = new WktPolygonConverter();
+    private static final Logger logger = LoggerFactory.getLogger(AtlasEntityPolygonsFilter.class);
+    private static final long serialVersionUID = -3474748398986569205L;
+    private List<Polygon> exclude = new ArrayList<>();
+    private List<Polygon> include = new ArrayList<>();
+
+    private static StringConverter<Optional<List<Polygon>>> getConverterForFormat(
+            final String format)
+    {
+        switch (format)
+        {
+            case "atlas":
+                return string -> Optional
+                        .of(Collections.singletonList(POLYGON_STRING_CONVERTER.convert(string)));
+            case "geojson":
+                return string ->
+                {
+                    final List<Polygon> polygons = new ArrayList<>();
+                    final GeoJsonReader reader = new GeoJsonReader(new StringResource(string));
+                    while (reader.hasNext())
+                    {
+                        final PropertiesLocated propertiesLocated = reader.next();
+                        if (propertiesLocated.getItem() instanceof Polygon)
+                        {
+                            polygons.add((Polygon) propertiesLocated.getItem());
+                        }
+                        else
+                        {
+                            logger.warn("Polygon Filter does not support item {}",
+                                    propertiesLocated.toString());
+                        }
+                    }
+                    return Optional.of(polygons).filter(list -> !list.isEmpty());
+                };
+            case "wkt":
+                return string -> Optional.of(
+                        Collections.singletonList(WKT_POLYGON_CONVERTER.backwardConvert(string)));
+            case "wkb":
+                return string -> Optional.of(Collections.singletonList(
+                        WKB_POLYGON_CONVERTER.backwardConvert(WKBReader.hexToBytes(string))));
+            default:
+                logger.warn("No converter set up for {} format. Supported formats are {}", format,
+                        POLYGON_STRING_FORMATS);
+                return string -> Optional.empty();
+        }
+    }
+
+    private static Predicate<Polygon> notOverlapping(final QuadTree<Polygon> vettedPolygons,
+            final String polygonType)
+    {
+        return polygon ->
+        {
+            if (vettedPolygons.get(polygon.bounds()).stream().noneMatch(polygon::overlaps))
+            {
+                vettedPolygons.add(polygon.bounds(), polygon);
+                return true;
+            }
+            else
+            {
+                logger.warn("Dropping {} polygon {}", polygonType, polygon);
+                return false;
+            }
+        };
+    }
+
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final WktPolygonConverter WKT_POLYGON_CONVERTER = new WktPolygonConverter();
+    private static final Logger logger = LoggerFactory.getLogger(AtlasEntityPolygonsFilter.class);
+    private static final long serialVersionUID = -3474748398986569205L;
+    private List<Polygon> exclude = new ArrayList<>();
+    private List<Polygon> include = new ArrayList<>();
+
+    private static StringConverter<Optional<List<Polygon>>> getConverterForFormat(
+            final String format)
+    {
+        switch (format)
```
MikeGost: How about an Enum here to call out all supported types?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+ * Configurable {@link AtlasEntity} filter that passes through anything that intersects with include
+ * polygons or anything that doesn't intersect with exclude polygons. Exclude polygons are looked at
+ * only if no include polygons exist. Include takes precedence and polygons intersecting with
+ * previously read polygons are dropped with a warning.
+ * 
+ * @author jklamer
+ */
+public class AtlasEntityPolygonsFilter implements Predicate<AtlasEntity>, Serializable
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+                return string -> Optional.of(Collections.singletonList(
+                        WKB_POLYGON_CONVERTER.backwardConvert(WKBReader.hexToBytes(string))));
+            default:
+                logger.warn("No converter set up for {} format. Supported formats are {}", format,
+                        POLYGON_STRING_FORMATS);
+                return string -> Optional.empty();
+        }
+    }
+
+    private static Predicate<Polygon> notOverlapping(final QuadTree<Polygon> vettedPolygons,
```
MikeGost: Maybe some javadoc to explain what's happening here?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+import org.openstreetmap.atlas.streaming.resource.StringResource;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.conversion.StringConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.vividsolutions.jts.io.WKBReader;
+
+/**
+ * Configurable {@link AtlasEntity} filter that passes through anything that intersects with include
```
MikeGost: If I wanted to extend this functionality to be slightly different - say instead of intersecting given polygons, want fully enclosed within given polygons - this means I'd have to change this core class quite a bit. What are your thoughts on making this more generic - and allowing you to plug any kind of filters/predicates in given some set of polygons? I think we talked offline about TaggableFilter being generic enough to handle any tag combination. In my mind, this would be a lot more powerful if it could handle any kind of entity inclusion/exclusion rules.

src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
```diff
@@ -0,0 +1,193 @@
...
+        this.assertCounts(testCountsAtlas, excludePolygon2, totalPointCount, totalLineCount - 2L,
+                totalAreaCount - 1L, totalRelationCount - 1L);
+        this.assertCounts(testCountsAtlas, excludePolygon2.negate(), 0L, 2L, 1L, 1L);
+
+        // testing tandem filter
+        this.assertCounts(testCountsAtlas, includePolygon1And2, totalPointCount - 1L,
+                totalLineCount - 2L, totalAreaCount, totalRelationCount);
+        this.assertCounts(testCountsAtlas, excludePolygon1And2, 1L, 2L, 0L, 0L);
+        this.assertCounts(testCountsAtlas, includePolygon1And2.negate(), 1L, 2L, 0L, 0L);
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+    private static final List<String> POLYGON_STRING_FORMATS = Arrays.asList("atlas", "geojson",
+            "wkt", "wkb");
+    private static final WkbPolygonConverter WKB_POLYGON_CONVERTER = new WkbPolygonConverter();
+    private static final WktPolygonConverter WKT_POLYGON_CONVERTER = new WktPolygonConverter();
+    private static final Logger logger = LoggerFactory.getLogger(AtlasEntityPolygonsFilter.class);
+    private static final long serialVersionUID = -3474748398986569205L;
+    private List<Polygon> exclude = new ArrayList<>();
+    private List<Polygon> include = new ArrayList<>();
+
+    private static StringConverter<Optional<List<Polygon>>> getConverterForFormat(
```
matthieun: Any reason this and the next method are static?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,168 @@
...
+import org.openstreetmap.atlas.geography.index.QuadTree;
+import org.openstreetmap.atlas.streaming.readers.GeoJsonReader;
+import org.openstreetmap.atlas.streaming.readers.json.serializers.PropertiesLocated;
+import org.openstreetmap.atlas.streaming.resource.StringResource;
+import org.openstreetmap.atlas.utilities.configuration.Configuration;
+import org.openstreetmap.atlas.utilities.conversion.StringConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.vividsolutions.jts.io.WKBReader;
```
matthieun: It looks like you are using the WKB converter below... I can't see where this import is used?

src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
```diff
@@ -0,0 +1,193 @@
...
+    @Test
+    public void testIncludeDominant()
+    {
+        final int numberOfCases = 4;
+        final Atlas includeExcludeArrangements = this.setup.getIncludeExcludeArrangements();
+        final String configurationStringFormat = "{\"filter.polygons\":{\"include.wkt\":[\"%s\"],\"exclude.wkt\":[\"%s\"]}}";
+        final Polygon includePolygon = this.getPolygonWithName(includeExcludeArrangements,
+                "include");
+
+        for (int caseNum = 1; caseNum <= numberOfCases; caseNum++)
```
matthieun: `caseNum` -> `caseNumber`

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: Okay, can we go over of the dimensions? So we want to be able to choose which of the following:
<li>Include versus exclude dominant</li>
<li>Intersects versus fully geometrically encloses </li>
<li> Overlapping allowed versus disallowed (And that effect of the dominance)</li>
<li>Settings specific to polygon type or generic</li>

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: They don't rely on any instance variables. Is that a bad design choice?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: WKB string are displayed in Hex format with two bytes per string character. This converts the Hex string to bytes for the converter [here](https://github.com/osmlab/atlas/pull/85/files#diff-327065d3a2dcb65a8d798a639325a5d9R86)

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: Also, should we make these setting file configurable or provided by the user depending on filter use case?

MikeGost: I would lean towards... we want as much provided by the user as possible. They can specify their own filters and sets of polygons. Making it configurable is nice, but they can do that in the client code. This way we don't have to deal with updating various configurations in this class.

jklamer:
Alright, Going to Close this to add
<li>MultiPolygon Support</li>
<li>Entity Polygon Filter customizability</li>

PR #86, opened 2018-02-22
Way Sectioning Cleanup

MikeGost:
Part of #17. A few changes here:

1. Bug Fixes
   a. When we filtered out removed `Point`s from the raw atlas, we didn't check that we had removed 
       them when adding members to the raw relations. Added a check for this when building a 
      `Relation`.
   b. There was a bug in how we were calculating the line count for the final `AtlasSize` estimate - we 
       were relying on the edge count (much larger than actual lines converted to edges), where we 
       should have been looking at the number of lines that were split into edges.
   c. There was a bug when adding `Edge`s to `Relation`s - we were adding all created `Edge`s to all 
       `Relation`s, instead of only the `Edge`s for the original `Line`.
   d. Avoided returning partially sliced lines in case of errors - this is now an erroneous event causing 
       sectioning to stop.
2. Closing the resource leak for raw atlas generation, similar to the issue in #83 - updating the fix for both issues to have a standalone Closeable class.
3. Explicitly handling entities that don't become `Edge`s, `Area`s or `Line`s in the final Atlas. Before everything that wasn't an `Area` or `Edge` became a `Line`. This was an issue as there were some invalid JTS country-slicing artifacts that created single-point lines. Now, we explicitly check for all 3 entities, and if it's not one of those, the feature will not be included in the `Atlas`. These are also excluded from any `Relation`s they were a part of. Some examples of things that were tossed include: https://www.openstreetmap.org/way/320353475 (invalid line overlapping itself)
4. Expanded the `Area` filter to include railway platforms, traverses and stations as well as highway platforms. This brought in valid features such as https://www.openstreetmap.org/way/402531564 and https://www.openstreetmap.org/way/295528557. Note: these are both not in the `Atlas` today, so this will be a minor improvement.
5. Added `traverse` value for the `RailwayTag`
6. Improved the logging - removed some of the chatty logging and tightened down on the log levels. Added logging to track time and start/stop of certain functions.
7. Renamed and added constants from previous PR comments. 
8. Simplified a `PackedAtlas` null check I encountered during debugging


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -166,8 +165,14 @@ public RawAtlasGenerator withMetaData(final AtlasMetaData metaData)
     private Atlas buildRawAtlas()
     {
         final Time parseTime = Time.now();
-        connectOsmPbfToPbfConsumer(this.pbfReader);
-        this.reader.run();
+        try (CloseableOsmosisReader reader = connectOsmPbfToPbfConsumer(this.pbfReader))
```
matthieun: Very good!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -238,33 +246,35 @@ else if (builder.peek().node(memberIdentifier) != null)
...
-                            changeSet.getLineToCreatedEdgesMapping().allValues().forEach(edge ->
-                            {
-                                bean.addItem(edge.getIdentifier(), member.getRole(), ItemType.EDGE);
-                                if (edge.hasReverse())
-                                {
-                                    bean.addItem(edge.getReversedIdentifier(), member.getRole(),
-                                            ItemType.EDGE);
-                                }
-                            });
+                            changeSet.getLineToCreatedEdgesMapping().get(memberIdentifier)
```
matthieun: Hah! That saves some time!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
MikeGost: Kudos to @AlexHsieh22 for pointing this out in #83!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Just a bit, and some memory too :)

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -924,16 +925,7 @@ protected void addRelation(final long relationIdentifier, final long relationOsm
...
-        for (final Long identifier : identifiers)
-        {
-            if (identifier == null)
-            {
-                someNull = true;
-                break;
-            }
-        }
-        if (someNull)
+        if (identifiers.stream().anyMatch(Objects::isNull))
```
AlexHsieh22: nice! so much simpler!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -337,8 +349,15 @@ private Atlas rebuildAtlas(final Atlas atlas, final Set<Long> pointsToRemove,
         atlas.relations().forEach(relation ->
         {
             final RelationBean bean = new RelationBean();
-            relation.members().forEach(member -> bean.addItem(member.getEntity().getIdentifier(),
-                    member.getRole(), member.getEntity().getType()));
+            relation.members().forEach(member ->
+            {
+                final long memberIdentifier = member.getEntity().getIdentifier();
+                if (!pointsToRemove.contains(memberIdentifier))
```
AlexHsieh22: Do we need to make sure that the relation member entity is a point?

MikeGost: Good point, there might be a slim chance that there is a line member with the same identifier.

MikeGost:
Let me look around for any possible unit test improvements to validate the bug fixes!

MikeGost:
@matthieun - most of these were uncovered when testing larger shards, capturing in unit test might be a little difficult and some were more functionality changes. Ready to merge once Travis finishes!

PR #87, opened 2018-02-24
Discard restriction relation with more than 1 via node

snb3300:
According to the Open street map wiki any restriction relation cannot have more than 1 via nodes in it. You can find the specs [here](https://wiki.openstreetmap.org/wiki/Relation:restriction#Members). This PR adds a check in TurnRestriction constructor to discard the relation if it has more than 1 via nodes.

Example:
https://www.openstreetmap.org/relation/4008061

turn_restriction

This one is a bi-directional edge with no_u_turn at each node within a relation. Even though it is a data issue in OSM, atlas should discard such cases.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
@@ -15,6 +15,7 @@ import org.openstreetmap.atlas.geography.Located;
 import org.openstreetmap.atlas.geography.Rectangle;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.geojson.GeoJsonBuilder.LocationIterableProperties;
+import org.openstreetmap.atlas.tags.RelationTypeTag;
 import org.openstreetmap.atlas.tags.TurnRestrictionTag;
 import org.openstreetmap.atlas.utilities.collections.Maps;
 import org.openstreetmap.atlas.utilities.collections.StringList;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
```diff
@@ -239,6 +240,21 @@ private TurnRestriction(final Relation relation)
...
+            // incorrectly modeled.
+            // To bring back the turn restriction OSM data needs to be modeled correctly
+            final long viaNodeCount = relation.members().stream().filter(
+                    member -> "via".equals(member.getRole()) && member.getEntity() instanceof Node)
+                    .count();
+            if (viaNodeCount > 1)
+            {
+                throw new CoreException(
+                        "Restriction relation should not have more than 1 via node. But, {} has {} via nodes",
+                        relation.getOsmIdentifier(), viaNodeCount);
```
AlexHsieh22: Instead of iterating through the relation members twice, maybe we can move this check to after the viaMembers have been populated and iterate through viaMembers instead?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
```diff
@@ -239,6 +240,21 @@ private TurnRestriction(final Relation relation)
...
+            }
+
             // Build the via members
             final Set<AtlasItem> viaMembers = new HashSet<>();
```
MikeGost: Nit: not your changes, one possibility to remove the set declaration and do it inline is:
`final Set<AtlasItem> viaMembers = relation.members().stream().filter(member -> "via".equals(member.getRole()) && (member.getEntity() instanceof Node || member.getEntity() instanceof Edge)).map(RelationMember::getEntity).map(entity -> (AtlasItem) entity).collect(Collectors.toSet());`

src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
MikeGost: Plus one on this and it would simplify the condition to check only for Nodes, since at this point we're only looking at the via roles!

src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/restriction/ComplexTurnRestrictionTest.java
```diff
@@ -194,4 +195,16 @@ public void testTurnRestrictionsFromComplexBigNodes()
         Assert.assertTrue("Verify that this explicit restricted path is returned", restrictedRoutes
                 .stream().anyMatch(route -> route.toString().equals(expectedRestrictedRoute)));
     }
+
+    @Test
+    public void testTurnRestrictionWithTwoViaNodesInRelation()
```
MikeGost: Nice test case!

src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
```diff
@@ -239,6 +240,21 @@ private TurnRestriction(final Relation relation)
             // Try to re-build the route, based on the "from", "via" and "to" members of the
             // relation
 
+            // According to OSM Wiki a restriction relation member can not have more than 1 via node
+            // https://wiki.openstreetmap.org/wiki/Relation:restriction#Members
+            // If the relation has more than 1 via node then discard the restriction as it is
+            // incorrectly modeled.
+            // To bring back the turn restriction OSM data needs to be modeled correctly
+            final long viaNodeCount = relation.members().stream().filter(
+                    member -> "via".equals(member.getRole()) && member.getEntity() instanceof Node)
```
MikeGost: Nit: To avoid directly using strings, we can consider using `RelationTypeTag` constants here and in the rest of the class.

jklamer:
Should this not be an Atlas Check?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/TurnRestriction.java
snb3300: Makes sense 👍

snb3300:
Valid point @jklamer about atlas checks. I am not well versed with atlas-checks repo so feel free to correct me here. After giving it a thought and reading the atlas-checks project description I feel this should be a part of atlas TurnRestriction api. Reason being, the OSM data itself is not adhering to the spec so we should not even include it in atlas. Let me know if you think otherwise.

jklamer:
@snb3300 I agree. Had a brief talk with @matthieun and because this does not actually delete the underlying relation, and just our object built on top of it, the Atlas is still being faithful to its mission of accurate representation. Also, because the relation still exists an Atlas Check to flag this invalid relation would allow these situations to be easily found and fixed.

jklamer:
Turns out [it already exists](https://github.com/jklamer/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/relations/InvalidTurnRestrictionCheck.java) and will now flag this as well because of this update! Well done @snb3300 ! 😄

PR #89, opened 2018-02-26
Atlas Entity Polygon Filter

jklamer:
This PR contains a configurable AtlasEntity filter based on either include or exclude ```Polygon```s or ```MultiPolygon```s. The filter is either an include or exclude filter. If this filter is constructed using a configuration then it will use whatever type ```Polygon```s/```MultiPolygon```s are present. If both include and exclude ```Polygon```s/```MultiPolygon```s are presents then only the includes are considered.  If there are no filter ```Polygon```s in the configuration then every entity is passed through. ```Polygon```s/```MultiPolygon```s are not allowed to overlap and they are dropped with a warning as they are read in if they overlap with any previously read ```Polygon```/```MultiPolygon```. 

The intersection logic can also be configured by implementing ``` IntersectionPolicy``` which allows you to decide what is intersection between ```AtlasEntities```  and ```Polygon```s and ```MultiPolygon```s.
The supported configuration ```Polygon```/```MultiPolygon``` string formats are
<li>Atlas's native format</li>
<li>wkt</li>
<li>wkb</li>
<li>geojson*</li>

*This is the only format where more that one polygon/multipolygon per string is permitted

Example Configurations:
```
{
    "filter.polygons":{
        "include.wkt":["POLYGON ((-81.1523656 22.5165725, -81.1140974 22.5202238, -81.0787037 22.5072781, -81.0769071 22.470924, -81.1119414 22.4493395, -81.1732065 22.4481772, -81.1971017 22.4740784, -81.1924305 22.5091038, -81.1577555 22.5288536, -81.1523656 22.5165725))"]
    }
}

```

```
{
    "filter.polygons":{
        "exclude.geojson":["{\"type\":\"FeatureCollection\",\"features\":[{\"type\":\"Feature\",\"geometry\":{\"type\":\"Polygon\",\"coordinates\":[[[-81.1523656,22.5165725],[-81.1140974,22.5202238],[-81.0787037,22.5072781],[-81.0769071,22.470924],[-81.1119414,22.4493395],[-81.1732065,22.4481772],[-81.1971017,22.4740784],[-81.1924305,22.5091038],[-81.1577555,22.5288536]]]},\"properties\":{}}]}"]
    }
}
```

Example Static constructions:
```
AtlasEntityPolygonsFilter.Type.INCLUDE.multiPolygons(Collections.singleton(multiPolygon1)
AtlasEntityPolygonsFilter.Type.INCLUDE.polygons(Arrays.asList(polygon1, polygon2)
AtlasEntityPolygonsFilter.Type.INCLUDE.polygonsAndMultiPolygons(Collections.singleton(polygon1),Arrays.asList(multiPolygon1, multiPolygon2)
AtlasEntityPolygonsFilter.Type.EXCLUDE.polygons(Collections.singleton(polygon1))
```

***Bonus***
<li>Added WkbMultiPolygonConverter</li>
<li>Added MultiPolygon::overlaps(MultiPolygon) function</li>
<li>Added functions so that MultiPolygons can be output as valid geojson**</li> 
**More to be done here

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
@@ -1,0 +1,42 @@
+package org.openstreetmap.atlas.geography.converters;
+
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.io.ParseException;
+import com.vividsolutions.jts.io.WKBReader;
+import com.vividsolutions.jts.io.WKBWriter;
+import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.geography.MultiPolygon;
+import org.openstreetmap.atlas.geography.converters.jts.JtsMultiPolygonToMultiPolygonConverter;
+import org.openstreetmap.atlas.utilities.conversion.TwoWayConverter;
+
+/**
+ * Converter class for conversion between Wkb byte array and {@link MultiPolygon}
+ *
+ * @author jklamer
+ */
+public class WkbMultiPolygonConverter implements TwoWayConverter<MultiPolygon, byte[]>
+{
+    @Override
+    public MultiPolygon backwardConvert(final byte[] wkb)
+    {
+        com.vividsolutions.jts.geom.MultiPolygon geometry = null;
+        final WKBReader myReader = new WKBReader();
+        try
+        {
+            geometry = (com.vividsolutions.jts.geom.MultiPolygon) myReader.read(wkb);
+        }
+        catch (final ParseException | ClassCastException e)
+        {
+            throw new CoreException("Cannot parse wkb : {}", WKBWriter.toHex(wkb));
+        }
+        return new JtsMultiPolygonToMultiPolygonConverter().convert(geometry);
+    }
+
+    @Override
+    public byte[] convert(final MultiPolygon multiPolygon)
+    {
+        final Geometry geometry = new JtsMultiPolygonToMultiPolygonConverter().backwardConvert(multiPolygon);
+        final byte[] wkb = new WKBWriter().write(geometry);
+        return wkb;
+    }
+}
--- a/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
@@ -1,13 +1,14 @@
 package org.openstreetmap.atlas.geography.converters;
 
-import com.vividsolutions.jts.geom.Geometry;
-import com.vividsolutions.jts.io.ParseException;
-import com.vividsolutions.jts.io.WKBReader;
-import com.vividsolutions.jts.io.WKBWriter;
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.MultiPolygon;
 import org.openstreetmap.atlas.geography.converters.jts.JtsMultiPolygonToMultiPolygonConverter;
 import org.openstreetmap.atlas.utilities.conversion.TwoWayConverter;
+
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.io.ParseException;
+import com.vividsolutions.jts.io.WKBReader;
+import com.vividsolutions.jts.io.WKBWriter;
 
 /**
  * Converter class for conversion between Wkb byte array and {@link MultiPolygon}
@@ -35,7 +36,8 @@ public class WkbMultiPolygonConverter implements TwoWayConverter<MultiPolygon, b
     @Override
     public byte[] convert(final MultiPolygon multiPolygon)
     {
-        final Geometry geometry = new JtsMultiPolygonToMultiPolygonConverter().backwardConvert(multiPolygon);
+        final Geometry geometry = new JtsMultiPolygonToMultiPolygonConverter()
+                .backwardConvert(multiPolygon);
         final byte[] wkb = new WKBWriter().write(geometry);
         return wkb;
     }
--- a/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
@@ -17,14 +17,15 @@ import com.vividsolutions.jts.io.WKBWriter;
  */
 public class WkbMultiPolygonConverter implements TwoWayConverter<MultiPolygon, byte[]>
 {
+    private static final WKBReader WKB_READER = new WKBReader();
+
     @Override
     public MultiPolygon backwardConvert(final byte[] wkb)
     {
         com.vividsolutions.jts.geom.MultiPolygon geometry = null;
-        final WKBReader myReader = new WKBReader();
         try
         {
-            geometry = (com.vividsolutions.jts.geom.MultiPolygon) myReader.read(wkb);
+            geometry = (com.vividsolutions.jts.geom.MultiPolygon) WKB_READER.read(wkb);
         }
         catch (final ParseException | ClassCastException e)
         {
--- a/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
@@ -439,7 +439,8 @@ public class GeoJsonBuilder
     }
 
     /**
-     * Creates multipolygons where first polygon assumed to be outer ring and the rest are inner.
+     * Creates multipolygon from {@link Iterable} of {@link Polygon}s where first polygon is assumed
+     * to be the outer ring and the rest are inner.
      * 
      * @param polygons
      *            an iterable of polygons where the first is assumed to be the outer polygon in a
--- a/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.Location;
@@ -53,14 +54,15 @@ public class GeoJsonBuilder
         }
     }
 
-    public static final String FEATURES = "features";
-    public static final String GEOMETRY = "geometry";
     public static final String COORDINATES = "coordinates";
-    public static final String TYPE = "type";
     public static final String FEATURE = "Feature";
+    public static final String FEATURES = "features";
     public static final String FEATURE_COLLECTION = "FeatureCollection";
-    public static final String GEOMETRY_COLLECTION = "GeometryCollection";
     public static final String GEOMETRIES = "geometries";
+    public static final String GEOMETRY = "geometry";
+    public static final String GEOMETRY_COLLECTION = "GeometryCollection";
+    public static final String PROPERTIES = "properties";
+    public static final String TYPE = "type";
     private static final Logger logger = LoggerFactory.getLogger(GeoJsonBuilder.class);
     private final int logFrequency;
 
@@ -229,6 +231,36 @@ public class GeoJsonBuilder
     public GeoJsonObject create(final PolyLine polyLine)
     {
         return this.create(polyLine, GeoJsonType.LINESTRING);
+    }
+
+    /**
+     * Creates a GeoJson FeatureCollection containing a list of GeoJsonObject Features
+     *
+     * @param objects
+     *            the features
+     * @return a GeoJson FeatureCollection
+     */
+    public GeoJsonObject createFeatureCollection(final Iterable<GeoJsonObject> objects)
+    {
+        final JsonObject result = new JsonObject();
+        result.addProperty(TYPE, FEATURE_COLLECTION);
+        final JsonArray features = new JsonArray();
+        int counter = 0;
+        for (final GeoJsonObject object : objects)
+        {
+            if (this.logFrequency > 0 && ++counter % this.logFrequency == 0)
+            {
+                logger.info("Processed {} features.", counter);
+            }
+            if (!Optional.ofNullable(object.jsonObject().get(TYPE))
+                    .filter(jsonObject -> jsonObject.getAsString().equals(FEATURE)).isPresent())
+            {
+                throw new CoreException("Illegal GeoJson Type for Feature collection");
+            }
+            features.add(object.jsonObject());
+        }
+        result.add(FEATURES, features);
+        return new GeoJsonObject(result);
     }
 
     /**
@@ -403,6 +435,45 @@ public class GeoJsonBuilder
         geometry.addProperty(TYPE, GeoJsonType.MULTI_POLYGON.getType());
         geometry.add(COORDINATES, coordinates);
         result.add(GEOMETRY, geometry);
+        return new GeoJsonObject(result);
+    }
+
+    /**
+     * Creates multipolygons where first polygon assumed to be outer ring and the rest are inner.
+     * 
+     * @param polygons
+     *            an iterable of polygons where the first is assumed to be the outer polygon in a
+     *            multipolygon
+     * @return a MultiPolygon geojson feature with one polygon that geometrically represents a
+     *         single outer Atlas Multipolygon
+     */
+    public GeoJsonObject createOneOuterMultiPolygon(final Iterable<Polygon> polygons)
+    {
+        // Create the coordinates for each polygon
+        final List<GeoJsonObject> objects = new ArrayList<>();
+        for (final Polygon polygon : polygons)
+        {
+            objects.add(this.create(polygon, GeoJsonType.MULTI_POLYGON));
+        }
+
+        final JsonObject result = new JsonObject();
+        result.addProperty(TYPE, FEATURE);
+        final JsonArray coordinates = new JsonArray();
+
+        // Add the coordinates back for the entire object
+        for (final GeoJsonObject object : objects)
+        {
+            coordinates
+                    .add(object.jsonObject().getAsJsonObject(GEOMETRY).getAsJsonArray(COORDINATES));
+        }
+
+        final JsonArray newCoordinates = new JsonArray();
+        newCoordinates.add(coordinates);
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty(TYPE, GeoJsonType.MULTI_POLYGON.getType());
+        geometry.add(COORDINATES, newCoordinates);
+        result.add(GEOMETRY, geometry);
+        result.add(PROPERTIES, new JsonObject());
         return new GeoJsonObject(result);
     }
 
--- a/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
@@ -3,18 +3,29 @@ package org.openstreetmap.atlas.utilities.filters;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.Collection;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
+import org.openstreetmap.atlas.geography.Located;
+import org.openstreetmap.atlas.geography.MultiPolygon;
 import org.openstreetmap.atlas.geography.Polygon;
+import org.openstreetmap.atlas.geography.atlas.items.Area;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasEntity;
+import org.openstreetmap.atlas.geography.atlas.items.LineItem;
+import org.openstreetmap.atlas.geography.atlas.items.LocationItem;
+import org.openstreetmap.atlas.geography.atlas.items.Relation;
+import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
+import org.openstreetmap.atlas.geography.converters.MultiPolygonStringConverter;
 import org.openstreetmap.atlas.geography.converters.PolygonStringConverter;
+import org.openstreetmap.atlas.geography.converters.WkbMultiPolygonConverter;
 import org.openstreetmap.atlas.geography.converters.WkbPolygonConverter;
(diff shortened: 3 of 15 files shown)


src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
```diff
@@ -406,6 +438,45 @@ public GeoJsonObject createMultiPolygons(final Iterable<Polygon> polygons)
...
+    /**
+     * Creates multipolygons where first polygon assumed to be outer ring and the rest are inner.
+     * 
+     * @param polygons
+     *            an iterable of polygons where the first is assumed to be the outer polygon in a
+     *            multipolygon
+     * @return a MultiPolygon geojson feature with one polygon that geometrically represents a
+     *         single outer Atlas Multipolygon
+     */
+    public GeoJsonObject createOneOuterMultiPolygon(final Iterable<Polygon> polygons)
```
MikeGost: The method name doesn't match up with the description.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+        {
+            return entity.intersects(polygon);
+        }
+    };
+    private static final long serialVersionUID = -3474748398986569205L;
+    private List<MultiPolygon> excludeMultiPolygons = new ArrayList<>();
+    private List<Polygon> excludePolygons = new ArrayList<>();
+    private Type filterType;
+    private List<MultiPolygon> includeMultiPolygons = new ArrayList<>();
+    private List<Polygon> includePolygons = new ArrayList<>();
```
MikeGost: Why do we separate the polygons and multipolygons into separate collections? What if use `Iterable<Polygon>` to combine both into one.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+            final QuadTree<Located> vettedPolygons, final Type polygonType)
+    {
+        return multiPolygon ->
+        {
+            if (vettedPolygons.get(multiPolygon.bounds()).stream()
+                    .filter(located -> located instanceof Polygon).map(located -> (Polygon) located)
+                    .noneMatch(multiPolygon::overlaps)
+                    && vettedPolygons.get(multiPolygon.bounds()).stream()
+                            .filter(located -> located instanceof MultiPolygon)
+                            .map(located -> (MultiPolygon) located).map(MultiPolygon::outers)
```
MikeGost: Do we care about the inners here?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+    private static Predicate<Polygon> notOverlappingPolygon(final QuadTree<Located> vettedPolygons,
+            final Type polygonType)
+    {
+        return polygon ->
+        {
+            if (vettedPolygons.get(polygon.bounds()).stream()
+                    .filter(located -> located instanceof Polygon).map(located -> (Polygon) located)
+                    .noneMatch(polygon::overlaps)
+                    && vettedPolygons.get(polygon.bounds()).stream()
+                            .filter(located -> located instanceof MultiPolygon)
```
MikeGost: Same question about inners here.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+    {
+        return entity -> this.includePolygonsEmpty() && this.excludePolygonsEmpty();
+    }
+
+    /**
+     * Interface to be implemented if custom intersection logic is required. The
+     * {@link AtlasEntityPolygonsFilter#DEFAULT_INTERSECTION_DECIDING} is that an entity is said to
+     * intersect when any part of the entity overlaps with any area of the Polygon or MultiPolygon.
+     */
+    public interface IntersectionDeciding extends Serializable
```
MikeGost: How about `IntersectionPolicy`?

src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
```diff
@@ -0,0 +1,443 @@
...
+import com.vividsolutions.jts.io.WKBWriter;
+
+/**
+ * Tests for {@link AtlasEntityPolygonsFilter}
+ *
+ * @author jklamer
+ */
+public class AtlasEntityPolygonsFilterTest
+{
+
```
MikeGost: Nit: Extra line

src/main/java/org/openstreetmap/atlas/geography/converters/WkbMultiPolygonConverter.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author jklamer
+ */
+public class WkbMultiPolygonConverter implements TwoWayConverter<MultiPolygon, byte[]>
+{
+    @Override
+    public MultiPolygon backwardConvert(final byte[] wkb)
+    {
+        com.vividsolutions.jts.geom.MultiPolygon geometry = null;
+        final WKBReader myReader = new WKBReader();
```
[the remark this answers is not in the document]
jklamer: That would make the class not serializable because the Converter Classes extend from Function which is not serializable.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: The ```Iterable<Polygon>``` of polygons returned by a ```MultiPolygon``` are not equal is the sense of intersection logic because they are both outers and inners. Also, the ```MultiPolygon``` objects has a lot of useful methods when it comes to determining intersection considering both its inner and outer rings

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: The ```Polygon::overlaps(MultiPolygon``` method handles the intersection and relationship between the `MultiPolygon```'s outers and inners

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: In this case no, because for a multi polygon to intersect with another multi polygon, the outers have to overlap with the other Multipolygon (inners and outs considered), and visa versa.

jklamer: That being said, not sure I did the visa versa part. So I'm going to make a function to do that.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+    };
+    private static final long serialVersionUID = -3474748398986569205L;
+    private List<MultiPolygon> excludeMultiPolygons = new ArrayList<>();
+    private List<Polygon> excludePolygons = new ArrayList<>();
+    private Type filterType;
+    private List<MultiPolygon> includeMultiPolygons = new ArrayList<>();
+    private List<Polygon> includePolygons = new ArrayList<>();
+    private IntersectionDeciding intersectionDeciding;
+
+    public static AtlasEntityPolygonsFilter forConfiguration(final Configuration configuration)
```
AlexHsieh22: Is there any value in a non-configuration based constructor?  Take in a Polygon/MultiPolygon or a set of them and do the same thing?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+            final Map<String, List<String>> includePolygonMap,
+            final Map<String, List<String>> includeMultiPolygonMap,
+            final Map<String, List<String>> excludePolygonMap,
+            final Map<String, List<String>> excludeMultiPolygonMap)
+    {
+        return forConfigurationValues(includePolygonMap, includeMultiPolygonMap, excludePolygonMap,
+                excludeMultiPolygonMap, DEFAULT_INTERSECTION_DECIDING);
+    }
+
+    private static StringConverter<Optional<List<MultiPolygon>>> getMultiPolygonConverterForFormat(
```
AlexHsieh22: There seems to be quite a bit of overlap from getMultiPolygon* and getPolygon* below here.  Would an interface or base class reduce code duplication?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: Thoughts on separating out these classes into separate files?  I think it would make this file more manageable.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: Check out the enum and test cases. Those have those construction methods

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: I wanted to do something of that nature, but because Polygon and Multipolygon  don't share a base class, nor are their converts exactly the same, this was the best solution I could come up with.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: Is this useful as it's own file? Right now this seems like a very niche use. Can you think of other situations where an intersection policy would be useful?

AlexHsieh22: I think if we make a AtlasEntityPolygonsFilter sub package, we can make this file a bit more scoped.  Being a very niche use doesn't mean we can't have things in separate files or in the same file.  Separating into multiple files also manages the code structure.

Currently, this file does:
- implements the AtlasFilter (the main goal of this file)
- handles configuration interaction
- converts all different acceptable formats to be used by this filter
- defines IntersectionPolicy
- defines the filter Type
- defines the PolygonStringFormat

In my opinion, it seems like quite a bit is going on.  I think by splitting this up a bit, the file can be a bit more readable and less intimidating.  Let me know what you think!

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,520 @@
...
+{
+    public static final String EXCLUDED_MULTIPOLYGONS_KEY = "filter.multipolygons.exclude";
+    public static final String EXCLUDED_POLYGONS_KEY = "filter.polygons.exclude";
+    public static final String INCLUDED_MULTIPOLYGONS_KEY = "filter.multipolygons.include";
+    public static final String INCLUDED_POLYGONS_KEY = "filter.polygons.include";
+    private static final Logger logger = LoggerFactory.getLogger(AtlasEntityPolygonsFilter.class);
+    public static final IntersectionDeciding DEFAULT_INTERSECTION_DECIDING = new IntersectionDeciding()
+    {
+        @Override
+        public boolean multiPolygonEntityIntersecting(final MultiPolygon multiPolygon,
```
AlexHsieh22: This seems to be a pretty standard definition of EntityIntersection.  Thoughts on pushing the default to the IntersectionPolicy class?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: We could have a generic base class called ```PolygonStringFormatConverter``` that exposes a ```handleAtlas```, ```handleGeoJson```(and so forth) and have common logic in the base class and implementations will implement the different handlers.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: Under that philosophy I would say, lets move IntersectionPolicy, and the PolygonsStringFormat into their own files. Move the polygon/multipolygon String converters into the PolygonStringFormat enum. I want to keep the Type enum in the class to keep the super simple construction methods.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: Hmm looking at the test cases, it looks like all users would have to do something like this:
```
final String configurationStringFormat = "{\"filter.polygons\":{\"include.wkt\":[\"%s\"],\"exclude.wkt\":[\"%s\"]}}";

AtlasEntityPolygonsFilter.forConfiguration(new StandardConfiguration(new StringResource(String.format(configurationStringFormat, polygonWkt))));
```

If all users are going to be making the same calls, would it be possible to have this common code within the AtlasEntityPolygonsFilter class?  Or if this class implements it, we don't even need to create a configuration.

I may have missed something about the enums/test cases, but let me know!

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: so make IntersectionPolicy an abstract class with a default anonymous implementation?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: I like it - if you're comfortable with it!

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: Hmm..or we could create a DefaultIntersectionPolicy?

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,518 @@
...
+     * @param polygonType
+     *            whether this is an include or exclude Polygon
+     * @return filter for multipolygons
+     */
+    private static Predicate<MultiPolygon> notOverlappingMultipolygon(
+            final QuadTree<Located> vettedPolygons, final Type polygonType)
+    {
+        return multiPolygon ->
+        {
+            if (vettedPolygons.get(multiPolygon.bounds()).stream()
```
snb3300: Here we are iterating over the stream twice. Not sure how much of an impact on runtime this would make. Do you think these can be combined into one iteration?

Maybe have a predicate which checks for instance type and ```multipolygon overlaps```?

src/test/java/org/openstreetmap/atlas/geography/MultiPolygonTest.java
```diff
@@ -80,6 +80,25 @@ public void testCoversPolyLine()
         Assert.assertFalse(multiPolygon.fullyGeometricallyEncloses(jumpyLine));
     }
 
+    @Test
+    public void testOverlap()
```
snb3300: Good test case 👍

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
```diff
@@ -231,6 +233,36 @@ public GeoJsonObject create(final PolyLine polyLine)
...
+     */
+    public GeoJsonObject createFeatureCollection(final Iterable<GeoJsonObject> objects)
+    {
+        final JsonObject result = new JsonObject();
+        result.addProperty(TYPE, FEATURE_COLLECTION);
+        final JsonArray features = new JsonArray();
+        int counter = 0;
+        for (final GeoJsonObject object : objects)
+        {
+            if (this.logFrequency > 0 && ++counter % this.logFrequency == 0)
```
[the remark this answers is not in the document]
jklamer: This function mimics another in this class that I did not write. I am not 100% sure of the utility but I am matching the functionality. This will all be cleaned up in an upcoming PR with improvements to code surrounding Geojson

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
```diff
@@ -0,0 +1,518 @@
...
+    private Predicate<AtlasEntity> isIncluded()
+    {
+        return entity -> this.filterType == Type.INCLUDE && (this.includePolygons.stream().anyMatch(
+                polygon -> this.intersectionPolicy.polygonEntityIntersecting(polygon, entity))
+                || this.includeMultiPolygons.stream()
+                        .anyMatch(multiPolygon -> this.intersectionPolicy
+                                .multiPolygonEntityIntersecting(multiPolygon, entity)));
+    }
+
+    private Predicate<AtlasEntity> isNotExcluded()
```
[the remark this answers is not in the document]
jklamer: Because the filter is only ever one type, there is no consideration of inclusion in this function. This function only returns true if it's an exclude filter and the Entity is not excluded.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: It probably has minimal impact on runtime performance because this happens once during initialization and the quad tree is optimized for these type of spacial queries, and because only not overlapping polygons are stored in the tree, not all polygons will be iterated over.

That being said, it is much better design and practice to make this into a unified ```Predicate<Located>```, however I will need to make two for both Polygons and MultiPolygons which will increase the code duplication that I know is bothering @AlexHsieh22

AlexHsieh22:
Random thought - could be stupid - is there a reason why we don't convert all Polygons to MultiPolygons using the ```MutliPolygon.forPolygon``` method?  Then we can Polygons and MultiPolygons the same way.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
AlexHsieh22: haha I'm working on getting over it :) also thinking of ways to simplify this more! Overall, very grateful for the work being done here!

jklamer:
The same reason we don't do that everywhere I imagine? My guess: Polygon is a much simpler implementation and more popular as a map object. That being said, Geojson Polygon implementation can geometrically translate directly to both our Polygon and MultiPolygon. We are going to have to blur the lines between the two when refactoring our GeoJson reading/writing code, and I think that would be a great time talk about that relationship.

src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
jklamer: I was wrong. I was able to avoid too much duplicated code by nulling inputs in a more general function. The duplicated code I was less able to avoid is in PolygonStringFormat enum now.

AlexHsieh22:
Great question @jklamer .  I was thinking about this question in relation to your PR.  I think there's value in having both a Polygon and a MultiPolygon class because MultiPolygon's are messier (holes in them), whereas Polygon's are a bit more straightforward typically.  Since we already have the code for MultiPolygon in your case here, would converting all Polygon's to MultiPolygon's simplify things?

For the geojson processing, we could do something like this: (no guarantees on the code working :))
```
final List<MultiPolygon> multiPolygons = new ArrayList<>();
final GeoJsonReader reader = new GeoJsonReader(new StringResource(string));
while (reader.hasNext())
{
    final PropertiesLocated propertiesLocated = reader.next();
    if (propertiesLocated.getItem() instanceof MultiPolygon)
    {
        multiPolygons.add((MultiPolygon) propertiesLocated.getItem());
    }
    else if(propertiesLocated.getItem() instance of Polygon)
    {
        multiPolygons.add(MultiPolygon.forPolygon((Polygon) propertiesLocated.getItem()));
    }
    else
    {
        logger.warn("MultiPolygon Filter does not support item {}", propertiesLocated.toString());
    }
}
return Optional.of(multiPolygons).filter(list -> !list.isEmpty());
```

jklamer:
Yes it would make the code simpler but it would honestly makes me super uncomfortable for a couple of reasons:
1) They are not the same object, and treating one like the other seems wrong. I can't site certain design principles.
2) We broke out a lot this code for the purpose of making things easier for future editors, and if the relationship between Polygon and Multipolyon were to ever change, someone would have to add all the functions for Polygons back into the code just to maintain the same functionality.
3) This would make it harder to debug because if a user were to pass in a Polygon and an error occurs on a MultiPolygon then that seems counter to having accurate as possible logging messages. 
4) This would complicate custom intersecting logic for the user especially because Polygon has many more/better tested utility functions.

jklamer:
I would be much more comfortable with ( and would prefer it to the current), having them both implement an interface called something to the effect of "GeometricArea" then using the contracts established there.

AlexHsieh22:
Thanks @mgcuthbert .  Is there any action towards @jklamer 's comment of:
> I would be much more comfortable with ( and would prefer it to the current), having them both implement an interface called something to the effect of "GeometricArea" then using the contracts established there.

AlexHsieh22:
@jklamer I never got back to your great response on why you're not comfortable with MultiPolygon instead of Polygon.  It sounds like overall, you feel (with good reason) there's a need for Polygon, and so I'm okay if you want to keep it that way.  

In terms of MultiPolygon, it is a separate class, but when the MultiPolygon.forPolygon method is called, we're creating a MultiPolygon with 1 outer and no inners.  This MultiPolygon implements Iterable<Polygon>, and also exposes the outers api to get the underlying Polygon back if we needed.  All in all, the atlas code here is pretty flexible.  Furthermore, the relationship between fundamental classes like Polygon and MultiPolygon shouldn't change - especially because a MultiPolygon is made up of Polygons.

Overall, nice addition, testing, enthusiasm, and work. Very glad you've spent lots of time thinking about this as well!

PR #90, opened 2018-02-27
Improve CountryBoundaryMap dynamic grid index generation performance and fix serialization

mertemin:
This PR makes significant updates to `CountryBoundaryMap`. Grid index cells generated by `DynamicGridIndexBuilder` are used in `CountryBoundaryMap` for Atlas slicing operation (slicing features that cross country boundaries). Right now grid index is not properly serialized, hence populated in Atlas generation process. Also `CountryBoundaryMap` API has grown a lot and not easy to work with. `slice` method could be subject to random inconsistency issues, because it is slicing features by the order of results coming from R tree.

Here is a quick summary of changes:

**CountryBoundaryMap**
- `CountryBoundaryMap` had several constructors and entry points for clients (existing `CountryBoundaryMap` objects, `Atlas` objects, shp files). These entry points were constrained by the bounds of operation. That resulted in several constructors. Now there is two entry points: default `CountryBoundaryMap` that doesn't have bound limits and one with bounds. In case we have an existing artifact like Atlas files or shp file, these artifacts can go through helper methods.
- Dropped all `createGridIndex` methods and changed them to `initializeGridIndex`.  Additional parameter `saveGridIndexCells` to collect grid cells is dropped as well. Now grid index is initialized with `initializeGridIndex` method and grid index cells can be extracted from `STRtree` without additional data structure. Collecting grid index cells from `STRtree` is done via `collectCells` method.
- `slice` method is updated for consistency. Cells returned from grid index are first sorted by country name before slicing operation to avoid randomness in the slicing results. Country name that comes first alphabetically will be used first for slicing.
- Read and write operations are updated and grid index cells are now saved to files and can be read back to populate grid index.
- Now that read/write operations of grid index cells are fixed, grid index generation operations in `OsmPbfProcessor` and `RawAtlasCountrySlicer` are removed.
- Useful helper methods are added to `CountryBoundaryMap` like `geometries` to read inner geometries of a given geometry collection, `boundariesHelper` to easily filter given geometries (mostly used in query operations.
- `removeDuplicate` is removed and replaced with `stream().distinct()` call.

**DynamicGridIndexBuilder**
- `GridWorkItem` and `Processor` inner classes introduced to generate grid index cells in parallel.
- Logic that generates grid cells stay the same.
- Extra data structure to cache grid cells is dropped, since now cells can be collected from `STRtree`.
- Test class is updated and new tests are added.

**Other classes**
- `CountryBoundaryMapCompareCommand` to compare different boundary maps. The order of countries could be different, but that should not affect the tree structure. This command is useful to make sure that.
- Classes that create `CountryBoundaryMap` instances are all updated.
- Boundary files with grid indices are all replaced with new files.

After these changes explained above, grid index generation performance improves significantly on multiple cores (i.e. a shp file with world countries generation time improved ~20x). Integration tests in `OsmPbfToSlicedRawAtlasTest` class is now running 30% faster. Atlas generation performance improves as well given that grid index is pre-populated and ready to use.


src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/OsmPbfToSlicedRawAtlasTest.java
```diff
@@ -94,7 +94,7 @@ public void testPbfToSlicedRawAtlas()
         Assert.assertEquals(0, slicedRawAtlas.numberOfEdges());
         Assert.assertEquals(0, slicedRawAtlas.numberOfAreas());
         Assert.assertEquals(648704, slicedRawAtlas.numberOfPoints());
-        Assert.assertEquals(57644, slicedRawAtlas.numberOfLines());
```
mertemin: Country borders can overlap each other. After the update in `slice` method, the country that comes first alphabetically is used for slicing first. That caused a change and one feature is now sliced once not twice. See image below for example scenario
screen shot 2018-02-27 at 2 22 54 pm

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMapArchiver.java
```diff
@@ -25,20 +27,41 @@
...
+            "Indicator whether to create a spatial grid index and include that in the output.",
             Boolean::parseBoolean, Optionality.OPTIONAL, Boolean.FALSE.toString());
 
+    private static final Function<Iterable<String>, Predicate<String>> COUNTRY_FILTER_GENERATOR = countryList ->
+    {
+        return countryLine ->
+        {
+            final String readCountryName = StringList
+                    .split(countryLine, CountryBoundaryMap.COUNTRY_BOUNDARY_DELIMITER).get(0);
+            for (final String countryName : countryList)
```
snb3300: Instead of for loop how about
``` return countryList.stream().anyMatch(countryName -> countryName.equals(readCountryName))```

File: src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java [hunk 3]
```diff
@@ -160,6 +158,87 @@
...
+     *
+     * @param root
+     *            Root node for collection
+     * @param cells
+     *            {@link MultiMap} to save cells into
+     */
+    @SuppressWarnings("unchecked")
+    static void collectCells(final AbstractNode root, final MultiMap<Geometry, Envelope> cells)
+    {
+        if (root.getLevel() > 0)
```
snb3300: If root level is always a whole number then how about
```
int rootLevel = root.getLevel();
if (rootLevel == 0)
{
    root.getChildBoundables().stream().forEach(item ->
            {
                final ItemBoundable boundable = (ItemBoundable) item;
                final Geometry polygon = (Geometry) boundable.getItem();
                final Envelope bounds = (Envelope) boundable.getBounds();
                cells.add(polygon, bounds);
            });
}
else
{
    root.getChildBoundables().stream().forEach(node ->
            {
                collectCells((AbstractNode) node, cells);
            });
}
```

I feel having an if, else would be much cleaner than if else if. What do you think?

mertemin: Theoretically root level should not be negative. Still If/else block right now is defensive. So I'd prefer to leave as-is.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMapArchiver.java
mertemin: Good point, I'll look into a way to improve this block of code.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
```diff
@@ -121,14 +119,6 @@ public void initialize(final Map<String, Object> metaData)
             logger.trace("Start loading atlas with polygon or bound {}", this.multiPolygon);
             logger.trace("With country code {}",
                     String.join(",", this.loadingOption.getCountryCodes()));
-
-            if (this.multiPolygon != null && this.loadingOption.getCountryBoundaryMap() != null)
-            {
-                logger.trace("Building grid index...");
```
matthieun: Does that mean that we would not support any country boundary map object that has not been pre-processed with grid index?

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -836,23 +987,81 @@ public CountryCodeProperties getCountryCodeISO3(final Geometry geometry, final b
      */
     public CountryCodeProperties getCountryCodeISO3(final Location location)
     {
-        return getCountryCodeISO3(JTS_POINT_CONVERTER.convert(location));
+        return this.getCountryCodeISO3(JTS_POINT_CONVERTER.convert(location));
+    }
+
+    /**
+     * @return the raw {@link STRtree} index used by this {@link CountryBoundaryMap}
```
matthieun: Copy paste error here?

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
```diff
@@ -1,40 +1,245 @@
...
+ * @author mkalender
  */
 public class DynamicGridIndexBuilder extends AbstractGridIndexBuilder
 {
+    /**
+     * A class to hold units of work while building grid index.
+     *
+     * @author mkalender
+     */
+    class GridWorkItem
```
matthieun: If those inner classes are not used outside, I would keep them private instead of package-protected, no?

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
```diff
@@ -1,40 +1,245 @@
...
+            {
+                while (!this.queue.isEmpty())
+                {
+                    final GridWorkItem itemToProcess = this.queue.poll();
+
+                    // Queue might have been emptied by another thread. That gives an null item.
+                    // Check for null item to avoid an exception.
+                    if (itemToProcess != null)
+                    {
+                        this.process(itemToProcess.minX, itemToProcess.minY, itemToProcess.maxX,
```
matthieun: How about passing the grid work item directly? Just a suggestion.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
mertemin: Nope. Once this is merged, I'll send a PR for atlas-generator to handle boundary map files without grid index and boundary map files with older version of grid index.

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
mertemin: I'll make them private.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
mertemin: Yeah

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
mertemin: Yeah, that'll be cleaner.

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfProcessor.java
mertemin: See https://github.com/osmlab/atlas-generator/pull/15

matthieun: So that means any local test with a boundary file manually created that does not contain the grid index would not work anymore? I agree the changes in atlas-generator make sense, but I would still like to have the option to test locally with a boundary file without grid index and have the grid index created on the fly when the slice operation starts. Is that possible? And maybe with a big warning saying the grid index is being re-created?

matthieun: I'd like to have @MikeGost's thoughts on this too... Maybe I am not seeing the whole picture here...

mertemin: I went through tests and made sure tests need grid index either work with boundary map file with grid index or generate grid index with `initializeGridIndex` call. You can now retrieve grid index by `getGridIndex` method.

matthieun: Understood, thanks!

MikeGost: I think this addresses all use cases - running with a boundary file with and without grid index serialization is supported.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1300,230 +1425,111 @@ private Geometry nearestNeighbour(final Envelope envelope, final Object object,
...
-                                .read(wktTokenizer.nextToken());
-                        addCountry(countryISO, multiPolygon);
-                    }
+                    writer.write(System.lineSeparator());
                 }
                 catch (final Exception e)
                 {
-                    logger.error("Invalid country boundary text file format.", e);
-                    e.printStackTrace();
+                    logger.error(String.format("Failed to write cells for {}.", country));
```
MikeGost: How much value is a grid index without data for a whole country or multiple countries? From above code, it doesn't look like we'll generate the missing portions, will be fall back on the raw index at that point and slow down overall processing? If that's the case, it might not make sense to generate a partially complete grid index here. Thoughts on throwing instead of logging?

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMapCompareCommand.java
```diff
@@ -0,0 +1,189 @@
...
+import com.vividsolutions.jts.geom.Envelope;
+import com.vividsolutions.jts.geom.Geometry;
+import com.vividsolutions.jts.index.strtree.STRtree;
+
+/**
+ * Compares given {@link CountryBoundaryMap}s to find if there is any differences between them.
+ *
+ * @author mkalender
+ */
+public class CountryBoundaryMapCompareCommand extends Command
```
MikeGost: Nice tool! This will be useful in the future.

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
```diff
@@ -1,40 +1,223 @@
...
+                    final GridWorkItem itemToProcess = this.queue.poll();
+
+                    // Queue might have been emptied by another thread. That gives an null item.
+                    // Check for null item to avoid an exception.
+                    if (itemToProcess != null)
+                    {
+                        this.process(itemToProcess);
+                    }
+                }
+
```
MikeGost: Nit: extra line

src/test/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilderTest.java
```diff
@@ -0,0 +1,133 @@
...
+ *
+ * @author mkalender
+ */
+public class DynamicGridIndexBuilderTest
+{
+    protected static final Logger logger = LoggerFactory
+            .getLogger(DynamicGridIndexBuilderTest.class);
+
+    @Test
+    public void testConsistencyOne()
```
MikeGost: How about a private method for the common code that takes input of countries and boundary file, then we can expand consistency to any number of combinations. Otherwise the code is almost identical.

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
```diff
@@ -1,40 +1,223 @@
...
+                        countryCode, e);
+                // Save the current box and do not iterate down.
+                synchronized (this.index)
+                {
+                    this.index.insert(box, polygon);
+                }
+            }
+            catch (final Exception e)
+            {
+                logger.error("Processor failed.", e);
```
MikeGost: What's the impact of failure here? Corrupted grid index?

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
AlexHsieh22: can we possibly change ```root``` -> ```inputNode```?  Root typically means level = 0 (or in the STRTree, max level).  Sometimes collectCell can take in a root node, but most times, it'll be a non-root node.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1190,12 +1315,32 @@ else if (geometry instanceof LineString || geometry instanceof Polygon)
         }
     }
 
+    private List<CountryBoundary> boundariesHelper(final Supplier<List<Polygon>> supplier,
```
AlexHsieh22: this seems to be a critical function for many of the methods above. Could we get a comment for this function?

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -254,123 +322,352 @@ public static void setGeometryProperty(final Geometry geometry, final String key
...
+            {
+                try
+                {
+                    final Geometry envelope = reader
+                            .read(line.substring(GRID_ENVELOPE_DELIMITER.length()));
+                    this.gridIndexEnvelope = envelope.getEnvelopeInternal();
+                }
+                catch (final ParseException e)
+                {
+                    logger.error("Failed to read grid index envelope.");
```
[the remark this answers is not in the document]
mertemin: We should better throw here. Although some functionality might work without this, it is safe to stop and figure out what went wrong. I'll update the code.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -254,123 +322,352 @@ public static void setGeometryProperty(final Geometry geometry, final String key
...
+                    while (geometryTokenizer.hasMoreTokens())
+                    {
+                        final String polygonString = geometryTokenizer.nextToken();
+                        final Geometry geometry = reader.read(polygonString);
+                        setGeometryProperty(geometry, ISOCountryTag.KEY, country);
+
+                        // NOTE: This is for backward compatibility. Older boundary maps save
+                        // MultiPolygons, but newer ones save Polygons and use the order per country
+                        // for grid indexing.
+                        if (geometry instanceof Polygon)
```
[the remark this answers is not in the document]
mertemin: I thought about this. There are several different dependencies within repo that needs an update. I think this is good for now. Once we transition all dependencies to the new format, verify results, we can get rid of this part.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -836,23 +987,81 @@ public CountryCodeProperties getCountryCodeISO3(final Geometry geometry, final b
...
-        return getCountryCodeISO3(JTS_POINT_CONVERTER.convert(location));
+        return this.getCountryCodeISO3(JTS_POINT_CONVERTER.convert(location));
+    }
+
+    /**
+     * @return {@link STRtree} grid index used by this {@link CountryBoundaryMap}
+     */
+    public STRtree getGridIndex()
+    {
+        return this.gridIndex;
```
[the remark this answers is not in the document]
mertemin: I agree on exposing it part, but that exposure is very limited and contained within small set of classes. Tree that being returned here could be big. Passing around copies could be safe, but expensive. 

Anyways classes, functionality around grid index and country boundary map need further cleanup. This can be addressed then I think.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1300,230 +1425,111 @@ private Geometry nearestNeighbour(final Envelope envelope, final Object object,
...
-                                    .read(gridIndexGeometryTokenizer.nextToken());
-                            gridIndex.insert(geometry.getEnvelopeInternal(), geometry);
+                            writer.write(GRID_INDEX_DELIMITER);
+                            writer.write(WKT_WRITER
+                                    .write(DynamicGridIndexBuilder.buildGeoBox(cell.getMinX(),
+                                            cell.getMaxX(), cell.getMinY(), cell.getMaxY())));
+                        }
+                        catch (final Exception e)
+                        {
+                            logger.error(
```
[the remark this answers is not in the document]
mertemin: Throwing should be safer here.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
mertemin: Throwing here as well sounds good.

src/main/java/org/openstreetmap/atlas/geography/boundary/DynamicGridIndexBuilder.java
mertemin: Yes, possibly corrupted index. I'll let it throw here.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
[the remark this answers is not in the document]
MikeGost: Anyone who has access to the CountryBoundaryMap can call this, which is any client who creates an Atlas. How about adding a comment in the javadoc that warns people against modifying the returned STRtree, as that can impact any slicing that happens.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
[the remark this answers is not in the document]
MikeGost: Fair enough. Maybe a TODO to remove the backward compatibility later? :)

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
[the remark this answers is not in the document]
mertemin: I'll limit it's scope and have a comment.

PR #92, opened 2018-03-07
Fix hasGridIndex behavior

mertemin:
This PR is fixing `hasGridIndex` method behavior. It should check if `gridIndex` is not null. I updated tests to utilize that method.


MikeGost:
Belated LGTM.

PR #93, opened 2018-03-07
Cross Shard Raw Atlas Sectioning

MikeGost:
Part of #17, featuring a few changes:
1. Updating `WaySectionProcessor` to handle sectioning given a `Shard`, `Sharding` and `Function` to fetch an Atlas for a given Shard.
2. Corresponding unit test
3. Updating `SlippyTile` to implement `Comparable` and corresponding unit test
4. Wrote a multi-shard unit test that will section consistently irrespective of which shard we start at. See diagram below:

img_1289

Next and final step: Update Atlas-Generator project to make use of the raw Atlas flow.


src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
```diff
@@ -182,6 +182,49 @@ public Rectangle bounds()
         return this.bounds;
     }
 
+    @Override
+    public int compareTo(final SlippyTile other)
```
matthieun: Where is that comparison used?

MikeGost: It's not used at this time, I was planning on using it when storing the processed Shards, but ended up not doing it. Might still prove to be useful so left it in.

matthieun: Ok, thanks!

PR #95, opened 2018-03-19
Return results if there is a full overlap in slicing process

mertemin:
This PR addresses an edge case in CountryBoundaryMap’s slicing logic. When there is an overlap between country borders and a feature is fully covered by one country, then we’d detect it and assign that country’s code to the feature. However, after detecting we continue to slice the feature. Other country’s (one that is NOT fully covering feature) boundary slices the geometry and tries to assign a country code, but that fails. That assigns sliced geometries to the same country, although we can just return the full geometry and be done with the slicing process.

The actual changes is happening on line 1186 in CountryBoundaryMap file - https://github.com/osmlab/atlas/pull/95/files#diff-624909305b238ebca80c0e72370bd906L1186

Other changes are minor improvements, like removing the list that was holding intersecting candidates `intersected`, renaming variables to make things more clear. `OsmPbfLoaderTest` is updated so that we hit the scenario above. Tests are updated to avoid resource leaks.

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderTest.java
@@ -140,10 +140,8 @@ public class OsmPbfLoaderTest
         {
             final OsmPbfLoader osmPbfLoader = new OsmPbfLoader(() -> osmosis,
                     // Here reduce the loading bounds to a small area around Edge 3. In this case
-                    // Node 3
-                    // is included in the country boundary but not in the loading area. It should
-                    // not
-                    // have any boundary tag.
+                    // Node 3 is included in the country boundary but not in the loading area. It
+                    // should not have any boundary tag.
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/pbf/OsmPbfLoaderTest.java
```diff
@@ -111,70 +128,84 @@ public void init()
...
-        final Atlas atlas = osmPbfLoader.read();
-        logger.info("{}", atlas);
-        final Edge edgeIn = atlas.edgesIntersecting(DE_ANZA_AT_280.bounds()).iterator().next();
-        final Node nodeIn = edgeIn.end();
-        Assert.assertNull(nodeIn.tag(SyntheticBoundaryNodeTag.KEY));
+        try (OsmosisReaderMock osmosis = new OsmosisReaderMock(this.store))
+        {
+            final OsmPbfLoader osmPbfLoader = new OsmPbfLoader(() -> osmosis,
+                    // Here reduce the loading bounds to a small area around Edge 3. In this case
+                    // Node 3
```
MikeGost: Nit: can you fix the formatting please

PR #96, opened 2018-03-20
Fix for DynamicTileSharding OOM

jwpgage:
This PR adds a memoization step to `DynamicTileSharding` which means that counts for each `SlippyTile` only needs to be calculated once. This keeps memory consumption low on zoom levels higher than 15, fixing an OOM issue.


src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -126,20 +137,29 @@ protected SwitchList switches()
...
 
+    /**
+     * Writes to file all the {@link SlippyTile} feature counts at a certain zoom level.
+     *
+     * @param countsAtZoom
+     *            HashMap containing the feature counts at that zoom level.
+     * @param bufferedWriter
+     *            The writer to write to.
+     */
+
```
lucaspcram: nit: you've got a stray newline here :)

jwpgage: Good eye, thanks!

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+import org.openstreetmap.atlas.streaming.resource.WritableResource;
+import org.openstreetmap.atlas.utilities.collections.StringList;
+import org.openstreetmap.atlas.utilities.runtime.Command;
+import org.openstreetmap.atlas.utilities.runtime.CommandMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Command that takes a file with feature counts for each {@link SlippyTile} at a certain zoom
+ * level, and then calculates summed counts for every lower zoom layer to the file. This is meant to
```
MikeGost: Nit: the wording here is a little awkward with "every lower zoom layer to the file" - do you mind updating?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+ * be used as a pre-processing step before creating the sharding tree in
+ * {@link DynamicTileSharding}.
+ *
+ * @author james-gage
+ * @author lcram
+ */
+public class AppendCountsForAllZoom extends Command
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+    public static final Switch<WritableResource> DEFINITION = new Switch<>("definition",
```
MikeGost: If we're just reading this, why is it a `WritableResource`?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+            "Resource containing tile to feature count mapping for zoom 0 to maxZoom-1.", File::new,
+            Optionality.REQUIRED);
+    public static final Switch<WritableResource> OUTPUT = new Switch<>("output",
+            "The resource where to save the feature counts for all zooms.", File::new,
+            Optionality.REQUIRED);
+    private static final Switch<Integer> ZOOM = new Switch<>("zoom",
+            "The zoom layer the input csv represents", Integer::valueOf, Optionality.REQUIRED);
```
MikeGost: Nit: from this description, it's a little unclear what the CSV input is - is it the definition file? If so, can you add that to the description please?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+
+    private int linesInFile(final WritableResource definition)
+    {
+        int numberLines = 0;
+        for (@SuppressWarnings("unused")
+        final String line : definition.lines())
+        {
+            numberLines++;
+        }
+        logger.info("There are {} tiles.", numberLines);
```
MikeGost: Nit: this may be slightly misleading - are we counting lines or tiles? Looks like each line is a separate tile, but this isn't apparent from the logging.

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+    protected SwitchList switches()
+    {
+        return new SwitchList().with(DEFINITION, OUTPUT, ZOOM);
+    }
+
+    private int linesInFile(final WritableResource definition)
+    {
+        int numberLines = 0;
+        for (@SuppressWarnings("unused")
+        final String line : definition.lines())
```
MikeGost: Thoughts on `return Iterables.size(definition.lines());` ?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+    {
+        final Integer zoom = (Integer) command.get(ZOOM);
+        final WritableResource definition = (WritableResource) command.get(DEFINITION);
+        final WritableResource output = (WritableResource) command.get(OUTPUT);
+        final BufferedWriter bufferedWriter = output.writer();
+
+        final int numberLines = linesInFile(definition);
+        HashMap<SlippyTile, Long> countsAtZoom = new HashMap<>(numberLines);
+        int counter = 0;
+        for (final String line : definition.lines())
```
MikeGost: This works well! Another alternative is to use `CsvReader` here.

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /**
+     * Writes to file all the {@link SlippyTile} feature counts at a certain zoom level.
+     *
+     * @param countsAtZoom
+     *            HashMap containing the feature counts at that zoom level.
+     * @param bufferedWriter
+     *            The writer to write to.
+     */
+    private void writeToFile(final HashMap<SlippyTile, Long> countsAtZoom,
```
MikeGost: Can we use the interface `Map` here instead of `HashMap` implementation?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+        countsAtZoom.forEach((tile, count) ->
+        {
+            try
+            {
+                bufferedWriter.write(tile.getName() + "," + count);
+                bufferedWriter.write(lineSeparator);
+            }
+            catch (final IOException e)
+            {
+                logger.error("Error writing converted data to {}", bufferedWriter);
```
MikeGost: Does it make sense to throw here? What happens if we don't write for a specific zoom?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+            final SlippyTile tile = SlippyTile.forName(split.get(0));
+            countsAtZoom.put(tile, Long.valueOf(split.get(1)));
+            if (++counter % READER_REPORT_FREQUENCY == 0)
+            {
+                logger.info("Read counts for {} zoom level {} tiles.", counter, zoom);
+            }
+        }
+        writeToFile(countsAtZoom, bufferedWriter);
+
+        for (int i = zoom - 1; i >= 0; i--)
```
MikeGost: Since we're using `i`inside the loop, can we give it a better name - maybe `zoomLevel`?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+     * Generates the counts for the nth zoom layer based on information from zoom layer n+1.
+     *
+     * @param zoomLayerToGenerate
+     *            the zoom layer for which to generate counts
+     * @param countsAtHigherZoom
+     *            HashMap containing counts for all {@link SlippyTile} in zoomLayerToGenerate+1
+     * @return a HashMap containing counts for all (@link SlippyTile} in zoomLayerToGenerate
+     */
+    private HashMap<SlippyTile, Long> calculateTileCountsForZoom(final int zoomLayerToGenerate,
+            final HashMap<SlippyTile, Long> countsAtHigherZoom)
```
MikeGost: Can we use the `Map` interface here as well?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
```diff
@@ -0,0 +1,179 @@
...
+
+        for (int i = zoom - 1; i >= 0; i--)
+        {
+            countsAtZoom = calculateTileCountsForZoom(i, countsAtZoom);
+            writeToFile(countsAtZoom, bufferedWriter);
+            logger.info("Wrote tiles for zoom {} to file!", i);
+        }
+        try
+        {
+            bufferedWriter.flush();
```
MikeGost: Thoughts on putting the entire function into the try block? That way we can close the bufferedWriter in the finally and make sure we never leak resources?

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
jwpgage: You're right, it does. If we don't write for a specific zoom the output of this job will cause `DynamicTileSharding` to fail.

jwpgage:
This change will break the current sharding tree pipeline, as `DynamicTileSharding` will now expect information about all zoom levels, not just the nth one. Therefore this should be merged at the same time an `AppendCountsForAllZoom` module is inserted into the sharding tree pipeline, directly before `DynamicTileSharding`.

src/main/java/org/openstreetmap/atlas/geography/sharding/AppendCountsForAllZoom.java
jwpgage: I'm open to it, but isn't the bufferedWriter already closed in the finally?

MikeGost: Maybe I mis-read! Here's what I was looking at: the buffered writer gets initialized on line 51, we write to file with it on line 66, potentially throwing at that point with the updates. If we throw, then we never hit the try block in the first place. Was thinking we could either initialize it with try-with-resources statement or null and initialize inside the catch block. That way, no matter what happens, we can always close it cleanly. Thoughts?

matthieun:
Closing, we can re-open when ready!

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -448,6 +447,55 @@ protected SwitchList switches()
...
+    private HashMap<SlippyTile, Long> calculateTileCountsForAllZoom(
+            final int firstZoomLayerToGenerate, final HashMap<SlippyTile, Long> counts)
+    {
+        for (int currentZoom = firstZoomLayerToGenerate; currentZoom >= 0; currentZoom--)
+        {
+            long count = 0;
+            long tilesCalculated = 0;
+            for (int x = 0; x < Math.pow(2, currentZoom + 1); x += 2)
+            {
+                for (int y = 0; y < Math.pow(2, currentZoom + 1); y += 2)
```
lucaspcram: Just a thought: `Math.pow()` can often be a pretty slow function, since it must work for the general case (doubles) using log-based formulas under the hood. In specific cases it can be worth it to roll a custom `pow()` function if you know you're only operating on integers. That being said it's not likely to be a performance bottleneck here, since the outer zoom loop is not particularly tight.

As another note: hopefully the JIT compiler is optimizing those `pow()` calls out of the inner loop conditions so that it is not called on every iteration (considering you've got nested loops here). I would imagine it's smart enough to do this, so this note is probably moot.

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -389,13 +389,16 @@ protected int onRun(final CommandMap command)
...
             final SlippyTile tile = SlippyTile.forName(split.get(0));
-            counts.put(tile, Integer.valueOf(split.get(1)));
+            counts.put(tile, Long.valueOf(split.get(1)));
             zoom = tile.getZoom();
             if (++counter % READER_REPORT_FREQUENCY == 0)
             {
                 logger.info("Read counts for {} zoom level {} tiles.", counter, zoom);
             }
         }
+        final HashMap<SlippyTile, Long> allCounts = calculateTileCountsForAllZoom(maximumZoom - 2,
```
matthieun: Can you add a code comment here to explain what the `2` means?

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -377,7 +377,7 @@ protected int onRun(final CommandMap command)
             numberLines++;
         }
         logger.info("There are {} tiles.", numberLines);
-        final Map<SlippyTile, Integer> counts = new HashMap<>(numberLines);
+        HashMap<SlippyTile, Long> counts = new HashMap<>(numberLines);
```
MikeGost: Nit: why did this become a HashMap instead of Map?

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -448,6 +447,55 @@ protected SwitchList switches()
...
+     * @param firstZoomLayerToGenerate
+     *            the first zoom layer for which to generate counts
+     * @param countsAtHigherZoom
+     *            HashMap containing counts for all {@link SlippyTile}s in
+     *            firstZoomLayerToGenerate+1
+     * @return a HashMap containing counts for all (@link SlippyTile}s in zoomLayerToGenerate and
+     *         below
+     */
+    private HashMap<SlippyTile, Long> calculateTileCountsForAllZoom(
+            final int firstZoomLayerToGenerate, final HashMap<SlippyTile, Long> counts)
```
MikeGost: Nit: Any specific reason to use `HashMap` implementation of `Map` for the return value and input?

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
jwpgage: Added!

PR #97, opened 2018-03-20
Fix for SQL file generation OOM

jwpgage:
This PR fixes an OOM issue in `TilePrinter`. Previously, all the tiles for a certain zoom level were calculated, and then as they were written to SQL each tile's bounds were calculated and stored. GC tried to collect these `Rectangle`s, but couldn't because they were still referenced by the tile object. The method `freeBoundsForGC` nulls the bounds and allows them to be collected, fixing the OOM.

UPDATE:

After Mert's comment I've refactored the fix. Now instead creating an `Iterable` of all slippy tiles at a certain zoom, we instead create an `Iterator` directly to only load a shard at a time. This will use even less memory than the original, while still solving the `GC overhead limit` issue. I implemented this by adding a method `allTilesIterator(final int zoom, final Rectangle bounds)` which can be called directly and is now used by `allTiles`. The method `freeBoundsForGC` has been removed.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
@@ -70,13 +70,37 @@ public class SlippyTile implements Shard, Comparable<SlippyTile>
         {
             throw new CoreException("Zoom too large.");
         }
+        final Iterable<SlippyTile> result = () -> allTilesIterator(zoom, bounds);
+        final List<SlippyTile> list = Iterables.asList(result);
+        if (list.isEmpty())
+        {
+            throw new CoreException("List cannot be empty");
+        }
+        return list;
+    }
+
+    /**
+     * Iterator for all tiles within some bounds
+     *
+     * @param zoom
+     *            The zoom to consider
+     * @param bounds
+     *            The bounds to consider
+     * @return Iterator for all tiles within some bounds
+     */
+    public static Iterator<SlippyTile> allTilesIterator(final int zoom, final Rectangle bounds)
+    {
+        if (zoom > MAX_ZOOM)
+        {
+            throw new CoreException("Zoom too large.");
+        }
         final SlippyTile lowerLeft = new SlippyTile(bounds.lowerLeft(), zoom);
         final SlippyTile upperRight = new SlippyTile(bounds.upperRight(), zoom);
         final int minX = lowerLeft.getX();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/sharding/preparation/TilePrinter.java
```diff
@@ -83,6 +83,7 @@ protected int onRun(final CommandMap command)
                     writer.write(";");
                 }
                 writer.write("\n");
+                tile.freeBoundsForGC();
```
MikeGost: Do you mind adding a quick comment, similar to the PR description about why this is needed her.

src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
```diff
@@ -237,6 +237,11 @@ public boolean equals(final Object other)
         return false;
     }
 
+    public void freeBoundsForGC()
```
AlexHsieh22: consider generalizing this to ```freeBounds()```?

mertemin:
Changes here could easily hit memory limit again. I'd suggest looking into `allTiles` method and returning an `Iterator`(probably through a new method) and processing tiles as you go. Such approach would decrease memory required to process tiles by a lot.

Otherwise, `allTiles` method would keep creating an iterator and and collecting all the tiles before returning. That is where the most of the memory spent. And then inside `TilePrinter` iterator being accessed again.

jwpgage:
@mertemin Thanks for the feedback! The issue actually wasn't directly due to OOM, it could load `allTiles` successfully but was hitting the GC overhead limit trying to deallocate the bounds objects in the loop unsuccessfully. 

I agree with you though, your solution is more elegant and should use less memory overall. I'll make the change!

src/main/java/org/openstreetmap/atlas/geography/sharding/SlippyTile.java
```diff
@@ -65,6 +65,30 @@
      * @return All tiles within some bounds
      */
     public static Iterable<SlippyTile> allTiles(final int zoom, final Rectangle bounds)
+    {
+        if (zoom > MAX_ZOOM)
+        {
+            throw new CoreException("Zoom too large.");
```
MikeGost: Nit: we should probably add the zoom here to make debugging a little easier, but this is super minor. Maybe you can update this in the other open PR, going to merge as is.

PR #100, opened 2018-03-22
GeometricSurface Interface added + Utilities

jklamer:
Adding in the GeometricSurface Interface to handle geometric surface interactions (namely Polygon and Multipolygon). Then used the interface to expand the usability of AtlasEntities and clean up the AtlasEntityPolygonFilter
Things to note:

- ```AtlasEntity::intersects(Polygon)``` ->  ```AtlaEntity::intersects(GeometricSurface)``` 

- I implemented ```MultiPolygon::fullyGeometricallyEncloses(MultiPolygon)```, ~~~```MultiPolygon::fullyGeometricallyEncloses(Segment)```~~~ , and ```MultiPolygon::surfaceOnSphere()```
- Backwards Compatible
- Some utility added in QuadTree

Some questions I have:
- Should I get ride of ```IntersectionPolicy ```  interface and replace it with a ```BiPredicate<GeometricSurface,AtlasEntity>``` ?
- Are there more functions to add to the Interface ?
- Should I allow overlapping GeometricSurfaces in AtlasEntityPolygonFilter ?

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
@@ -17,7 +17,7 @@ import org.openstreetmap.atlas.geography.converters.WktMultiPolygonConverter;
 import org.openstreetmap.atlas.geography.geojson.GeoJsonBuilder;
 import org.openstreetmap.atlas.geography.geojson.GeoJsonBuilder.LocationIterableProperties;
 import org.openstreetmap.atlas.geography.geojson.GeoJsonObject;
-import org.openstreetmap.atlas.geography.index.QuadTree;
+import org.openstreetmap.atlas.geography.index.RTree;
 import org.openstreetmap.atlas.streaming.resource.WritableResource;
 import org.openstreetmap.atlas.streaming.writers.JsonWriter;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
@@ -31,7 +31,7 @@ import org.openstreetmap.atlas.utilities.scalars.Surface;
  *
  * @author matthieun
  */
-public class MultiPolygon implements Iterable<Polygon>, GeometricSurface, Located, Serializable
+public class MultiPolygon implements Iterable<Polygon>, GeometricSurface, Serializable
 {
     private static final long serialVersionUID = 4198234682870043547L;
     private static final int SIMPLE_STRING_LENGTH = 200;
@@ -245,7 +245,7 @@ public class MultiPolygon implements Iterable<Polygon>, GeometricSurface, Locate
     @Override
     public boolean fullyGeometricallyEncloses(final MultiPolygon that)
     {
-        final QuadTree<Polygon> thisOuters = QuadTree.forLocated(this.outers());
+        final RTree<Polygon> thisOuters = RTree.forLocated(this.outers());
         // each outer of that must fit within one of this' outer/inner groups
         for (final Polygon thatOuter : that.outers())
         {
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
@@ -9,7 +9,6 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.openstreetmap.atlas.geography.clipping.Clip;
 import org.openstreetmap.atlas.geography.clipping.Clip.ClipType;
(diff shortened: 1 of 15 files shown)


jklamer:
Also PR number #100 !!! celebrate.gif

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -238,6 +240,55 @@ public boolean fullyGeometricallyEncloses(final PolyLine polyLine)
...
+            if (!cleared)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
+    public boolean fullyGeometricallyEncloses(final Segment segment)
```
MikeGost: Something we discussed offline: could we potentially leverage this method when looking at `fullyGeometricallyEncloses(final PolyLine polyline)`? That might need more testing.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -238,6 +240,55 @@ public boolean fullyGeometricallyEncloses(final PolyLine polyLine)
...
+            }
+            if (!cleared)
+            {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    @Override
```
MikeGost: To be consistent, can we add a javadoc here as well.

src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
```diff
@@ -99,6 +100,37 @@ public boolean polygonEntityIntersecting(final Polygon polygon, final AtlasEntit
                 return false;
             }
         }
+
+        // note this is the only one now used by the filter
```
MikeGost: What does this comment mean?

jklamer: The geometricSrufaceEntityIntersecting is the only function that is being used now. Which relates to my overall PR question of should I get rid of the Intersection Policy Interface and replace it with a customizable BiPredicate of AtlasEntity and GeometricSurface that I cast to serializable?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Area.java
```diff
@@ -62,9 +63,9 @@ public ItemType getType()
     }
 
     @Override
-    public boolean intersects(final Polygon polygon)
+    public boolean intersects(final GeometricSurface geometricSurface)
```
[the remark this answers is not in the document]
jklamer: sure, we already have a class called surface so I just avoided it first go through.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
jklamer: Did some performance testing and this is not as performant so I am going to remove and remove from the interface.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -30,14 +31,13 @@
  *
  * @author matthieun
  */
-public class MultiPolygon implements Iterable<Polygon>, Located, Serializable
+public class MultiPolygon implements Iterable<Polygon>, GeometricSurface, Located, Serializable
```
matthieun: You can remove the `Located` here!

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -219,23 +220,50 @@ public boolean fullyGeometricallyEncloses(final Location location)
...
+     * {@link MultiPolygon}
+     * 
+     * @param that
+     *            the provided {@link MultiPolygon} to test
+     * @return true if the conditions are met, false otherwise
+     */
+    @Override
+    public boolean fullyGeometricallyEncloses(final MultiPolygon that)
+    {
+        final QuadTree<Polygon> thisOuters = QuadTree.forLocated(this.outers());
```
matthieun: `QuadTree` is usually a good choice for Point objects. For anything with a shape, I would recommend using an `RTree` instead.

matthieun: Did you test the performance on this one out of curiosity?

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -190,6 +191,12 @@ public boolean fullyGeometricallyEncloses(final Location location)
         }
     }
 
+    @Override
+    public boolean fullyGeometricallyEncloses(final MultiPolygon multiPolygon)
+    {
+        return multiPolygon.outers().stream().allMatch(this::fullyGeometricallyEncloses);
```
matthieun: What if this polygon entirely falls within an inner polygon of the multipolygon passed in? I believe this is not a covered use case here? If not, mind adding a test for that specific corner case?

src/main/java/org/openstreetmap/atlas/geography/index/QuadTree.java
```diff
@@ -33,6 +36,21 @@
     private final Quadtree tree = new Quadtree();
     private Rectangle bound;
 
+    public static <K extends Located> QuadTree<K> forLocated(final Collection<K> locatedCollection)
```
matthieun: If you use only `forEach` how about making it more generic and using `Iterable` instead of `Collection`?

src/main/java/org/openstreetmap/atlas/geography/index/QuadTree.java
```diff
@@ -33,6 +36,21 @@
...
+    {
+        final QuadTree<K> toReturn = new QuadTree<>();
+        locatedCollection.forEach(located -> toReturn.add(located.bounds(), located));
+        return toReturn;
+    }
+
+    public static <K> QuadTree<K> forCollection(final Collection<K> collection,
```
matthieun: Same comment here.

src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
MikeGost: Good question. Personally I think having a BiPredicate of something like that can get fairly messy and abstracting it away as a policy makes more sense. Open to other ideas though :)

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
jklamer: For a polygon to geometrically enclose a multi-polygon, it must geometrically enclose all the outer polygons of the multi polygon. If the polygon lies within the inner of the passed in multi-polygon then, the polygon does not geometrically enclose the passed in multi-polygon.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
jklamer: I didn't take any time statistics. Are there two different options you would like me to compare the performance of?

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
jklamer: I added a test case to display the interaction I believe to be desirable based on what I understand of the method.

PR #101, opened 2018-03-24
Updating Geojson Polygon/MultiPolygon Read/Write

jklamer:
Updating our Geojson read/write so that we're writing and reading Polygons and MultiPolygons more to [specifications](https://tools.ietf.org/html/rfc7946):
- Stop constructing our polygons/multipolygons with repeated first and last Locations
- Start printing our polygons/multipolygons with repeated first and last locations
- deprecate inaccurate functions

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
@@ -25,6 +25,8 @@ import org.openstreetmap.atlas.utilities.collections.MultiIterable;
 import org.openstreetmap.atlas.utilities.collections.StringList;
 import org.openstreetmap.atlas.utilities.maps.MultiMap;
 import org.openstreetmap.atlas.utilities.scalars.Surface;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * Multiple {@link Polygon}s some inner, some outer.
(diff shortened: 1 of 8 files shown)


src/main/java/org/openstreetmap/atlas/streaming/readers/json/converters/PolygonCoordinateConverter.java
```diff
@@ -45,6 +46,13 @@ public JsonArray convert(final Iterable<Location> object)
                     result.add(this.coordinateConverter.revert().convert(array2));
                 });
             });
+            if (result.isEmpty() || !result.get(0).equals(result.get(result.size() - 1)))
```
jwpgage: Does this mean that a polygon with one point is a valid polygon?

jklamer: I should clarify the error. These are the conditions by which a geojson polygon object won't be accurately geometrically translated (ignoring the inner ring case until a more permanent solution is implemented). To be valid Geojson you technically also have to have 4 points (Repeated first et end) and follow the right hand rule (although interpreters can't assume this). However, anything that follows these rules can be accurately read into an Atlas Polygon. But if it's one point, it'll just get removed and then the Polygon constructor will throw an exception.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -113,6 +120,18 @@ public GeoJsonObject asGeoJsonFeatureCollection()
...
 
+    /**
+     * @return Optional of {@link Polygon} representation if possible
+     */
+    public Optional<Polygon> asSimplePolygon()
+    {
+        if (this.isSimplePolygon())
+        {
+            return outers().stream().findFirst();
+        }
```
matthieun: Maybe add a WARN log here to say "trying to save a complex multipolygon as a simple polygon"

jklamer: What context do you want the logger warning?

PR #103, opened 2018-03-29
Ingest OSM Nodes with no tags, not a relation member or part of an OSM Way

MikeGost:
The main change here is to ingest stand-alone OSM Nodes that have no tagging, that aren't part of an OSM Relation and are not a part of an OSM Way. Here are a few examples of such nodes in Antarctica: (thanks for sharing these @jwpgage!) https://www.openstreetmap.org/node/2343553527 and https://www.openstreetmap.org/node/4353689487. In the current code, these nodes are not in the final Atlas. After these changes, if using the Raw Atlas flow, these will become Atlas `Points`. There is an integration test verifying this behavior. This brings an added benefit that Atlas is more accurate with what the OSM data actually has. We can then leverage the Atlas Checks project to find any such Nodes, flag and eventually remove them.

The other change are minor comment refactoring, variable rename for clarity and removing unused test code.


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -639,15 +639,26 @@ private boolean isAtlasNode(final WaySectionChangeSet changeSet, final Point poi
     private boolean isAtlasPoint(final WaySectionChangeSet changeSet, final Point point)
     {
         final boolean hasExplicitOsmTags = pointHasExplicitOsmTags(point);
+        final boolean isRelationMember = !point.relations().isEmpty();
```
jwpgage: Great idea, this makes the logic very easy to follow!

jwpgage:
Thanks @MikeGost! LGTM, nice change and test.

PR #104, opened 2018-04-02
AtlasStatistics Unterminated CSV quoted fields Fix

jwpgage:
This change will escape double quote characters (") present within `sub_type` strings. This fixes an issue with unterminated CSV quoted fields.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/statistics/AtlasStatistics.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/statistics/AtlasStatistics.java
@@ -76,7 +76,7 @@ public class AtlasStatistics implements Iterable<AtlasStatistics.StatisticKey>, 
         @Override
         public String toString()
         {
-            return this.tag + "," + this.type + "," + this.subType;
+            return this.tag + "," + this.type + "," + this.subType.replace("\"", "\"\"");
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/statistics/AtlasStatistics.java
```diff
@@ -76,7 +76,7 @@ public int hashCode()
         @Override
         public String toString()
         {
-            return this.tag + "," + this.type + "," + this.subType;
+            return this.tag + "," + this.type + "," + this.subType.replace("\"", "\"\"");
```
AlexHsieh22: Could [StringEscapeUtils](https://commons.apache.org/proper/commons-text/javadocs/api-release/org/apache/commons/text/StringEscapeUtils.html) be helpful here to cover more cases?

jwpgage: Thanks for the suggestion! Double quotes are the only thing that needs to be escaped in CSV, but this will wrap strings with a newline or comma in double quotes as well. I'll make the change.

jwpgage:
@MikeGost A test case is a good idea, I'll add one.

PR #105, opened 2018-04-03
Serialization of PackedAtlas features in protocol buffer format

lucaspcram:
Getting this PR for the PackedAtlas serialization started. So far the only proto serializable feature is the LongArray.

List of complex types that need to be proto-ized:
AtlasMetaData
IntegerDictionary
~LongArray~
LongToLongMap
LongArrayOfArrays
PackedTagStore
LongToLongMultiMap
PolyLineArray
PolygonArray
ByteArrayOfArrays
IntegerArrayOfArrays

- Proto files are currently being defined as glorified byte buffers, this is bad because:
	+ you lose protobuf variable length encodings, one of protobuf's big draws
	+ you are now forced to write type <-> byte[] conversion code, which is essentially just doing the work that serialization libraries (like protobuf) are supposed to be doing for you
	+ you lose the utility of having nice, well defined types eg. int64, string, float32
	+ programmers who would like to deserialize Atlas files into objects in other languages now have to port all the type <-> byte[] conversion code in addition to the core atlas functionality (which is a nontrivial amount of code)
	+ at this point protobuf provides no advantages over just writing raw bytes out to a binary file in a custom format, since that's all the 'repeated bytes' type is anyway

UPDATE:
We are going to try moving forward with the autoboxed types, and just eat the performance hit.

EDIT:
Since we are moving forward with a different design, I am going to close this PR and rebranch off of dev with the updated code.


src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPrimitiveLongArrayAdapter.java
```diff
@@ -0,0 +1,53 @@
...
+ * ProtoPrimitiveLongArray. Also provides a method to obtain a LongArray back from a
+ * ProtoPrimitiveLongArray.
+ *
+ * @author lcram
+ */
+public class ProtoPrimitiveLongArrayAdapter
+{
+    /*
+     * TODO perhaps this class would be better implemented as a TwoWayConverter and not an Adapter?
+     */
```
lucaspcram: This class initially was an Adapter class, but I ended up changing it to more of a Converter model. This might need a change.

src/main/java/org/openstreetmap/atlas/utilities/conversion/LongArrayToListOfByteArrayConverter.java
```diff
@@ -0,0 +1,112 @@
...
+                bytesWritten = 0;
+                resultListOfByteArrays.add(temporaryBuffer);
+            }
+        }
+        // there are still some unflushed data in the temporary buffer, trim the temporary buffer
+        // and flush to the list
+        if (bytesWritten > 0)
+        {
+            final byte[] temporaryTrimmedBuffer = new byte[bytesWritten];
+            System.arraycopy(temporaryBuffer, 0, temporaryTrimmedBuffer, 0, bytesWritten);
```
lucaspcram: This will throw an error in the case there longArray is null, since in that case the temporary buffer will never be allocated. Need to fix this issue.

PR #106, opened 2018-04-05
Raw Atlas Enhancements

MikeGost:
Couple of notable changes here:

1. Raw Atlas creation constrained by bounding box. Instead of building everything in a given PBF, there is now the option to pass in a `MultiPolygon` to limit what is brought in from the given PBF.
2. The above change caused resulted in the `OsmPbfCounter` having to figure out which features to bring in or leave out. The logic for that is pretty straight-forward and is explained in a comment at the top of the class javadoc.
3. Update the `OsmPbfReder` to be aware of the the changes coming from the `OsmPbfCounter` and include only those features in the raw Atlas that the counter deemed suitable to bring in. As a bonus side-affect - the ingestion logic is now flexible enough to work with corrupt PBFs - it will simply not bring in features that have missing data. This is reflected in one of the test case updates.
4. Fixed an issue with slicing logic. We are filtering features out if they are outside the desired bound, where a bound is decided by a set of countries we're interested in. This was working great for any `Line` features that were being sliced, but because the raw Atlas contained Points, the Points for any Lines left out due to being outside the target country were not being filtered out. For example - if we removed a Line because it was outside of the given country set - we now clean up all the Points that make up this line as well.
5. Added tests for filtering PBFs by bounding box and removing Points by target countries
6. Updated constructors and tests accoringly
7. Fixed logging based on local testing 
8. Added more commenting and general explanation of logic


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
```diff
@@ -106,12 +193,243 @@ else if (rawEntity instanceof Bound)
...
+                    }
+                });
+                extensionCounter++;
+            }
+        }
+    }
+
+    private boolean isHighwayOrFerry(final Way way)
+    {
+        final Taggable taggableWay = new Taggable()
```
matthieun: Isn't there a converter for osmosis ways to taggable? I am not sure. Might be dreaming

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfReader.java
```diff
@@ -252,7 +278,7 @@ private RelationBean constructRelationBean(final Relation relation)
                 }
                 else
                 {
-                    logger.warn(MISSING_MEMBER_MESSAGE, relation.getId(), EntityType.Node,
+                    logger.debug(MISSING_MEMBER_MESSAGE, relation.getId(), EntityType.Node,
```
matthieun: `trace` even, no? This happens a lot because of the way osmisis cuts pbfs based on a shard boundary...

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -201,6 +201,10 @@ public RawAtlasSlicer(final Set<IsoCountry> countries,
     protected Map<String, String> createPointTags(final Location location,
             final boolean fromRawAtlas)
     {
+        if (location.equals(Location.forWkt("POINT (-63.0473703 18.001941)")))
+        {
+            System.out.println("Trying to get tags for our target point");
```
matthieun: This looks like debug code :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
MikeGost: There isn't! But doing some digging, I can leverage TagMap which handles translation between a Collection<Tag> to a Taggable object. Let me know what you think - simplifies the code quite a bit!

jklamer:
May not be useful but is there any added utility in using GeometricSurface instead of MultiPolygon for the  bounding box?

MikeGost:
@jklamer - Good point! I think it makes sense to be flexible and use either Polygon or MulitPolygon as the bounding box, so sticking with the interface instead of trying it to a specific implementation makes sense.

PR #107, opened 2018-04-05
Serialize the PackedAtlas fields using protocol buffer format

lucaspcram:
EDIT: I think we are ready for merge.

Moving forward with implementation using builtin protocol buffer types and the consequent autoboxing this incurs. Note that these current proto implementations do not support the "very large" functionality of the PackedAtlas field types. This is probably OK for our use-case, as we are unlikely to have arrays with size greater than Integer.MAX_VALUE (~2 billion).

I added some new interfaces (basically a GoF adapter) in order to minimize changes to the PackedAtlasSerializer. Using this method, we can add new fields to PackedAtlas at-will without changing the serializer code - provided those fields obey the interface contracts I am laying out in this PR.

Using the interface is easy and intuitive! Here's how.
```proto
message ProtoSomePackedType {
    // define fields here
}
```
```java

public class SomePackedType implements ProtoSerializable
{
    @Override
    public ProtoAdapter getProtoAdapter()
    {
        return new SomePackedTypeAdapter();
    }
    // rest of class goes here
    .
    .
    .
}

public class SomePackedTypeAdapter implements ProtoAdapter
{
    @Override
    public ProtoSerializable deserialize(byte[] byteArray)
    {
        // type specific code that uses the generated proto classes
    }

    @Override
    public byte[] serialize(ProtoSerializable serializable)
    {
        // type specific code that uses the generated proto classes
    }
}
```

This PR will grow as more types are converted. I will update this list below as I go along.

Types that need adapters
--------------------------
~AtlasMetaData~
~IntegerDictionary~
~LongArray~
~LongToLongMap~
~LongArrayOfArrays~
~PackedTagStore~
~LongToLongMultiMap~
~PolyLineArray~
~PolygonArray~
~ByteArrayOfArrays~
~IntegerArrayOfArrays~


src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoAdapter.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author lcram
+ */
+public interface ProtoAdapter
+{
+    /**
+     * @param byteStream
+     *            the raw byte representation of the ProtoSerializable in protocol buffer format
+     * @return The object represented by the byte stream.
+     */
+    ProtoSerializable deserialize(byte[] byteStream);
```
matthieun: How about calling this `byteArray` instead?

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoLongArrayAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+        {
+            protoLongArray = ProtoLongArray.parseFrom(byteStream);
+        }
+        catch (final InvalidProtocolBufferException excep)
+        {
+            throw new CoreException("Error encountered while parsing protobuf bytestream", excep);
+        }
+
+        final LongArray longArray = new LongArray(protoLongArray.getElementsCount());
+        for (int i = 0; i < protoLongArray.getElementsCount(); i++)
```
matthieun: Could a `for (Xxx xxx : yyy)` work here?

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoLongArrayAdapter.java
```diff
@@ -0,0 +1,59 @@
...
+        return longArray;
+    }
+
+    @Override
+    public byte[] serialize(final ProtoSerializable serializable)
+    {
+        if (!(serializable instanceof LongArray))
+        {
+            throw new CoreException(
+                    "Invalid ProtoSerializable type was provided to " + this.getClass().getName()
```
matthieun: Here you can use `new CoreException ("blah blah {} and {}", one, two);` and it will become `blah blah one and two` like when using a `logger`.

src/main/java/org/openstreetmap/atlas/utilities/arrays/LongArray.java
```diff
@@ -53,6 +58,12 @@ public LongArray(final long maximumSize, final int memoryBlockSize, final int su
         super(maximumSize, memoryBlockSize, subArraySize);
     }
 
+    @Override
+    public ProtoAdapter getProtoAdapter()
+    {
+        return this.protoLongArrayAdapter;
```
matthieun: How about returning a new one every time? That would let you not have any transient variable. Less risky if you try to de-serialize a Java based Atlas and then re-save it... You would get a NPE here, no?

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoLongArrayAdapter.java
lucaspcram: Unfortunately the generated proto code does not implement Iterable.

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/AtlasSize.java
```diff
@@ -112,6 +114,46 @@ public AtlasSize(final long edgeNumber, final long nodeNumber, final long areaNu
...
+            }
+            if (this.getPointNumber() != that.getPointNumber())
+            {
+                return false;
+            }
+            if (this.getRelationNumber() != that.getRelationNumber())
+            {
+                return false;
+            }
+
```
snb3300: Nit: Extra Line

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedTagStore.java
```diff
@@ -115,6 +135,35 @@ public boolean containsKey(final long index, final String key)
...
+            final PackedTagStore that = (PackedTagStore) other;
+            if (!this.keys.equals(that.keys))
+            {
+                return false;
+            }
+            if (!this.values.equals(that.values))
+            {
+                return false;
+            }
+            // We use reference equality for the dictionary, since PackedTagStores do not actually
```
snb3300: Could you elaborate more on when do you think the references should be equal? Maybe an example in the comment might make it better to understand

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedTagStore.java
```diff
@@ -142,6 +191,24 @@ public String get(final long index, final String key)
...
+        return new ProtoPackedTagStoreAdapter();
+    }
+
+    @Override
+    public int hashCode()
+    {
+        final int initialPrime = 31;
+        final int hashSeed = 37;
+
+        int hash = initialPrime;
```
snb3300: instead of assigning ```initialPrime``` to ```hash``` how about

```
final int initialPrime = 31;
final int hashSeed = 37;

int hash = hashSeed * initialPrime + this.keys.hashCode();
hash = hashSeed * hash + this.values.hashCode();

return hash;
```

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoAtlasMetaDataAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+        final ProtoAtlasMetaData.Builder builder = ProtoAtlasMetaData.newBuilder();
+        builder.setEdgeNumber(atlasMetaData.getSize().getEdgeNumber());
+        builder.setNodeNumber(atlasMetaData.getSize().getNodeNumber());
+        builder.setAreaNumber(atlasMetaData.getSize().getAreaNumber());
+        builder.setLineNumber(atlasMetaData.getSize().getLineNumber());
+        builder.setPointNumber(atlasMetaData.getSize().getPointNumber());
+        builder.setRelationNumber(atlasMetaData.getSize().getRelationNumber());
+        builder.setOriginal(atlasMetaData.isOriginal());
+
+        final Optional<String> codeVersionOptional = atlasMetaData.getCodeVersion();
```
snb3300: Instead of extracting in a local variable you can also use a lambda

```
atlasMetaData.getCodeVersion().ifPresent(value -> {
  builder.setCodeVersion(value);
});
```

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoAtlasMetaDataAdapter.java
```diff
@@ -0,0 +1,108 @@
...
+        if (shardNameOptional.isPresent())
+        {
+            builder.setShardName(shardNameOptional.get());
+        }
+
+        final Map<String, String> tags = atlasMetaData.getTags();
+        builder.addAllTags(ProtoAtlasMetaDataAdapter.PROTOTAG_LIST_CONVERTER.backwardConvert(tags));
+
+        final ProtoAtlasMetaData protoMetaData = builder.build();
+        return protoMetaData.toByteArray();
```
snb3300: how about

```
return builder.build().toByteArray();
```

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoIntegerStringDictionaryAdapter.java
```diff
@@ -0,0 +1,81 @@
...
+        }
+        catch (final ClassCastException exception)
+        {
+            throw new CoreException(
+                    "This adapter is incompatible with its owner's ({}) current type parametrization. Must be java.lang.String",
+                    serializable.getClass().getName(), exception);
+        }
+
+        final ProtoIntegerStringDictionary protoDictionary = dictionaryBuilder.build();
+        return protoDictionary.toByteArray();
```
snb3300: How about

```
return dictionaryBuilder.build().toByteArray();
```
This way you don't need to instantiate a local variable before returning.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedTagStore.java
lucaspcram: Good idea. A lot of the equals/hashCode implementations are still in flux.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedTagStore.java
lucaspcram: This makes a lot more sense haha.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoAtlasMetaDataAdapter.java
lucaspcram: Good tip. I am new to using Optional. Thanks!

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoAtlasMetaDataAdapter.java
lucaspcram: I had that intermediate variable there just to make it very clear to the reader what we are returning.

I'm actually going to build on your suggestion (no pun intended) by refactoring the name of 'builder' to be a little more descriptive, something like 'protoMetaDataBuilder'.

Best of both worlds that way!

src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasMetaData.java
```diff
@@ -105,6 +159,14 @@ public AtlasSize getSize()
         return new HashMap<>(this.tags);
     }
 
+    @Override
+    public int hashCode()
+    {
+        final int sizeHash = this.getSize().hashCode();
+        return Objects.hash(Integer.valueOf(sizeHash), Boolean.valueOf(this.original),
```
MikeGost: Just checking - data version field was left out intentionally here?

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/AtlasSize.java
```diff
@@ -17,11 +18,85 @@
...
+            this.edgeEstimate = DEFAULT_ESTIMATE;
+            this.nodeEstimate = DEFAULT_ESTIMATE;
+            this.areaEstimate = DEFAULT_ESTIMATE;
+            this.lineEstimate = DEFAULT_ESTIMATE;
+            this.pointEstimate = DEFAULT_ESTIMATE;
+            this.relationEstimate = DEFAULT_ESTIMATE;
+        }
+
+        /**
+         * Builds an AtlasSize. By default it uses a small value for the size estimates.
```
MikeGost: Nit:  Consider adding a link to `DEFAULT_VALUE` constant in the javadoc by using `@link(#[your-variable])` notation. This way if it ever changes, it will get updated automatically.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -54,6 +54,7 @@
  * Atlas that packs data in large arrays
  *
  * @author matthieun
+ * @author lcram
```
MikeGost: Nit: Not sure this warrants an author update :P

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializer.java
```diff
@@ -349,6 +364,27 @@ private void load(final String name)
         }
     }
 
+    private Resource makeProtoResource(final ProtoSerializable field, final String name)
+    {
+        // We automatically get the correct adapter for whatever type 'field' happens to be
+        final ProtoAdapter adapter = field.getProtoAdapter();
+        // The adapter handles all the actual serialization using the protobuf classes. Easy!
```
MikeGost: Nice and clean!

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedTagStore.java
```diff
@@ -48,7 +69,7 @@ protected PackedTagStore(final long maximumSize, final int memoryBlockSize,
      */
     public void add(final long index, final String key, final String value)
     {
-        if (index > this.index)
+        if (index > size())
         {
             throw new CoreException(
```
MikeGost: Nit: not part of your changes, but consider updating the `CoreException` to follow insertion format:
`throw new CoreException("Cannot add. Invalid index {} is bigger than the size {}", index, size());`

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoIntegerArrayOfArraysAdapter.java
```diff
@@ -0,0 +1,64 @@
...
+        {
+            throw new CoreException(
+                    "Invalid ProtoSerializable type was provided to {}: cannot serialize {}",
+                    this.getClass().getName(), serializable.getClass().getName());
+        }
+        final IntegerArrayOfArrays integerArrayOfArrays = (IntegerArrayOfArrays) serializable;
+
+        if (integerArrayOfArrays.size() > Integer.MAX_VALUE)
+        {
+            // TODO see note about this check in ProtoLongArrayOfArraysAdapter
```
MikeGost: Maybe add this constraint in a single place (possibly a README on this) and remove the TODOs instances.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoIntegerStringDictionaryAdapter.java
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Implements the {@link ProtoAdapter} interface to connect {@link IntegerDictionary} and
+ * {@link ProtoIntegerStringDictionary}. This adapter will fail when attempting to serialize an
+ * {@link IntegerDictionary} that is not type parameterized using {@link String}.
+ *
+ * @author lcram
+ */
+public class ProtoIntegerStringDictionaryAdapter implements ProtoAdapter
+{
+    // NOTE this method will ALWAYS give back an IntegerDictionary<String>, no matter what type
```
MikeGost: Nit: This may be important enough to add to the javadoc so anyone using this class can see it.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoLongArrayAdapter.java
```diff
@@ -0,0 +1,70 @@
...
+        }
+        catch (final InvalidProtocolBufferException exception)
+        {
+            throw new CoreException("Error encountered while parsing protobuf bytestream",
+                    exception);
+        }
+
+        final int elementCount = protoLongArray.getElementsCount();
+        final LongArray longArray = new LongArray(elementCount, elementCount, elementCount);
+        for (int i = 0; i < protoLongArray.getElementsCount(); i++)
```
MikeGost: Nit: i --> index to be consistent with other classes

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+
+import com.google.protobuf.InvalidProtocolBufferException;
+
+/**
+ * Implements the {@link ProtoAdapter} interface to connect {@link PackedTagStore} and
+ * {@link ProtoPackedTagStore}.
+ *
+ * @author lcram
+ */
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+public class ProtoPackedTagStoreAdapter implements ProtoAdapter
+{
+    /*
+     * PackedTagStore does not provide an interface for setting its arrays directly. This class's
```
MikeGost: Nice commenting!

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+        Field valueArrayField = null;
+        Field indexField = null;
+
+        try
+        {
+            keyArrayField = store.getClass().getDeclaredField(PackedTagStore.FIELD_KEYS);
+            keyArrayField.setAccessible(true);
+            keyArrayField.set(store, keyArray);
+        }
+        catch (NoSuchFieldException | SecurityException | IllegalArgumentException
```
MikeGost: All the exceptions get handled the same way - there is no difference what kind of Exception it is, we can't recover from it. Might be a good spot to just catch `Exception`.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
```diff
@@ -0,0 +1,166 @@
...
+                    store.getClass().getName(), exception);
+        }
+
+        try
+        {
+            valueArrayField = store.getClass().getDeclaredField(PackedTagStore.FIELD_VALUES);
+            valueArrayField.setAccessible(true);
+            valueArrayField.set(store, valueArray);
+        }
+        catch (NoSuchFieldException | SecurityException | IllegalArgumentException
```
MikeGost: Same comment here and in all below occurrences.

src/main/java/org/openstreetmap/atlas/utilities/arrays/LongArray.java
```diff
@@ -1,11 +1,16 @@
...
 
+import org.openstreetmap.atlas.proto.ProtoSerializable;
+import org.openstreetmap.atlas.proto.adapters.ProtoAdapter;
+import org.openstreetmap.atlas.proto.adapters.ProtoLongArrayAdapter;
+
 /**
  * {@link LargeArray} for type {@link Long}
  *
  * @author matthieun
+ * @author lcram
```
MikeGost: Nit: Not sure some of these changes warrant an author update.

src/main/java/org/openstreetmap/atlas/utilities/maps/LongToLongMultiMap.java
```diff
@@ -74,6 +91,86 @@ public void add(final Long key, final long value)
         this.put(key, result);
     }
 
+    @Override
+    public boolean equals(final Object other)
+    {
+        if (other instanceof LargeMap)
```
MikeGost: Should this be `instanceof LongToLongMultiMap`?

src/main/java/org/openstreetmap/atlas/utilities/arrays/LongArray.java
lucaspcram: I thought a bit about this. Where should the line be? My internal thought was if I added at least one new method, or significantly changed a method, I added an author tag - if anything to document from whom the code in the class is coming from. 

Looking at LongArray, I think this is probably a case where I was too overzealous with the tag.  But I think for the classes where I started adding new interfaces, including equals() and hashCode() implementations, that tag is a good idea for future users of the class.

src/main/java/org/openstreetmap/atlas/utilities/maps/LongToLongMultiMap.java
lucaspcram: Yeah... copy paste strikes again.

src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasMetaData.java
lucaspcram: No it should be there haha. Since I check against it in equals(), I should be using it in the hash computation. Good catch.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoIntegerArrayOfArraysAdapter.java
lucaspcram: After talking with Matthieu a bit about the size constraints, I think it might be best to just remove these checks. The arrays we pass in will never be that large anyway, and I was not consistent about enforcing that check across all the different adapters. If it becomes a problem, in the future I can take some time to implement that check correctly everywhere.

src/main/java/org/openstreetmap/atlas/utilities/arrays/LongArray.java
MikeGost: This is a tricky area - everyone has different opinions on what's the right convention. This is only my opinion, so take that for what it's worth :) - I try to answer both of these questions with a yes before adding an author tag:
1. Did I make a change that can have downstream/user impact and/or add significant functionality on top of the existing codebase?
2. Do I feel comfortable explaining this class and its functionality and/or debugging/being responsible for any bugs that are a result of my changes?

In your case case, most of the author tags are spot on! I'd keep all of them as is, including the ones with equals/hashcode implementations. Not a fan of the fine-grained per-method author tags as that just brings more chaos to the code. I just nit-picked two classes that stood out to me as a bit overzealous (PackedAtlas and LongArray). Again, feel free to ignore :)

lucaspcram: Got it. I'll revert the per-method tags in a future commit. All this makes sense. Yeah I do agree that PackedAtlas and LongArray are overzealous haha.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -61,6 +61,11 @@
     private static final Logger logger = LoggerFactory.getLogger(PackedAtlas.class);
 
     // Keep track of the field names for reflection code in the Serializer.
+    protected static final String FIELD_PREFIX = "FIELD_";
+    protected static final String FIELD_BOUNDS = "bounds";
+    protected static final String FIELD_LOGGER = "logger";
+    protected static final String FIELD_SERIAL_VERSION_UID = "serialVersionUID";
+
```
lucaspcram: pre-emptive nit: extra line
:)

lucaspcram:
There is a weird issue where `FullProtoSuiteTest.testEmptyTags()` is passing locally, but failing on Travis. Definitely need to look into why this is.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -57,11 +57,31 @@
...
+        JAVA,
+        PROTOBUF,
+    }
+
     private static final long serialVersionUID = -7582554057580336684L;
     private static final Logger logger = LoggerFactory.getLogger(PackedAtlas.class);
 
     // Keep track of the field names for reflection code in the Serializer.
+    protected static final String FIELD_PREFIX = "FIELD_";
+    protected static final String FIELD_BOUNDS = "bounds";
```
matthieun: Why would those be needed if they are not serialized?

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -106,11 +126,16 @@
     // Serializer.
     private transient PackedAtlasSerializer serializer;
 
+    // Serialization formats for saving/loading this PackedAtlas
+    private transient AtlasSerializationFormat saveSerializationFormat = AtlasSerializationFormat.JAVA;
+    private transient AtlasSerializationFormat loadSerializationFormat = AtlasSerializationFormat.JAVA;
```
matthieun: Isn't loadSerializationFormat always dependent on the file that it tries to load from? To my mind only the `saveSerializationFormat` would be needed, no?

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializer.java
```diff
@@ -260,6 +290,69 @@ private Object deserializeResource(final Resource resource)
...
+            protoHandle = (ProtoSerializable) handle;
+        }
+        catch (final ClassCastException exception)
+        {
+            throw new CoreException("{} is not ProtoSerializable", fieldClass.getName(), exception);
+        }
+
+        final ProtoAdapter adapter = protoHandle.getProtoAdapter();
+        final ProtoSerializable deserializedMember = adapter
+                .deserialize(resource.readBytesAndClose());
```
matthieun: This would return a `byte[]` correct? In case the field is big, would it be possible to stream the resource, one byte buffer at a time using an InputStream instead? That would save memory. Only if it is easy though, as the tradeoff might not be worth it..

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializer.java
```diff
@@ -85,24 +94,44 @@ protected static PackedAtlas load(final Resource resource)
...
+     * Try loading the meta data to make sure the data format is appropriate. Keep trying formats
+     * until we find the right one
+     */
+    private static void determineAtlasLoadFormat(final PackedAtlas atlas)
+    {
+        final AtlasSerializationFormat[] possibleFormats = AtlasSerializationFormat.values();
+        for (final AtlasSerializationFormat candidateFormat : possibleFormats)
+        {
+            logger.info("Trying load format {}", candidateFormat);
+            atlas.setLoadSerializationFormat(candidateFormat);
```
matthieun: I would keep that load format local to the `PackedAtlasSerializer` here as it is very dependent on the resource being read.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializer.java
lucaspcram: I will look into this. Could definitely be worth it and shouldn't be too difficult.

hallahan:
This is awesome! What's left? When will this get merged?

lucaspcram:
@hallahan Still working out a couple kinks surrounding some changes in `PackedAtlas`. I'm hoping to get this merged very soon, and then potential iterate on it if necessary. The protobuf scheme is going to start out as opt-in, but will eventually be the default moving forward.

PR #108, opened 2018-04-10
MultiPolygon empty method and empty Clipping test

MikeGost:
I ran into an interesting issue: when creating an `AND` `Clip` between two disjoint `MultiPolygon`s, the return value is a `MULTIPOLYGON EMPTY` String. This can happen when converting from an empty JTS `MultiPolygon` to an Atlas `MultiPolygon` or by initializing the MultiPolygon constructor with an empty `MultiMap`. This PR adds a MultiPolygon `isEmpty()` method and safeguards the bounds call before calculating. Added are also two tests - one to test null bounds on an empty MultiPolygon and the other to test a disjoint MultiPolygon clipping.


src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -238,29 +238,10 @@ public boolean fullyGeometricallyEncloses(final Location location)
         return false;
     }
 
-    /**
-     * @param polyLine
-     *            A {@link PolyLine} item
-     * @return True if the {@link MultiPolygon} contains the provided {@link PolyLine}.
-     */
-    @Override
-    public boolean fullyGeometricallyEncloses(final PolyLine polyLine)
```
MikeGost: Ignore the formatting here, I've double checked my Eclipse settings and ran spotlessApply before pushing the changes. The only worthwhile changes here are lines 336-343 and line 142

jwpgage:
Looks good! Thanks @MikeGost, nice tests as well.

PR #111, opened 2018-04-14
Make Country->Shard listing re-usable

matthieun:
This country to shard listing was initially in atlas-generator. Now it is presented as a utility class.

Also updated and un-skipped some unit tests related.


src/main/java/org/openstreetmap/atlas/geography/boundary/CountryShardListingProcessor.java
```diff
@@ -0,0 +1,95 @@
...
+    @Override
+    public void run()
+    {
+        try
+        {
+            while (!this.queue.isEmpty())
+            {
+                final CountryBoundary itemToProcess = this.queue.poll();
+
+                // Queue might have been emptied by another thread. That gives an null item.
```
mertemin: `an null item` -> `a null item`

PR #114, opened 2018-04-19
PackedAtlas: Do not load non serializable fields

matthieun:
In the unusual case of loading then saving an Atlas right away, some fields are not properly filtered using reflection, and cause errors.
This PR fixes that issue and adds a test to verify it.


lucaspcram:
Looks good! Apparently I was not the only one running into this bug. A welcome fix.

PR #115, opened 2018-04-19
StringList Serialization

jwpgage:
Made `StringList` serializable.


lucaspcram:
Really solid code James. Mind giving us a quick summary about some of the design decisions that went into these changes?

PR #117, opened 2018-04-21
Make ConfiguredTaggableFilter serializable

matthieun:
This is a simple change to allow Spark to pass configurations between driver and executors.


jwpgage:
Looks good! Thanks @matthieun

PR #118, opened 2018-04-22
Fixing raw Atlas sectioning and slicing corner cases

MikeGost:
Multiple changes here:

1. Fixing a `ClassCastException` when merging raw `Atlas Relation`s.
2. Making `Relation` handling more dynamic - check for an empty bean before adding a `Relation`.
3. When we filter our a sliced chunk of a way - we weren't getting rid of the shape points for the filtered slices. They were remaining as part of the Atlas when they should not have been.
4. During way-sectioning - our previous predicate was to bring in all Points in the neighboring shards. This caused some issues since we want to keep isolated Points. We now only bring in points that are part of a Line that qualifies to be an Edge. This slows down the DynamicAtlas to some degree, but allows for a more accurate result in the end. Thinking of other optimizations for this.
5. Updated `RawAtlasRelationSlicer` to avoid returning sliced members and instead just update the change set. This was a remnant from a recursive implementation attempt that never got cleaned up!
6. Added better logging around raw atlas generation, logging the shard name being processed, filtered and generated.
7. For a shard that contains no data, or is empty as a result of filtering out features, doing a null check before processing.
8. Accounting for repeated shape points in an OSM Way when way-sectioning and getting rid of single-`Node` `Edge`s. Two test cases for this behavior
9. Added more logging around way-sectioning and time spent on each step
10. Added better logging around failures when building a `PackedAtlas`. 
11. Cosmetic cleanup

Tested locally on multiple PBFs that ran into the issue.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
@@ -80,7 +80,7 @@ public class WaySectionProcessor
     // this further by only considering the edges crossing the shard boundary and their intersecting
     // edges to reduce the memory overhead on each slave.
 
-    // Combination of the line and point predicate used to
+    // Dynamic expansion filter will be a combination of points and lines
     private final Predicate<AtlasEntity> dynamicAtlasExpansionFilter = entity -> this.pointPredicate
             .test(entity) || this.linePredicate.test(entity);
 
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
@@ -144,7 +144,7 @@ public class WaySectionProcessor
     public Atlas run()
     {
         final Time time = Time.now();
-        logger.info("Started Way-Sectioning for Shard {}", this.loadedShards.get(0));
+        logger.info("Started Way-Sectioning for Shard {}", getShardOrAtlasName());
 
         // Create a changeset to keep track of any intermediate state transitions (points becomes
         // nodes, lines becoming areas, etc.)
@@ -161,7 +161,7 @@ public class WaySectionProcessor
         sectionEdges(changeSet);
 
         final Atlas atlas = buildSectionedAtlas(changeSet);
-        logger.info("Finished Way-Sectioning for Shard {} in {}", this.loadedShards.get(0),
+        logger.info("Finished Way-Sectioning for Shard {} in {}", getShardOrAtlasName(),
                 time.untilNow());
 
         return cutSubAtlasForOriginalShard(atlas);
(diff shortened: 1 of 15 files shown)


MikeGost:
@matthieun - Most of these are issues that are hard to test for specifically, but do occur when testing at scale. I'll try to add a unit test where applicable. Also, adding a DNM label for now as I'm working on adding another issue.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -188,6 +201,9 @@ private void addPointToNodeList(final Location location,
     private Atlas buildExpandedAtlas(final Shard initialShard, final Sharding sharding,
             final Function<Shard, Optional<Atlas>> rawAtlasFetcher)
     {
+        final Time time = Time.now();
+        logger.info("Started Dynamic Atlas Construction for Shard {}", initialShard.getName());
```
matthieun: Maybe here add the stage name too!
```
"Started Dynamic Atlas Construction for way-sectioning Shard {}"
```

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -220,6 +236,9 @@ private Atlas buildExpandedAtlas(final Shard initialShard, final Sharding shardi
 
         final DynamicAtlas atlas = new DynamicAtlas(policy);
         atlas.preemptiveLoad();
+
+        logger.info("Finished Dynamic Atlas Construction for {} in {}", initialShard.getName(),
```
matthieun: same here

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -462,8 +494,13 @@ private Atlas cutSubAtlasForOriginalShard(final Atlas atlas)
      */
     private void distinguishPointsFromShapePoints(final WaySectionChangeSet changeSet)
     {
+        final Time time = Time.now();
+        logger.info("Started Shape Point Detection...");
```
matthieun: Mention which shard here? Is it possible given the only parameter passed in is the changeset?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -479,6 +516,9 @@ private void distinguishPointsFromShapePoints(final WaySectionChangeSet changeSe
      */
     private void identifyEdgesNodesAndAreasFromLines(final WaySectionChangeSet changeSet)
     {
+        final Time time = Time.now();
+        logger.info("Started Atlas Feature Detection...");
```
matthieun: Same, if possible...

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good point, it is, we have all loaded shards!

PR #119, opened 2018-04-24
Fix for edge case of LargeArray.trim() causing ArrayIndexOutOfBoundsException 

lucaspcram:
The previous version of `LargeArray`'s `trim()` method had a bug that would only appear when the rightmost subarray was completely full. This bug was due to the logic that calculated the size to which the subarray would be trimmed. Checking out `LargeArray.trimIfFilledLessThan()`, you can see the code effectively does the following:

```java
final int arrayIndex = this.arrays.size() - 1;
final int indexInside = this.nextIndex % this.subArraySize;
this.arrays.set(arrayIndex, rightmost.trimmed(indexInside));
```

In the case that the rightmost subarray is full, `this.nextIndex % this.subArraySize` will of course be zero. So then the `arrays.set` would reset the rightmost subarray by calling `trimmed(0)` on it, which would wipe it out completely.

The fix is simply to exit `LargeArray.trimIfFilledLessThan()` early in the case that the rightmost subarray is full.

There is another possible issue I discovered while fixing this bug that is orthogonal to the `trim()` bug. Currently, `LargeArray` allows for `memoryBlockSize` to exceed `subArraySize`. This seems to be a problem. When `memoryBlockSize > subArraySize`, the LargeArray will over allocate each subarray by `memoryBlockSize - subArraySize`. Apparently there are many places in the code where arrays are created with `memoryBlockSize = 1024` and `subArraySize = 1`. This means that at each of these points, 1023 blocks are completely wasted by the `LargeArray`. This does not seem like something we want, but I am unsure of what to do here.
@matthieun we definitely need to take a closer look at this

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/utilities/arrays/LargeArrayTest.java
+++ b/src/test/java/org/openstreetmap/atlas/utilities/arrays/LargeArrayTest.java
@@ -26,6 +26,76 @@ public class LargeArrayTest
         new LargeArrayTest().largeSerialization();
     }
 
+    @Test
+    public void exposeTrimEdgecaseBug1()
+    {
+        final int maxSize = 30;
+        final int subArraySize = 10;
+        final int blockSize = subArraySize;
+
+        final LongArray array = new LongArray(maxSize, blockSize, subArraySize);
+
+        // Completely fill the first subarray
+        for (int i = 0; i < subArraySize; i++)
+        {
+            array.add(2L);
+        }
+
+        // Now, we trim the array. This will corrupt the rightmost* subarray if and only if it is
+        // completely full (ie. a subsequent call to add() will force a new array allocation)
+        // * Here, rightmost indicates it is the last subarray in LargeArray's list of subarrays
+        array.trim();
+
+        // this assert will throw an ArrayIndexOutOfBounds exception because the trim method resized
+        // the rightmost subarray to a size-0 array
+        try
+        {
+            Assert.assertEquals(2L, array.get(1).longValue());
+        }
+        catch (final ArrayIndexOutOfBoundsException exception)
+        {
+            exception.printStackTrace();
+            Assert.fail("Caught the bug!");
+        }
+    }
+
+    @Test
+    public void exposeTrimEdgecaseBug2()
+    {
+        final int maxSize = 30;
+        final int subArraySize = 10;
+        final int blockSize = subArraySize;
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/utilities/arrays/LargeArrayTest.java
```diff
@@ -26,6 +26,76 @@ public static void main(final String[] args)
         new LargeArrayTest().largeSerialization();
     }
 
+    @Test
+    public void exposeTrimEdgecaseBug1()
```
jwpgage: Maybe TrimEmptyRightmostSubarrayTest?

src/test/java/org/openstreetmap/atlas/utilities/arrays/LargeArrayTest.java
```diff
@@ -26,6 +26,76 @@ public static void main(final String[] args)
...
+            array.add(2L);
+        }
+
+        // Now, we trim the array. This will corrupt the rightmost* subarray if and only if it is
+        // completely full (ie. a subsequent call to add() will force a new array allocation)
+        // * Here, rightmost indicates it is the last subarray in LargeArray's list of subarrays
+        array.trim();
+
+        // this assert will throw an ArrayIndexOutOfBounds exception because the trim method resized
+        // the rightmost subarray to a size-0 array
```
jwpgage: Except this has been fixed now, correct? Might want to change the tense here

lucaspcram:
I don't believe that it does, as I have not changed any of the fields. I am fairly certain that adding a private method/changing method code will not affect serialization. I will do a couple local tests reading/writing with AtlasDelta and make sure nothing is affected.

lucaspcram:
I went ahead and ran an AtlasDelta test on one of my Atlas files, and it looks all good!

```java
@Test
public void test()
{
        final ByteArrayResource resource = new ByteArrayResource(1024 * 1024 * 24);
        final Atlas atlas = PackedAtlas
                .load(new File(...));
        atlas.save(resource);
        final Atlas readIn = PackedAtlas.load(resource);
        
        Assert.assertEquals(atlas, readIn);
        
        final SortedSet<Diff> diff = new AtlasDelta(atlas, readIn, true).generate()
                .getDifferences();
        Assert.assertTrue(diff.isEmpty());
}
```
I think this is ready for merge.

PR #120, opened 2018-04-24
Fix performance issue with fullyGeometricallyEncloses (and awtOverflows)

matthieun:
With complex `Polygon`s (many shape points), deciding to use AWT for `fullyGeometricallyEncloses` or not is in itself a performance hog. Calling `bounds()` twice each time this method is invoked is really expensive.

This change fixes the issue when `fullyGeometricallyEncloses` is called many times with the same Polygon, which happens often with the `OsmPbfProcessor`. The decision flag is simply stored in a transient `Boolean`.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
@@ -61,6 +61,7 @@ public class Polygon extends PolyLine implements GeometricSurface
     private static final int MINIMUM_N_FOR_SIDE_CALCULATION = 3;
     private Area awtArea;
     private java.awt.Polygon awtPolygon;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -436,6 +438,7 @@ public boolean overlaps(final MultiPolygon multiPolygon)
      *            The {@link PolyLine} to test
      * @return True if this {@link Polygon} intersects/overlaps the given {@link PolyLine}.
      */
+    @Override
```
jklamer: My bad. Forgot these. Thank you!

matthieun:
I might have found another related hotspot...

PR #122, opened 2018-05-01
Add FerryTag 

isabelleh:
This PR adds the Ferry Tag and some of the more common values.

https://taginfo.openstreetmap.org/keys/ferry#overview


src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
```diff
@@ -0,0 +1,47 @@
...
+
+import org.openstreetmap.atlas.tags.annotations.Tag;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+
+/**
+ * OSM Ferry Tag
+ *
+ * @author isabellehillberg
+ */
+@Tag(taginfo = "https://taginfo.openstreetmap.org/keys/ferry#values")
```
MikeGost: Nit: Can you link the OSM wiki please:

`osm = "https://wiki.openstreetmap.org/wiki/Key:ferry"`

src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private static final EnumSet<FerryTag> CAV_NAVIGABLE = EnumSet.of(MOTORWAY, TRUNK, PRIMARY,
+            SECONDARY, TERTIARY, UNCLASSIFIED, RESIDENTIAL, SERVICE);
+    private static final EnumSet<FerryTag> PEDESTRIAN_NAVIGABLE = EnumSet.of(FOOTWAY, TRACK,
+            PEDESTRIAN);
+
+    @TagKey
+    public static final String KEY = "ferry";
+
+    public static boolean isCarNavigableHighway(final FerryTag tag)
```
MikeGost: Nit: What do you think of `isCarNavigableFerry`?

src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
```diff
@@ -0,0 +1,47 @@
...
+    {
+        return CAV_NAVIGABLE.contains(tag);
+    }
+
+    public static boolean isPedestrianNavigableHighway(final FerryTag tag)
```
MikeGost: Nit: What do you think of `isPedestrianNavigableFerry`?

src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
isabelleh: The key is not documented. The route=ferry tag is the official way to tag ferries in OSM, but the ferry tag is used by the community in some cases.

src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
isabelleh: Yes! copy fail from the highway tag :)

src/main/java/org/openstreetmap/atlas/tags/FerryTag.java
MikeGost: It is - the wiki above! It's pretty brief but at least gives users a reference point. I'd add it for completeness.

isabelleh: Got it!

PR #123, opened 2018-05-08
Parallelizing Raw Atlas Way-Sectioning and Point/Line Slicing

MikeGost:
A few changes here:

1. Fully parallelizes `Point` way-sectioning logic. This reduces total runtime by ~30-35%.
2. Parallelizes `Point` and `Line` slicing logic, this adds an additional 10-15% improvement.
3. Fixes a way-sectioning edge case and adds a test for it
4. Improved logging and time tracking
5. Adds missing serial version UID for StringList

P.S. We need an `Iterables.parallelStream(...)` method - created an issue :)

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
@@ -80,7 +80,7 @@ public class WaySectionProcessor
     // this further by only considering the edges crossing the shard boundary and their intersecting
     // edges to reduce the memory overhead on each slave.
 
-    // Combination of the line and point predicate used to
+    // Dynamic expansion filter will be a combination of points and lines
     private final Predicate<AtlasEntity> dynamicAtlasExpansionFilter = entity -> this.pointPredicate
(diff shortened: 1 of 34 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -500,8 +501,8 @@ private void distinguishPointsFromShapePoints(final WaySectionChangeSet changeSe
         final Time time = Time.now();
         logger.info("Started Shape Point Detection for Shard {}", getShardOrAtlasName());
 
-        Iterables.stream(this.rawAtlas.points()).filter(point -> isAtlasPoint(changeSet, point))
-                .forEach(changeSet::recordPoint);
+        StreamSupport.stream(this.rawAtlas.points().spliterator(), true)
```
mgcuthbert: Out of genuine curiosity, how much performance gain do you get by using spliterator here?

MikeGost: ~3x increase in performance.

lucaspcram:
Looks good to me. The `spliterator` performance increases are very cool.

PR #126, opened 2018-05-14
Fixed bugs with MultiEntities that caused incorrect empty relations() returns

lucaspcram:
**READY FOR MERGE**

In the specific case that a `Node` was present in two different `Atlas`es, but was a member of a relation that was fully contained in only one of those `Atlas`es, then `MultiNode.relations()` would fail if the `Atlas` that did not contain the relation was added to the `MultiAtlas` first. This was due to an assumption in `MultiNode.relations()` that does not hold in all cases. See commit `f5cbddb` for the fix, and commit `44cd90d` for a minimal unit test that exposes the bug.

A similar bug also existed for `MultiEdge`, but in that case `MultiAtlasBorderFixer` was incorrectly performing an overwrite with a `MultiMapWithSet.putAll` call when it should have been performing an explicit set union. This should be fixed now as well.

Bugs also existed in `MultiPoint`, `MultiLine`, `MultiArea`, and `MultiRelation`. These have been fixed as well.

See the special unit test `MissingMultiEntityRelationTest` for full exposure of the bugs. It should now pass in all cases.


src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlasBorderFixer.java
```diff
@@ -264,7 +266,16 @@ protected void fixBorderIssues()
 
                 // Persist relations
                 relationsToUpdate.addAll(candidateRelations);
```
lucaspcram: Is set wipeout occurring here as well? This may be worth confirming.

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlasBorderFixer.java
```diff
@@ -264,7 +266,16 @@ protected void fixBorderIssues()
...
-                relationMembersToUpdate.putAll(candidateRelationMembers);
```
lucaspcram: NOTE: This is not the only place where `putAll` is called. There may be set wipeout happening in other spots.

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
```diff
@@ -516,7 +516,8 @@ public Area area(final long identifier)
     @Override
     public Iterable<Area> areas()
     {
-        return Iterables.translate(this.areaIdentifierToAtlasIndex, identifier -> area(identifier));
+        return Iterables.translate(this.areaIdentifierToAtlasIndices,
+                identifier -> area(identifier));
```
MikeGost: Nit: You can shorten this slightly using Java 8's: `this::area`. Up to you!

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
```diff
@@ -571,7 +572,8 @@ public Line line(final long identifier)
     @Override
     public Iterable<Line> lines()
     {
-        return Iterables.translate(this.lineIdentifierToAtlasIndex, identifier -> line(identifier));
+        return Iterables.translate(this.lineIdentifierToAtlasIndices,
+                identifier -> line(identifier));
```
MikeGost: `this::line`

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
```diff
@@ -842,9 +844,17 @@ protected void setEdgeIdentifierToAtlasIndex(final LongToIntegerMap edgeIdentifi
      *            The identifier to query
      * @return The {@link Area} with this identifier
```
MikeGost: Comment needs an update!

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlasBorderFixer.java
```diff
@@ -836,6 +847,28 @@ private boolean hasInconsistentIdentifier(final long osmIdentifier,
         return false;
     }
 
+    private boolean hasInconsistentRelations(final MultiMap<Long, Edge> identifierToEdgeList)
```
MikeGost: Can you add a javadoc here?

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
lucaspcram: I'll do that. I actually didn't write that code, git just picked it up because I edited the name of the MultiMap. Looks like a good improvement though, so changed.

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiArea.java
```diff
@@ -1,12 +1,15 @@
...
 
+import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 
 import org.openstreetmap.atlas.geography.Polygon;
 import org.openstreetmap.atlas.geography.atlas.items.Area;
 import org.openstreetmap.atlas.geography.atlas.items.Relation;
 
+import com.google.common.collect.Sets;
```
jwpgage: Looks like we need a core `Sets` implementation :)

lucaspcram: Lol

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiArea.java
```diff
@@ -39,25 +42,38 @@ public long getIdentifier()
...
-        return this.getSubArea().getTags();
+        return this.getRepresentativeSubArea().getTags();
     }
 
     @Override
     public Set<Relation> relations()
     {
-        return multiAtlas().multifyRelations(getSubArea());
+        Set<Relation> unionOfAllParentRelations = new HashSet<>();
+        for (final Area subArea : getSubAreas().getSubAreas())
```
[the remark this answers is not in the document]
lucaspcram: I'm going to leave the "sub" terminology/interface for now, as it is what @matthieun already established in the `MultiNode` and `MultiRelation` classes. We can address coming up with more descriptive naming in a future PR. Ideas are welcome! :)

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
```diff
@@ -868,23 +876,31 @@ protected Edge subEdge(final long identifier)
...
      *            The identifier to query
-     * @return The {@link Line} with this identifier
+     * @return The {@link Line}s that have this identifier
      */
-    protected Line subLine(final long identifier)
+    protected SubLineList subLines(final long identifier)
     {
-        return this.atlases.get(this.lineIdentifierToAtlasIndex.get(identifier)).line(identifier);
+        final List<Line> subLines = new ArrayList<>();
+        for (final int index : this.lineIdentifierToAtlasIndices.get(identifier))
```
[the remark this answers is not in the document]
lucaspcram: What would be a good initial capacity? I'm fairly certain that the `subLine` list for a MultiLine is unlikely to ever be larger that the `ArrayList` default initial capacity - considering a `Line` is unlikely to be shared between more than 4 shards.

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiArea.java
```diff
@@ -39,25 +42,38 @@ public long getIdentifier()
         return this.identifier;
     }
 
+    public SubAreaList getSubAreas()
```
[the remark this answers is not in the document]
lucaspcram: Concerning thread safety, yes this is a valid concern. I will point out that the `MultiAtlas` was already *not* thread safe to begin with before this PR, so we're not necessarily violating any contracts with users of the class. That being said, it may still be worth it to consider this, and update the classes in the `multi` package accordingly!

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiArea.java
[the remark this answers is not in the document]
MikeGost: Agreed with confusion, but this follows existing logic. Let's leave as is, these class will need a refactor with the upcoming switch to PBF ingest refactor atlases.

PR #127, opened 2018-05-17
Performance updates for AtlasEntityPolygonFilter

jklamer:
Stores filter polygons in Rtree to reduce average number of intersection calculations for multiple filter GeometricSurfaces and a large number of objects.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilter.java
@@ -16,6 +16,7 @@ import org.openstreetmap.atlas.geography.Polygon;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasEntity;
 import org.openstreetmap.atlas.geography.converters.PolygonStringFormat;
 import org.openstreetmap.atlas.geography.index.QuadTree;
+import org.openstreetmap.atlas.geography.index.RTree;
 import org.openstreetmap.atlas.utilities.configuration.Configuration;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -37,7 +38,7 @@ public final class AtlasEntityPolygonsFilter implements Predicate<AtlasEntity>, 
     private static final Logger logger = LoggerFactory.getLogger(AtlasEntityPolygonsFilter.class);
     private static final long serialVersionUID = -3474748398986569205L;
     private Type filterType;
-    private List<GeometricSurface> geometricSurfaces = new ArrayList<>();
+    private RTree<GeometricSurface> geometricSurfaces = new RTree<>();
     private IntersectionPolicy intersectionPolicy;
 
     public static Collection<GeometricSurface> createSurfaceCollection(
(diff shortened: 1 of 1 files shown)


jklamer:
Performance numbers look good. 
Methodology: Load an modernly sized atlas with a good number of complex entities using AtlasResourceLoader with a number of different shape/combination filters. Similar performance numbers are achievable through a sub atlas. The different configurations were:
- One Square Filter 
-Two Square Filter
-One Polysided shape Filter
-Two Polysided shape Filter
-Multiple of both forms of shape Filter

For an Include Filter, my load times for each implementation were
```
List 
Include::
one square: 9.781 seconds
two square: 9.875 seconds
polysided: 9.554 seconds
two polysided: 13.189 seconds
multi multi: 18.261 seconds

one square: 10.178 seconds
two square: 10.687 seconds
polysided: 9.020 seconds
two polysided: 12.422 seconds
multi multi: 17.703 seconds

one square: 10.203 seconds
two square: 9.927 seconds
polysided: 8.719 seconds
two polysided: 13.261 seconds
multi multi: 18.954 seconds

Rtree
Include::
one square: 10.641 seconds
two square: 8.492 seconds
polysided: 9.007 seconds
two polysided: 8.803 seconds
multi multi: 9.036 seconds

one square: 10.916 seconds
two square: 8.669 seconds
polysided: 8.498 seconds
two polysided: 8.985 seconds
multi multi: 9.203 seconds

one square: 9.874 seconds
two square: 8.490 seconds
polysided: 8.646 seconds
two polysided: 9.000 seconds
multi multi: 9.506 seconds
```

Using the same shapes for an exclude filter yielded the following times for each implementation 
```
List
Exclude::
one square: 15.414 seconds
two square: 15.515 seconds
polysided: 13.925 seconds
two polysided: 17.057 seconds
multi multi: 23.333 seconds

one square: 15.739 seconds
two square: 14.534 seconds
polysided: 13.971 seconds
two polysided: 17.792 seconds
multi multi: 22.725 seconds

one square: 15.113 seconds
two square: 14.688 seconds
polysided: 15.426 seconds
two polysided: 17.615 seconds
multi multi: 22.517 seconds

Rtree 
Exclude::
one square: 15.238 seconds
two square: 14.124 seconds
polysided: 13.766 seconds
two polysided: 13.590 seconds
multi multi: 13.602 seconds

one square: 16.483 seconds
two square: 14.350 seconds
polysided: 13.892 seconds
two polysided: 14.786 seconds
multi multi: 15.394 seconds

one square: 16.861 seconds
two square: 15.581 seconds
polysided: 13.774 seconds
two polysided: 13.906 seconds
multi multi: 14.344 seconds
```

To note: This time relations will vary significantly based on the size of atlas being filtered as well as the size and number of filter polygons. 


The Rtree times have the desired relationship of fairly constant performance independent of the type or number of polygons used in the filter. The Rtree is also significantly faster for larger and more complex filters.

These results make sense because both .bounds() and .overlaps() iterate through the geometries of the entity. In the list implementation we were are calling .overlaps() for multiple polygons in some cases, where we can minimize that with the RTree implementation, but we guarantee two geometry iterations for any entity whose bound returns a polygon from the Rtree. This is why the Rtree implementation is somewhat less performant for one single polygon filter. However in every other case it reduces the number of iterations through entity geometry.

PR #130, opened 2018-05-22
Geometric surface for sharding

jklamer:
Use GeometricSurface to simplify and expand sharding code.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/sharding/Sharding.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/sharding/Sharding.java
@@ -1,18 +1,13 @@
 package org.openstreetmap.atlas.geography.sharding;
 
 import java.io.Serializable;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Set;
 
 import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.geography.GeometricSurface;
 import org.openstreetmap.atlas.geography.Location;
-import org.openstreetmap.atlas.geography.MultiPolygon;
 import org.openstreetmap.atlas.geography.PolyLine;
-import org.openstreetmap.atlas.geography.Polygon;
 import org.openstreetmap.atlas.geography.Rectangle;
 import org.openstreetmap.atlas.streaming.resource.File;
-import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.StringList;
 
 /**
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/geography/sharding/Sharding.java
```diff
@@ -81,39 +76,11 @@ static Sharding forString(final String sharding)
     /**
      * Generate shards. This needs to be deterministic!
      *
-     * @param multiPolygon
+     * @param surface
      *            The bounds to limit the shards.
      * @return The shards {@link Iterable}.
      */
-    default Iterable<? extends Shard> shards(final MultiPolygon multiPolygon)
```
jwpgage: Why is there no longer a default?

src/test/java/org/openstreetmap/atlas/geography/sharding/SlippyTileTest.java
```diff
@@ -66,9 +68,9 @@ public void testMultiPolygonShards()
                 .wkt("MULTIPOLYGON (((40 40, 20 45, 45 30, 40 40)),"
                         + "((20 35, 10 30, 10 10, 30 5, 45 20, 20 35),"
                         + "(30 20, 20 15, 20 25, 30 20)))");
-        @SuppressWarnings("unchecked")
-        final Set<Shard> coveredShards = (Set<Shard>) new SlippyTileSharding(8)
-                .shards(multiPolygon);
+        final Set<Shard> coveredShards = StreamSupport
+                .stream(new SlippyTileSharding(8).shards(multiPolygon).spliterator(), false)
```
jwpgage: Did you notice a performance increase using `spliterator()`?

jklamer: Not sure. This was just implement set construction in one line. The previous test used the knowledge that the returned iterable was a Set and directly cast to that. That broke when I got rid of the default.

src/main/java/org/openstreetmap/atlas/geography/sharding/Sharding.java
jklamer: Both Polygon and MultiPolygon are GeometricSurfaces. That interface contains the contract for the overlapping logic that is in this default function. Ideally, all MultiPolygon intersection is handled by the shape class and no implementation knowledge is needed. Deleting this removes duplicate work and prevents it breaking if MultiPolygon implementation is ever changed.

PR #136, opened 2018-06-06
Allow force-country-slicing based on a predicate

matthieun:
For the country slicing step, this allows the user to specify a `Predicate<Taggable>` for which all the features will undergo slicing even if they intersect one country only.

This allows slicing for specific features that shoot out of a country into the international waters. Those are not sliced by default.


src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1125,8 +1134,10 @@ public int size()
         List<Polygon> candidates = this.query(target.getEnvelopeInternal());
 
         // Performance improvement, if only one polygon returned no need to do any further
-        // evaluation.
-        if (isSameCountry(candidates))
+        // evaluation (except when geometry has to be sliced at all costs)
+        if (isSameCountry(candidates)
+                && (sources == null || sources.length == 0 || this.forceSlicingPredicate == null
+                        || !this.forceSlicingPredicate.test(sources[0])))
```
mertemin: Two things: 
- Only the first element in `sources` is being used. This might be intentional, but it could be source of confusion.
- If predicate is added to stop slicing operation, then I'd suggest reverting it's conditional, applying it once before collecting candidates for slice operation - right where method starts. If it fails to pass, do not slice at all.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1089,6 +1090,11 @@ public void initializeGridIndex(final Set<String> countries)
         }
     }
 
+    public void setForceSlicingPredicate(final Predicate<Taggable> forceSlicingPredicate)
+    {
+        this.forceSlicingPredicate = forceSlicingPredicate;
```
mertemin: Name doesn't really indicate what this predicate is testing. A name like `shouldSlice` or `shouldNotSlice` could be more clear.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1212,7 +1223,10 @@ public int size()
         }
 
         // Performance: short circuit, if all intersected polygons in same country, skip cutting.
-        if (isSameCountry(candidates))
+        // (except when geometry has to be sliced at all costs)
+        if (isSameCountry(candidates)
+                && (sources == null || sources.length == 0 || this.forceSlicingPredicate == null
```
MikeGost: It might be worth extracting the sources and slicing logic to a private method to avoid duplication. The multiple sources is a bit confusing as well, as @mertemin pointed out.

src/test/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMapTest.java
```diff
@@ -228,6 +229,35 @@ public void testFeatureRightByCountryBoundary() throws ParseException
...
+        final CountryBoundaryMap map = CountryBoundaryMap.fromPlainText(new InputStreamResource(
+                CountryBoundaryMapTest.class.getResourceAsStream("HTI_DOM_osm_boundaries.txt.gz"))
+                        .withDecompressor(Decompressor.GZIP));
+        Assert.assertFalse(map.hasGridIndex());
+        final Set<String> countries = new HashSet<>();
+        countries.add("HTI");
+        countries.add("DOM");
+        map.initializeGridIndex(countries);
+        // Crosses HTI only and falls in the international waters on both sides
+        final LineString lineString = (LineString) TestUtility.createJtsGeometryFromWKT(
```
MikeGost: Nice test!

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
matthieun: Thanks for the comments! The multiple sources come from the use of `Taggable...` which allows me to make it totally optional and ensure I do not break the current API... I can rename it to source though.

According to changing the logic of that code, I am not too sure. I'd rather leave the flow as-is for that PR.

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
matthieun: Sounds good!

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
matthieun: Will do

matthieun:
I think this needs more testing...

src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/slicing/CountrySlicingProcessor.java
```diff
@@ -877,7 +878,12 @@ private void processWays()
             way.getTags().add(new Tag(ISOCountryTag.KEY, ISOCountryTag.COUNTRY_MISSING));
             return Optional.empty();
         }
-        else if (slices.size() == 1 || CountryBoundaryMap.isSameCountry(slices))
+        else if (slices.size() == 1 || CountryBoundaryMap.isSameCountry(slices)
+                // In case a way is "force-sliced", when it shoots into international waters, the
+                // slice number can be >1 and the country might be the same. In that case, we still
+                // want to include both slices always. This last check is to skip the coming block
+                // in that case.
+                && !this.boundaryMap.shouldForceSlicing(Taggable.with(way.getTags())))
```
MikeGost: I think we need to apply this to the new Raw Atlas flow as well.

matthieun: Sure. Can you give me a pointer to where and I can add that to this PR too!

MikeGost: https://github.com/osmlab/atlas/blob/8e2c26e3bf1621d78de19cb8612e72303f260071/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java#L211

PR #138, opened 2018-06-13
Raw Atlas Bug Fixes

MikeGost:
Fixing two issues here:

1. Fixing an issue when way-sectioning ran into an issue sectioning ways that had stacked nodes on top of each other. The existing check was checking the case when the start and end node were equal and making sure that the two points were not adjacent. The correct check is to make sure the current and previous adjacent nodes are not identical when the start and end node are identical. This is now fixed.

2. Fixing an issue when overlapping OSM Nodes were merged irrespective of layer tags. This caused issues because in the way-sectioning logic, we would start to connect Ways that had different layer tags. The fix is two-part: a. Look at the individual node tagging - if the two or more nodes to be merged have different layer tags, don't merge the nodes. b. If the individual nodes have no tags (most frequent case) and they are shape points of some way - look at the ways and if the two or more ways have different layer tags, don't merge the nodes. In both cases - we still add the new synthetic_duplicate_osm_node tag to both the duplicated nodes to make them easy to identify and fix in OSM. This change caused a few unit test case failures. The most notable being 4 new points added to the raw atlas generation piece. I've walked through each case to verify the validity of the added points. Here are two examples:

- Example 1: Way [564131194](https://www.openstreetmap.org/way/564131194), with layer 0 has Node [4867644195](https://www.openstreetmap.org/node/4867644195). This overlaps Node [4867644214](https://www.openstreetmap.org/node/4867644214) from Way [494985585](https://www.openstreetmap.org/way/494985585), which has layer -1. Because of the difference in layer tags, these two ways (both becoming edges) should not be connected and way-sectioned at this point. This is now reflected in the final atlas. Both nodes are at the identical location and should be combined in OSM.

- Example 2: Way [468892386](https://www.openstreetmap.org/way/468892386), with layer 0 has Node [4639536827](https://www.openstreetmap.org/node/4639536827). This overlaps Node [4633475126](https://www.openstreetmap.org/node/4633475126) from Way [468892406](https://www.openstreetmap.org/way/468892406), which is layer 1. Again, these two nodes shouldn't be merged and are not after this fix.


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -276,6 +281,21 @@ private void countOsmPbfEntities()
         logger.info("Counted PBF Entities in {}", countTime.elapsedSince());
     }
 
+    private long getLayerTagValueForPoint(final Atlas atlas, final long identifier)
+    {
+        final Optional<Long> layerValue = LayerTag.getTaggedOrImpliedValue(atlas.point(identifier),
```
AlexHsieh22: Thoughts on updating getTaggedOrImpliedValue to always return the implied value?  It looks like the only case where Optional.empty() can be returned is when the layer tag is populated but invalid.  If we fix that up, getTaggedOrImpliedValue doesn't need to return an optional anymore and will make this code (and other getLayerTag* calls) much simpler!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -446,10 +488,12 @@ private Atlas rebuildAtlas(final Atlas atlas, final Set<Long> pointsToRemove,
...
      * <li>2. There are multiple {@link Point}s at a {@link Location}. In this case, we sort all the
      * points, keep the one with the smallest identifier, add a {@link SyntheticDuplicateOsmNodeTag}
-     * and remove the rest of the duplicate points. Note: we can potentially be tossing out OSM
-     * Nodes with non-empty tags. However, this is the most deterministic and simple way to handle
-     * this. The presence of the synthetic tag will make it easy to write an Atlas Check to resolve
-     * the data error.
+     * and remove the rest of the duplicate points. Two notes: 1. We keep Nodes if they have
+     * different layer tagging. This way, we aren't creating a false connection between an overpass
+     * and a road beneath it, which happened to have a way node at the identical location. 2. We are
+     * potentially be tossing out OSM Nodes with non-empty tags. However, this is the most
```
matthieun: We are potentially ~~be~~ tossing out

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -446,10 +488,12 @@ private Atlas rebuildAtlas(final Atlas atlas, final Set<Long> pointsToRemove,
...
+     * deterministic and simple way to handle this. The presence of the synthetic tag will make it
+     * easy to write an Atlas Check to resolve the data error.
```
matthieun: So when `PackedAtlasBuilder` is called and two nodes are added with the same location, the builder will connect all the edges to one of the nodes and leave the other node connected to nothing. In the end it seems it would still connect that overpass and the below road in the final atlas. Let me know if there is something I misunderstood there!

MikeGost: There won't be two nodes at the same location. The logic would be something like this:

1. There would be two Atlas Points at that location in the raw Atlas.
2. There would be two Atlas Points at that location in the sliced raw Atlas.
3. During way-sectioning, when determining whether a raw Atlas Point is a shape-point, an Atlas Edge or an Atlas Point - we used to make this a Node (ignored layer values) and the PackedAtlasBuilder would connect the edges to this Node. Now, we factor in layer values, and instead of the raw Atlas Point become a Node in the final Atlas, it's simply a shape-point - since this location doesn't have unique tagging, is not a self-intersection and is not at a valid intersection with an edge of the same layer. When the PackedAtlasBuilder get() is called, there is no connection that is formed at the location. Just visualized a few example of this in JOSM as well. Let me know if I missed anything!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
MikeGost: I was thinking about that too when calling this, seemed a little clunky...my concern was in the case of a present, but invalid value. We can easily check for this and return... something like:

```
static Long getTaggedOrImpliedValue(final Taggable taggable, final Long impliedValue)
    {
        final Optional<String> tagValue = taggable.getTag(KEY);
        if (tagValue.isPresent())
        {
            final LongExtractor extractor = new LongExtractor();
            final Optional<Long> extractedValue = extractor.validateAndExtract(tagValue.get(),
                    LayerTag.class.getDeclaredAnnotation(Tag.class));
            if (extractedValue.isPresent())
            {
                return extractedValue.get();
            }
        }
        return impliedValue;
    }
```

However, getTaggedValue is relying on getTaggedOrImpliedValue. We can update getTaggedValue to return a Long as well, but then you would have to be doing a null check any time it's called. It seems like either a null check or Optional.empty() check is required here, without duplicating some code to remove getTaggedValue's dependency on getTaggedOrImpliedValue. I image getTaggedValue would look like (in which case you still have to do Optional check for this call):

```
static Optional<Long> getTaggedValue(final Taggable taggable)
    {
        final Optional<String> tagValue = taggable.getTag(KEY);
        if (tagValue.isPresent())
        {
            final LongExtractor extractor = new LongExtractor();
            return extractor.validateAndExtract(tagValue.get(),
                    LayerTag.class.getDeclaredAnnotation(Tag.class));
        }
        return Optional.empty();
    }
```

Thoughts?

MikeGost: TLDR: might reduce overhead but cause some code duplication.

MikeGost:
Running one more quick test at scale, then removing DNM label.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
AlexHsieh22: Hmm great point!  Thinking about this more, I think we have the dependency in the wrong order.  I think getTaggedOrImpliedValue should depend on getTaggedValue and not the other way around.  Thoughts? (Haven't fully thought through all the side effects)

This would result in:
getTaggedValue:
```
        final Optional<String> tagValue = taggable.getTag(KEY);
        if (tagValue.isPresent())
        {
            final LongExtractor extractor = new LongExtractor();
            return extractor.validateAndExtract(tagValue.get(),
                    LayerTag.class.getDeclaredAnnotation(Tag.class));
        }
        return Optional.empty();
```

getTaggedOrImpliedValue:
```
static Long getTaggedOrImpliedValue(final Taggable taggable, final Long impliedValue)
    {
        final Optional<Long> taggedValue = getTaggedValue(taggable);
        return taggedValue.isPresent() ? taggedValue.get() : impliedValue;
    }
```

MikeGost: I agree, this would work better!

PR #140, opened 2018-06-18
Added a simple Python version of the Atlas

lucaspcram:
See the readme under pyatlas/README.md for more information.


pyatlas/package.sh
```diff
@@ -0,0 +1,148 @@
...
+
+# copy the LICENSE to the pyatlas folder
+cp "$gradle_project_root_dir/LICENSE" "$pyatlas_root_dir"
+
+# grab the build version from gradle.properties and inject it into setup.py
+atlas_version=$(grep "version=" "$gradle_project_root_dir/gradle.properties" | cut -f2 -d "=")
+# GNU and BSD sed have different "in-place" flag syntax
+if [ "$(uname)" == "Darwin" ];
+then
+    sed -i "" "s/version=.*/version=\"$atlas_version\",/" "$pyatlas_root_dir/setup.py"
```
matthieun: Awesome!

pyatlas/pyatlas/atlas.py
```diff
@@ -0,0 +1,1239 @@
...
+        if field_name not in self._FIELD_NAMES_TO_LOAD_METHODS:
+            raise KeyError('unrecognized field {}'.format(field_name))
+
+        # reflection code to get the appropriate load method for the field we are loading
+        method_name = self._FIELD_NAMES_TO_LOAD_METHODS[field_name]
+        load_method_to_call = getattr(self, method_name)
+        load_method_to_call()
+
+
+class _PackedTagStore(object):
```
matthieun: Are those included in the same file to make them "private"?

pyatlas/pyatlas/atlas_entities.py
```diff
@@ -0,0 +1,933 @@
...
+            return relation_set
+
+        for relation_index in relation_map[index]:
+            relation = Relation(self.get_parent_atlas(), relation_index)
+            relation_set.add(relation)
+
+        return frozenset(relation_set)
+
+
+class Point(AtlasEntity):
```
matthieun: Pretty cool flyweight implementation!

pyatlas/pyatlas/geometry.py
```diff
@@ -0,0 +1,577 @@
...
+        longitude = degree_as_dm7(longitude)
+
+        locations.append(Location(latitude, longitude))
+
+    return locations
+
+
+def _urshift32(to_shift, shift_amount):
+    """
+    Perform a 32 bit unsigned right shift (drag in leading 0s).
```
matthieun: Nice!

pyatlas/pyatlas/spatial_index.py
```diff
@@ -0,0 +1,184 @@
...
+
+        self.tree = _CustomSTRtree(hacktree_arguments)
+
+    def add(self, entity):
+        """
+        Insert an AtlasEntity into the _RTree. The underlying STRtree is
+        immutable, so this method forces a rebuild of the entire tree.
+        """
+        self.contents.append(entity)
+        self._construct_tree_from_contents()
```
matthieun: Would it make sense to have two methods, one called "add" and another one called "finishedAdding"? That way you do not build the rtree at every call?

pyatlas/unit_tests/test_atlas.py
```diff
@@ -0,0 +1,139 @@
+import unittest
+
+from pyatlas.atlas import Atlas
+from pyatlas import geometry
+from pyatlas.geometry import Rectangle
+
+
+class AtlasTest(unittest.TestCase):
```
matthieun: If those tests fail, are they build breaking?

lucaspcram:
Glad to get this thing started!

Yes the `__init__.py` is a pythonism that allows the interpreter to detect a directory as a python package. [Here is some info on the subject](https://docs.python.org/3/tutorial/modules.html#packages)

matthieun:
I was wondering if it was ok to have it in the `autogen` folder. Thanks!

pyatlas/pyatlas/atlas.py
lucaspcram: It's not the fact that they are included in the same package, rather the leading `_` character is what makes them private. You can see in the `geometry` module that there are many classes defined there, none of which are private.

The reason I placed `_PackedTagStore` and `_IntegerDictionary` in the `atlas` module is because they are only called from within that module, and they don't need to be exposed to the user for import or documentation.

pyatlas/package.sh
lucaspcram: A bit of a hack but it works! Maintaining it is my problem anyway :)

pyatlas/unit_tests/test_atlas.py
lucaspcram: Yes, failing unit tests cause the build to abort with a diagnostic detailing the failed test and the cause of the failure. It's very similar to the JUnit setup we have in Java Atlas.

pyatlas/pyatlas/atlas_entities.py
lucaspcram: I take no credit for this design. Copied straight from your `PackedAtlas` :)

pyatlas/pyatlas/spatial_index.py
lucaspcram: I suppose I could change the name to something different, just to make it more clear.

The reason I implemented it this way is because the STR packed R-tree provided by GEOS (thru Shapely) is effectively immutable once it has been built (you can see what I mean by "built" by [checking out this class](https://github.com/OSGeo/geos/blob/master/src/index/strtree/AbstractSTRtree.cpp), "building" occurs as a side effect of queries to the tree ).
So effectively, the only way to add elements to the tree once it has been "built" is to rebuild the tree from scratch. However, this limitation *does not* apply when constructing the tree for the first time, which is why I provided the `initial_entities` parameter in the constructor of the `RTree` class.

Ideally, users will not be building their own `RTrees` anyway. The `Atlas` lazily constructs the spatial indices it needs at query time. However, I made this class is public in case someone really wants to extend the spatial capabilities of an Atlas without necessary modifying the `Atlas` class itself.

lucaspcram:
You have to, since the `autogen` folder is included as a subpackage by other modules in the main `pyatlas` package. You can see at the top of `atlas.py` that it has lots of `import autogen.FOO_pb2` statements. Hope that clears everything up! :-)

pyatlas/pyatlas/spatial_index.py
matthieun: It would be awesome to add all this as a comment in that method!

README.md
```diff
@@ -160,6 +160,10 @@ final Atlas atlas2;
 new PackedAtlasCloner().cloneFrom(new MultiAtlas(atlas1, atlas2)).save(new File("/path/to/file.atlas"));
 ```
 
+# PyAtlas
+
+`Atlas` features a lightweight Python version of the core functionality. The `pyatlas` is available as a Python package from PyPI, but can also be built and installed locally. Check out the `pyatlas` subfolder and the `pyatlas/README.md` file for more information!
```
matthieun: You could link it directly here using `... [Readme file](pyatlas/README.md) ...`

pyatlas/pyatlas/spatial_index.py
lucaspcram: Haha sounds good, I'll clean this up to be a bit more concise and to the point.

README.md
lucaspcram: Ah ok let me add that. I did not realize you could do relative URLs.

pyatlas/pyatlas/pyatlas_globalfunc.py
```diff
@@ -0,0 +1,10 @@
+"""
+This module contains global utility functions for pyatlas.
+"""
+
+
+def hello_atlas():
+    """
+    Print a welcome message!
+    """
+    print "Hello Atlas!"
```
jwpgage: Is this just a default global function to be replaced in future?

pyatlas/package.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env bash
+
+# THIS ENTIRE SCRIPT IS A MASSIVE HACK
+# THIS SHOULD REALLY BE DONE WITH GRADLE
```
jwpgage: amazing :)

pyatlas/unit_tests/test_atlas.py
```diff
@@ -0,0 +1,139 @@
...
+        # NOTE point 1 does not show up in the results because it lies on the polygon border
+        test_results = atlas.points_within(Rectangle(lower_left, upper_right))
+        self.assertEquals({atlas.point(2), atlas.point(3)}, test_results)
+
+        test_results = atlas.points_within(
+            Rectangle(lower_left, upper_right), lambda p: p.get_identifier() % 2 != 0)
+        self.assertEquals({atlas.point(3)}, test_results)
+
+        test_results = atlas.points_at(geometry.location_with_degrees(38, -118))
+        self.assertEquals({atlas.point(1)}, test_results)
```
jwpgage: Have these been tested and verified against the Java version?

pyatlas/pyatlas/atlas.py
```diff
@@ -0,0 +1,1239 @@
...
+        if word in self.word_to_index:
+            return self.word_to_index[word]
+        self.word_to_index[word] = index
+        self.index_to_word[index] = word
+
+    def size(self):
+        return len(self.word_to_index)
+
+    def word(self, index):
+        # TODO this could throw a KeyError, how to handle?
```
MikeGost: Something to address or ignore for now?

lucaspcram: I changed the way that class is used internally, so I believe that comment is outdated now. Good catch.

pyatlas/unit_tests/test_atlas.py
lucaspcram: There is no automatic verification, but I have tested it manually against the Java version with the help of a script not included in this PR. Based on my tests the query results are consistent. But more tests are always welcome!

pyatlas/package.sh
lucaspcram: I really need to remove this comment... :P

pyatlas/pyatlas/pyatlas_globalfunc.py
lucaspcram: Good question :) It's pretty much just there for the `Getting Started` section of the README, which has the user call it to test that the package built correctly. In the future this module may have additional function definitions.

PR #141, opened 2018-06-18
Fixed bug in areasCovering that overincluded results.

lucaspcram:
Previously, the `areasCovering` method in `AbstractAtlas` did not check if the target `Location` was actually fully enclosed by the `Area` in question. This caused the resulting set to over-include `Locations` that were within the bounding box of the `Area`, but not necessarily enclosed by the true `Area` geometry.

Interestingly enough, this change seems to break the `testLocationContaining` unit test in `PackedAtlasTest`. This is because the unit test uses `Locations` that are on the boundaries of the `Areas` in question. Since the new `areasCovering` code uses `fullyGeometricallyEncloses`, points on the border of a polygon are not considered enclosed by the polygon. I have fixed the unit test to better reflect the semantics implied by `areasCovering`.

Unrelated, but also removed a misleading, outdated comment in the `RelationMember` class.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
@@ -109,7 +109,7 @@ public abstract class AbstractAtlas extends BareAtlas
     public Iterable<Area> areasCovering(final Location location)
     {
         final Iterable<Area> areas = this.getAreaSpatialIndex().get(location.bounds());
-        return Iterables.filter(areas, area ->
+        return Iterables.stream(areas).filter(area ->
         {
             final Polygon areaPolygon = area.asPolygon();
             return areaPolygon.fullyGeometricallyEncloses(location);
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
@@ -108,7 +108,12 @@ public abstract class AbstractAtlas extends BareAtlas
     @Override
     public Iterable<Area> areasCovering(final Location location)
     {
-        return this.getAreaSpatialIndex().get(location.bounds());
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
```diff
@@ -108,7 +108,12 @@ public static Atlas forOsmPbf(final Resource resource)
     @Override
     public Iterable<Area> areasCovering(final Location location)
     {
-        return this.getAreaSpatialIndex().get(location.bounds());
+        final Iterable<Area> areas = this.getAreaSpatialIndex().get(location.bounds());
+        return Iterables.filter(areas, area ->
```
matthieun: This works well, however how about this:
```
Iterables.stream(areas).filter(area -> {
    final Polygon areaPolygon = area.asPolygon();
    return areaPolygon.fullyGeometricallyEncloses(location);
});
```
You choose the one you find the most readable :)

src/main/java/org/openstreetmap/atlas/geography/atlas/items/RelationMember.java
```diff
@@ -4,8 +4,7 @@
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 
 /**
- * A {@link Relation} member. It has a role and an {@link AtlasEntity}. The {@link AtlasEntity} may
- * be null in case the {@link Atlas} that created it has a sliced {@link Relation}.
+ * A {@link Relation} member. It has a role and an {@link AtlasEntity}.
```
matthieun: +1

lucaspcram:
@jwpgage Indeed. I believe it is due to the way we implement `Polygon.fullyGeometricallyEncloses(Location)`. In the case that there is no AWT overflow, it defaults to the AWT `Polygon.contains` implementation, [which defines "insideness" like so.](https://docs.oracle.com/javase/8/docs/api/java/awt/Shape.html#def_insideness). If you plot out the test geometry, it looks like the test `Locations` violate rule 2 of the definition I linked.

PR #142, opened 2018-06-19
Aerialway & Building Tag Updates

Bentleysb:
Added `KEY` to `AerialWayTag` and `ENTRANCE` to `BuildingTag`.

[Tag:building=entrance](https://wiki.openstreetmap.org/wiki/Tag:building%3Dentrance) is a deprecated tag, but is relatively prevalent in some areas (see [taginfo](https://taginfo.openstreetmap.org/tags/building=entrance#map)).

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/tags/AerialWayTag.java
+++ b/src/main/java/org/openstreetmap/atlas/tags/AerialWayTag.java
@@ -1,6 +1,7 @@
 package org.openstreetmap.atlas.tags;
 
 import org.openstreetmap.atlas.tags.annotations.Tag;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
 
 /**
  * OSM aerialway tag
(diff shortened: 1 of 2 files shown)


MikeGost:
@Bentleysb - was looking at this one more time, are we sure we want to include entrance in building - it looks to be deprecated per the wiki here: https://wiki.openstreetmap.org/wiki/Key:building#Additional_Attributes 

It looks to be replaced with an entrance tag key. Also see this note: https://wiki.openstreetmap.org/wiki/Key:entrance#Possible_Tagging_Mistakes

Bentleysb:
It is deprecated, but still has a good bit of use in Europe, see [https://taginfo.openstreetmap.org/tags/building=entrance#map](https://taginfo.openstreetmap.org/tags/building=entrance#map).

I submit to your judgment on whether it should be included. I had been testing for it and noticed it was missing was all.

MikeGost:
Talked with @matthieun offline - let's include it since there's a large use case and this wasn't added to the `VALID_BUILDINGS EnumSet`. Thanks again!

PR #151, opened 2018-06-25
Handle null subatlas returned after filtering single resource in load()

remegraw:
Return null rather than throw NoSuchElementException.


src/test/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoaderTest.java
```diff
@@ -10,11 +10,14 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Edge;
 import org.openstreetmap.atlas.geography.atlas.items.Node;
 import org.openstreetmap.atlas.streaming.compression.Decompressor;
 import org.openstreetmap.atlas.streaming.resource.File;
 import org.openstreetmap.atlas.streaming.resource.InputStreamResource;
 import org.openstreetmap.atlas.streaming.resource.Resource;
+import org.openstreetmap.atlas.tags.HighwayTag;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
 
 import com.google.common.collect.Iterables;
```
MikeGost: Nit: I know this isn't part of your changes, but since we're here, do you mind replacing the Google `Iterables` with our implementation?

Issue #149, opened 2018-06-25
StandardConfiguration does not support iteration over Configurables.


lucaspcram:
As far as I am aware, this class does not support the following use case. Say I have a JSON file that looks like:

```
{
    "testobject1": {
        "testparam1": "testvalue1",
        "testparam2": "testvalue2",
        "testparam3": {
            "testparam4": "testvalue4"
        }
    },

    "testobject2": {
        "testparam1": "testvalue1"
    }
}
```

Assuming I have loaded the JSON file into a `Resource`, I can load the `Resource` into a `StandardConfiguration` and query it like so:

```java
final StandardConfiguration configuration = new StandardConfiguration(jsonResource);
final Configurable config = configuration.get("testobject1");
```

This gives back a `Configurable` object that lets me access the parameters in `testobject1`. However, what I would like to do is:
```java
final StandardConfiguration configuration = new StandardConfiguration(jsonResource);
for (Configurable config : configuration.allKeys())
{
    // do something with the config here
}
```
Having some kind of `allKeys` method would be extremely useful.

If this is already possible, or I am missing something obvious, please let me know! Otherwise this seems like a really useful class, and I would like to use the rest of its functionality. I just really need some way to iterate!

Thanks!

lucaspcram:
Fixed by PR #152

PR #152, opened 2018-06-25
Added a method to get a StandardConfiguration data keyset.

lucaspcram:
This should solve Issue #149.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/configuration/MergedConfiguration.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/configuration/MergedConfiguration.java
@@ -5,11 +5,13 @@ import static org.openstreetmap.atlas.utilities.collections.Iterables.join;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -81,6 +83,16 @@ public class MergedConfiguration implements Configuration
 
     private final List<Configuration> configurations;
 
+    public MergedConfiguration(final Configuration... configurations)
+    {
+        this(Arrays.asList(configurations));
+    }
+
+    public MergedConfiguration(final List<Configuration> configurations)
+    {
+        this.configurations = Collections.unmodifiableList(configurations);
+    }
+
     public MergedConfiguration(final Resource first, final Iterable<Resource> configurations)
     {
         final LinkedList<Configuration> mergedConfigurations = new LinkedList<>();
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/utilities/configuration/MergedConfiguration.java
```diff
@@ -81,29 +83,42 @@
...
     }
 
+    @Override
+    public Set<String> configurationDataKeySet()
+    {
+        // merge the keysets of the underlying StandardConfigurations
+        final Set<String> keySet = new HashSet<>();
+        this.configurations.forEach(configuration ->
+        {
+            keySet.addAll(configuration.configurationDataKeySet());
```
MikeGost: Nit: I don't think we want to handle this, but this potentially has unexpected behavior in the case of two underlying configurations having the same key and us merging them into one.

jklamer: If two underlying configurations have the same key, then there is only one value returned for that key, so merging makes sense to me looking at the configuration level.

lucaspcram: This is a good point. I have not grokked the `MergedConfiguration` class fully enough to understand what happens when you have two underlying configurations with duplicate top level keys. What occurs, in that case, when calling `configuration.get("duplicateTopLevelKey")`? The answer to this question will determine what sort of behavior the `MergedConfiguration#configurationDataKeySet` method should exhibit. Maybe @jklamer or @davis20 can chime in here with some clarification?

To be honest, my initial thought was to leave this method unimplemented for `MergedConfiguration` and just throw a `CoreException("Not implemented.")`. Especially since this method is only being added to `StandardConfiguration` to support my very specific use case. I added this slipshod implementation to `MergedConfiguration` for consistency, but perhaps it might be better to let @jklamer or @davis20 implement this method for `MergedConfiguration` in another PR - since they know/use this class much better than I.

jklamer: The number of keys returned here is the the number of top level endpoints available(from multiple configurations if needed). If two stacked configurations have the same key, then the value on the higher level configuration is returned and the lower configuration value isn't even available through the interface. This implementation makes sense given the design of the interface and what is exposed to the user. 

For example, if two configurations have a "key1", and we returned two copies of that "key1", it would be disingenuous, because calling configuration.get("key1").value() twice would still only return the value of the higher level configuration.

lucaspcram: I've added a doc comment to the `configurationDataKeySet` method in `MergedConfiguration`, just to make sure the behaviour is absolutely clear to the user.

MikeGost:
Will leave it to @lucaspcram to close the corresponding issue!

Issue #156, opened 2018-07-06
A potential bug in JtsMultiPolyLineConverter


Karpinchyk:
We are getting a ClassCastException when trying to convert a MultiPolyLine to WKT.

Here is the stack trace:
```
Exception in thread "main" java.lang.ClassCastException: [Ljava.lang.Object; cannot be cast to [Lcom.vividsolutions.jts.geom.LineString;
	at org.openstreetmap.atlas.geography.converters.jts.JtsMultiPolyLineConverter.convert(JtsMultiPolyLineConverter.java:42)
	at org.openstreetmap.atlas.geography.converters.WktMultiPolyLineConverter.convert(WktMultiPolyLineConverter.java:39)
...
```

This fix in src/main/java/org/openstreetmap/atlas/geography/converters/jts/JtsMultiPolyLineConverter.java is working for us:

```
     private static final JtsCoordinateArrayConverter COORDINATE_ARRAY_CONVERTER = new JtsCoordinateArrayConverter();
+    private static final JtsPolyLineConverter POLYLINE_CONVERTER = new JtsPolyLineConverter();
     private static final GeometryFactory FACTORY = JtsPrecisionManager.getGeometryFactory();
 
     @Override
@@ -38,8 +43,10 @@ public class JtsMultiPolyLineConverter implements TwoWayConverter<MultiPolyLine,
     @Override
     public MultiLineString convert(final MultiPolyLine multiPolyLine)
     {
-        final LineString[] lineStrings = (LineString[]) ImmutableList
-                .copyOf(multiPolyLine.iterator()).toArray();
-        return new MultiLineString(lineStrings, FACTORY);
+        final List<LineString> lineStringList =
+            StreamSupport.stream(
+                    Spliterators.spliteratorUnknownSize(multiPolyLine.iterator(), Spliterator.ORDERED), false).
+                    map(pl -> POLYLINE_CONVERTER.convert(pl)).collect(Collectors.toList());
+        return new MultiLineString(lineStringList.toArray(new LineString[lineStringList.size()]), FACTORY);
     }
```

MikeGost:
Can you open a PR with the fix? Two minor requests:

1. Can you add a test to the MultiPolyLineTest that converts the MultiPolyLine to WKT. 
2. Consider using the below simplification of your fix:

```
    @Override
    public MultiLineString convert(final MultiPolyLine multiPolyLine)
    {
        final List<LineString> lineStringList = Iterables.stream(multiPolyLine)
[...]

PR #157, opened 2018-07-09
Fixing JtsMultiPolyLineConverter behavior

MikeGost:
Resolves https://github.com/osmlab/atlas/issues/156. New test was failing before fix and passes after the fix. The test class shows many changes, only a single test was added: `testCreateMultiLineStringFromMultiPolyLine` - the rest is Eclipse reformatting.


lucaspcram:
Looks good @MikeGost . Not super important, but the `Lists.newArrayList` call on line 29 in the `JtsMultiPolylineConverter` class is deprecated by Google for Java 7+. Definitely not a breaking issue, but the maintainers of the `Lists` class now recommend using the `ArrayList<>()` syntax instead of `Lists.newArrayList`.

MikeGost:
Thanks @lucaspcram - updated!

PR #158, opened 2018-07-09
Snake features losing nearest neighbor country code tag

matthieun:
In case a feature sneaks in and out of a country boundary like below, the country boundary map would lose the `synthetic_nearest_neighbor_country_code=yes` tag for all the features that are outside (it would keep the country code tag though.

image

This PR makes sure that when the country-less features are a `GeometryCollection`, each part still gets all the tags from the parent.


src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -279,8 +278,26 @@ public static String getGeometryProperty(final Geometry geometry, final String k
         }
         else
         {
-            return propertyMap.get(key);
+            result.putAll(propertyMap);
```
AlexHsieh22: If propertyMap is null, and we return null, do we have to worry about NullPointerExceptions in the use cases below?  Would it be better to return the empty Map?

PR #162, opened 2018-07-10
Using a Double to convert angle to radians

jklamer:
Changing to use a double factor to convert an angle (dm7 representation) to radians. The int factor was causing an issue in ```Location::midPoint``` where the loss of precision was causing the midPoint of two Locations on the antiMeridian(-180) to have a Longitude of 179.9999999, and visa versa two Locations on the antiMeridian(180) had a midpoint  Longitude of -179.9999999.

Other tests changed to reflect new precision of output.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
@@ -21,6 +21,8 @@ public class Angle implements Serializable
     protected static final int DM7_PER_DEGREE = 10_000_000;
     // There are approximately 57 degrees per radian
     public static final int DM7_PER_RADIAN = 572_957_795;
+    // When precision is needed
+    public static final double DM7_PER_RADIAN_DOUBLE = Double.valueOf(1_800_000_000) / Math.PI;
     // An angle is >= -180 degrees
     protected static final int MINIMUM_DM7 = -1_800_000_000;
     // An angle is < 180 degrees
@@ -160,7 +162,7 @@ public class Angle implements Serializable
      */
     public double asRadians()
     {
-        return (double) this.asDm7() / DM7_PER_RADIAN;
+        return (double) this.asDm7() / DM7_PER_RADIAN_DOUBLE;
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
@@ -19,14 +19,14 @@ public class Angle implements Serializable
 
     // There are ten million microdegrees per degree
     protected static final int DM7_PER_DEGREE = 10_000_000;
-    // There are approximately 57 degrees per radian
-    public static final int DM7_PER_RADIAN = 572_957_795;
-    // When precision is needed
-    public static final double DM7_PER_RADIAN_DOUBLE = Double.valueOf(1_800_000_000) / Math.PI;
     // An angle is >= -180 degrees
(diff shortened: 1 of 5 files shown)


src/test/java/org/openstreetmap/atlas/geography/MultiPolyLineTest.java
```diff
@@ -82,13 +82,13 @@ public void testCreateMultiLineStringFromMultiPolyLine()
     @Test
     public void testCreateMultiPolyLineFromPolyLine()
     {
-        final PolyLine polyLine1 = PolyLine.wkt(
-                "LINESTRING (10.5553105 48.3419094, 10.5552096 48.3417501, 10.5551312 48.3416583, "
+        final PolyLine polyLine1 = PolyLine
+                .wkt("LINESTRING (10.5553105 48.3419094, 10.5552096 48.3417501, 10.5551312 48.3416583, "
```
jklamer: Gradle spotlessApply formatting. No number changes.

src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
```diff
@@ -21,6 +21,8 @@
     protected static final int DM7_PER_DEGREE = 10_000_000;
     // There are approximately 57 degrees per radian
     public static final int DM7_PER_RADIAN = 572_957_795;
```
MikeGost: Might be confusing for users of this class to have two types of the same measure. Thoughts on replacing `DM7_PER_RADIAN` to be a double and having a single constant? Not sure of the impact right away, but it's only being used in the `radians()` call right now.

src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
```diff
@@ -21,6 +21,8 @@
...
+    // When precision is needed
+    public static final double DM7_PER_RADIAN_DOUBLE = Double.valueOf(1_800_000_000) / Math.PI;
```
matthieun: I would use `MAXIMUM_DM7` here instead of `1_800_000_000`

src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
```diff
@@ -160,7 +162,7 @@ public double asPositiveRadians()
      */
     public double asRadians()
     {
-        return (double) this.asDm7() / DM7_PER_RADIAN;
+        return (double) this.asDm7() / DM7_PER_RADIAN_DOUBLE;
```
matthieun: I think you can get rid of the cast if you divide by a double. It will already yield a double.

src/test/java/org/openstreetmap/atlas/geography/MultiPolyLineTest.java
matthieun: Wow how come spotless did not catch that before!

src/main/java/org/openstreetmap/atlas/utilities/scalars/Angle.java
jklamer: The int constant is used in a number of different contexts outside the class and the precision loss may not matter there + the refactoring of all int instances to double was outside of my ambition for this PR. This alone will cause refactoring for any precise relying unit tests.

src/test/java/org/openstreetmap/atlas/geography/MultiPolyLineTest.java
matthieun: Created another PR in the mean time to fix that. It should seamlessly merge with your change.

PR #166, opened 2018-07-10
Add sonar integration

matthieun:
This simple update to the travis scripts add support for running sonar reports during CRON builds. The results will be available in the [Sonar osmlab project](https://sonarcloud.io/organizations/osmlab/projects):

Here is how a sample run looks like:

image


lucaspcram:
Looks awesome! But it really doesn't like the autogenerated code...

matthieun:
Yep. Next step is to go deeper and understand how to skip some parts of the codebase.

lucaspcram:
I'm actually pretty interested in looking into this. I'll investigate a little tomorrow and see if there is any Sonar documentation on it.

lucaspcram:
Here's an update to force Sonar to exclude the autogenerated code.

In `build.gradle` right below the `protobuf` configuration block, add the following code block:

```
sonarqube {
    properties {
        property "sonar.exclusions", "src/generated/**/*.java"
    }
}
```

PR #169, opened 2018-07-12
Allow objects as geojson properties

remegraw:
This PR adds support for building GeoJSON objects with objects as properties, while maintaining backward compatibility with the current support for string-only properties. I deprecated the string-specific code, thinking that over time clients can migrate to using objects as properties and we can eventually remove the string-specific methods. But until that is done, there are redundant methods in both the builder and the builder tests. Test coverage for the new methods is equal to the old methods, plus a bit more to cover the cases of objects as properties.

Note that I tried to limit code cut-and-paste in the builder by having the string-specific methods call their object supporting counterparts where possible.


lucaspcram:
Looks good to me. This actually reveals some reliances on `instanceof` in `GeoJsonBuilder` that might be worth refactoring down the road. From what I can tell this PR works around them well enough for now. Thanks for the tests!

PR #171, opened 2018-07-13
Some diff readability enhancements for AtlasDelta

lucaspcram:
Previously, the diff view given by `AtlasDelta` was difficult for human readers to parse. This update aims at making it much easier to spot what changed when reviewing an `AtlasDelta` diff report.

The changes are backwards compatible, and adds a new `DiffViewFriendlyItem` interface for any entity that wants to present a diff friendly view of itself.


src/test/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaEdgeTest.java
```diff
@@ -38,7 +38,8 @@ public void testAdded()
         final Atlas alter = alterBuilder.get();
 
         final SortedSet<Diff> diffs = new AtlasDelta(base, alter).generate().getDifferences();
-        logger.debug("testAdded(): {}", Diff.toString(diffs));
+        logger.info("testAdded(): {}", Diff.toString(diffs));
+        logger.info("testAdded(): {}", Diff.toDiffViewFriendlyString(diffs));
```
matthieun: Any reason this became info from debug?

lucaspcram: I changed it to see output while debugging. Forgot to change it back :)

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Node.java
```diff
@@ -53,6 +55,23 @@ public ItemType getType()
      */
     public abstract SortedSet<Edge> outEdges();
 
+    @Override
+    public String toDiffViewFriendlyString()
+    {
+        final Set<Relation> relations = this.relations();
```
MikeGost: The same parent relation code is duplicated for every entity. Wondering if we can move this up to the `AtlasEntity` class and call `parentRelationsAsDiffViewFriendlyStrings`() in the implementations.

lucaspcram: Ah this is a good idea.

PR #174, opened 2018-07-16
Logging Instrumentation

MikeGost:
Around 3 specific operations:
1. `MultiAtlas` construction time during `DynamicAtlas` creation
2. `SubAtlas` creation time
3. `Atlas` fetch time when using the FetcherPolicy

Also revamped the started/completed logging code to make it easier to add tasks and reuse code for when something starts or finishes.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
@@ -41,6 +41,7 @@ import org.openstreetmap.atlas.streaming.resource.WritableResource;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.StreamIterable;
 import org.openstreetmap.atlas.utilities.collections.StringList;
+import org.openstreetmap.atlas.utilities.time.Time;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -136,6 +137,7 @@ public class DynamicAtlas extends BareAtlas
 
     public void buildUnderlyingMultiAtlas()
     {
+        final Time buildTime = Time.now();
         if (logger.isDebugEnabled())
         {
             logger.debug("{}: Loading MultiAtlas with {}", this.getName(),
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -162,6 +164,8 @@ public void buildUnderlyingMultiAtlas()
         {
             throw new CoreException("Cannot load shards with no data!");
         }
+        logger.trace("Built underlying MultiAtlas for {} in {}", this.getName(),
```
matthieun: It would be nice to keep the same logging pattern:
```
logger.trace("{}: Built underlying MultiAtlas in {}", this.getName(), buildTime.elapsedSince());
```

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -756,6 +778,23 @@ private boolean locationIsPartOfAnIntersectingEdgeOfTheSameLayer(final Location
...
+        logger.info(COMPLETED_TASK_MESSAGE, taskName, shardName, duration);
+    }
+
+    private void logTaskCompletionAsTrace(final String taskName, final String shardName,
+            final Duration duration)
+    {
+        logger.trace(COMPLETED_TASK_MESSAGE, taskName, shardName, duration);
+    }
+
+    private void logTaskStartedAsInfo(final String taskname, final String shardName)
```
matthieun: You could have this method log the start and return the timer!

src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
MikeGost: Good point, let me update for consistency.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good idea! This is cleaner.

MikeGost:
Not sure line number matters here? It is simply a measure of time and binary state (completed/finished). If there is an exception - you will have the stack trace and you can check for the presence/absence of a started/completed event. Let me know if you still have concerns.

Issue #165, opened 2018-07-10
Spotless does not catch issue in Test


matthieun:
https://github.com/osmlab/atlas/pull/164 fixes a spotless issue that did make it through travis even though it should not.

PR #178, opened 2018-07-17
Fix error catching issues in travis build

matthieun:
This might be a resolution for https://github.com/osmlab/atlas/issues/165


src/main/java/org/openstreetmap/atlas/utilities/archive/Extractor.java
```diff
@@ -201,9 +201,9 @@ else if (current.isDirectory())
                 }
                 else
                 {
-                    try (final BufferedOutputStream bos = new BufferedOutputStream(
+                    try (BufferedOutputStream bos = new BufferedOutputStream(
                             new FileOutputStream(outputFile));
-                            final InputStream inputStream = file.getInputStream(current))
```
jwpgage: Why were all these final keywords removed?

PR #179, opened 2018-07-17
Longitude Compare at +180 degrees

jklamer:
All ```Longitude```s are stored as an underlying ```Angle``` from [-180,180). But if a location is created at 180 the Longitude representation and functionality remain 180. However when using the underlying comparison methods of the parent ```Angle``` class, this distinction is lost as the ```this.getDm7()``` methods grabs the raw stored ```Angle``` value. So that Longitude(179).isGreaterThan(Longitude(180)) is true. 

With this change the range methods will use the ```asDm7``` method, which is overridden by ```Longitude``` to return 180 if that is what it is created with. 

One oddity this will create Longitude(-180).isGreaterThanOrEqualTo(Longitude(180)) will now return false, but Longitude(-180).equals(Longitude(180)) will return true as before. 

EDIT UPDATE:
Reverting to using ```getDm7``` when the classes are different and thus have potentially different ```asDm7``` range /values;  for instance ```Heading.isLessThan(Longitude)```.


matthieun:
Closing, will re-open when https://github.com/osmlab/atlas/pull/178 is merged. Unfortunately travis checks are not reliable until that one is merged.

jklamer:
What is your thinking? I understand wanting to keep the continuity of -180 is 180, but without treating 180 as larger in the range of Longitudes we'll get incorrect results for operations involving Rectangles with their East Side on the antiMeridian (and possibly other places).

MikeGost:
Per offline discussion with @matthieun - I would add more tests surrounding bounding boxes and intersection/overlap calls with these. Specific cases to consider are bounding boxes that stretch across the antimeridian and ones that greater than 180 degrees but don't cross the antimeridian.

jklamer:
I added a test to Rectangle that shows two different cases in which without the changes, they would fail. I confirmed this by reverting my changes and running the test. Let me know what you think. I believe the interactions displayed in the Rectangle test are the desired ones.

MikeGost:
Sorry for nit-picking - can we add more tests. Specifically:
1. Rectangle greater than 180 degrees in width not crossing anti-meridian
2. Rectangle greater than 180 degrees in width crossing anti-meridian

MikeGost:
@jklamer - does construction matter in this case? I just wanted to verify backward compatibility across Polygons > 180 degrees and wrapping around meridian.

MikeGost:
Scratch that, @jklamer and I talked offline, not a viable test case. PR looks good!

PR #180, opened 2018-07-19
Fixed some code smells.

lucaspcram:
Code smells addressed include those listed here:

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CQl6vpliCeK6yC&open=AWSBg5CQl6vpliCeK6yC

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CQl6vpliCeK6yD&open=AWSBg5CQl6vpliCeK6yD

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CQl6vpliCeK6yE&open=AWSBg5CQl6vpliCeK6yE

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CQl6vpliCeK6yF&open=AWSBg5CQl6vpliCeK6yF

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CGl6vpliCeK6x3&open=AWSBg5CGl6vpliCeK6x3

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CGl6vpliCeK6x4&open=AWSBg5CGl6vpliCeK6x4

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5CGl6vpliCeK6xy&open=AWSBg5CGl6vpliCeK6xy

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5Byl6vpliCeK6xZ&open=AWSBg5Byl6vpliCeK6xZ

https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWSBg5Byl6vpliCeK6xa&open=AWSBg5Byl6vpliCeK6xa

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Location.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Location.java
@@ -508,13 +508,12 @@ public class Location implements Located, Iterable<Location>, Serializable
                         * Math.sin((double) distance.asMillimeters()
                                 / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
                         * Math.cos(bearing));
-        final double longitude2 = longitude1 + Math.atan2(
-                Math.sin(bearing)
-                        * Math.sin((double) distance.asMillimeters()
-                                / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
-                        * Math.cos(latitude1),
-                Math.cos((double) distance.asMillimeters()
+        final double longitude2 = longitude1 + Math.atan2(Math.sin(bearing)
+                * Math.sin((double) distance.asMillimeters()
                         / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
+                * Math.cos(latitude1), Math
+                        .cos((double) distance.asMillimeters()
+                                / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
                         - Math.sin(latitude1) * Math.sin(latitude2));
         return new Location(Latitude.radiansBounded(latitude2),
                 Longitude.radiansBounded(longitude2));
--- a/src/main/java/org/openstreetmap/atlas/geography/Location.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Location.java
@@ -55,8 +55,10 @@ public class Location implements Located, Iterable<Location>, Serializable
      *            The {@link Location} as a {@link String} in "latitude(degrees),longitude(degrees)"
      *            format
      * @return The corresponding {@link Location}
+     * @throws NumberFormatException
+     *             if the latitude or longitude in the string is not a valid Double
      */
-    public static Location forString(final String locationString)
+    public static Location forString(final String locationString) throws NumberFormatException
     {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/geography/Location.java
```diff
@@ -81,8 +81,8 @@ public static Location forStringLongitudeLatitude(final String locationString)
         {
             throw new CoreException("Invalid Location String: {}", locationString);
         }
-        final double latitude = Double.valueOf(split.get(1));
```
mgcuthbert: You need to be a little careful with using parseDouble functions. They can throw NumberFormatExceptions if the string you are parsing is not valid. And it seems like you don't handle that exception in this function and the one above. I would suggest throwing it from the function.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -142,7 +141,7 @@ public Rectangle bounds()
         if (this.bounds == null && !this.isEmpty())
         {
             final Set<Location> locations = new HashSet<>();
-            forEach(polygon -> polygon.forEach(location -> locations.add(location)));
```
mgcuthbert: Nice I really like this type of code style, much easier to read in my opinion.

src/main/java/org/openstreetmap/atlas/geography/Location.java
lucaspcram: This is certainly true. In fact, the original `Double.valueOf` call had the same problem, if we check out the implementation in `java.lang.Double`

```
public static Double valueOf(String s) throws NumberFormatException {
        return new Double(parseDouble(s));
}
```

I am a little hesitant about declaring `NumberFormatException` to be thrown at the function level, since it is a `RuntimeException`, and in general we avoid declaring those explicitly in the `Atlas` codebase (clutters the function signature). However, in this case it may be worth it, since it would be possible for the caller to recover from invalid input.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -99,11 +99,10 @@ public GeoJsonObject asGeoJson()
     public GeoJsonObject asGeoJsonFeatureCollection()
     {
         final GeoJsonBuilder builder = new GeoJsonBuilder();
-        return builder.createFeatureCollection(Iterables.translate(outers(), outerPolygon ->
-        {
```
jklamer: oh common sonar, have mercy, I did this for clarity

lucaspcram: Sonar has spoken. Question not its wisdom. :D

PR #182, opened 2018-07-25
Overlapping allowed in AtlasEntityPolygonsFilter

jklamer:
Originally the `AtlasEntityPolygonFilter` didn't allow overlapping Surfaces because it allowed both Include and Exclude Surfaces, and the only way to avoid ambiguous behavior was to not allow overlapping. Overlapping is now allowed because this expands the usability and makes more sense, given if an entity is included in two overlapping surfaces or excluded in two overlapping surfaces, the behavior is the same.


src/test/java/org/openstreetmap/atlas/utilities/filters/AtlasEntityPolygonsFilterTest.java
```diff
@@ -430,7 +431,13 @@ public void testMultiPolygonForms()
                 3, 1, -1, -1);
     }
 
+    /**
+     * Ignored because of changed functionality. Functionality a legacy of first version of filter.
+     * Originally overlapping polygons were not allowed in the case of both include polygons. Now,
+     * only one type of polygon exists in the filter at once.
+     */
     @Test
+    @Ignore
```
MikeGost: Instead of ignoring, let's update it to reflect updated behavior or remove it altogether.

PR #183, opened 2018-07-31
Added generic Resource caching capabilities

lucaspcram:
**EDIT: Ready for merge**

This PR adds extensible `Resource` caching capabilities to the atlas.

The basic idea is that users can create a cache with a default population mechanism, choose a caching strategy, and then fetch desired resources (URIs) using the strategy. A very simple use case might look like:

```java
// Let's read a resource at an arbitrary URI into a ByteArrayResource
// notice the function provided to withFetcher conforms to the ResourceFetchFunction functional interface

final URI LOCAL_TEST_FILE_URI = URI.create("file:///path/to/some/file.txt");

// cache file contents into memory (a byte array)
final ConcurrentResourceCache resourceCache = new ConcurrentResourceCache(new ByteArrayCachingStrategy(), uri -> new File(uri.getPath()));

// this will cache miss the first time and read the file from disk (using the provided fetcher)
final ByteArrayResource fileBytes = new ByteArrayResource();
fileBytes.copyFrom(resourceCache.get(LOCAL_TEST_FILE_URI).get());

// this time we hit the cache, and read the bytes from memory instead of from disk
final ByteArrayResource fileBytesAgain = new ByteArrayResource();
fileBytesAgain.copyFrom(resourceCache.get(LOCAL_TEST_FILE_URI).get());



// Manually setting fetchers and CachingStrategies can be a pain.
// You can abstract all this away by creating case-specific subclasses.
// This subclass cache uses ByteArrayCachingStrategy by default and
// abstracts the get URI parameter by just taking a path string as a parameter instead.
final LocalFileInMemoryCache fileCache = new LocalFileInMemoryCache();

final ByteArrayResource fileBytes3 = new ByteArrayResource();
fileBytes3.copyFrom(fileCache.get("/path/to/another/file.txt").get());
```

See class `CachingTests` for test cases that demonstrate in detail how the caches can be used.

In an upcoming PR to `AtlasGenerator`, I will add a subclass of `ResourceCache` with an interface designed specifically for caching files with structured URIs from remote blob stores. See `LocalFileInMemoryCache` for an example of how to modify the interface of `ResourceCache`.


lucaspcram:
Haha @jklamer does this look like something you might find useful?

src/main/java/org/openstreetmap/atlas/utilities/caching/ResourceFetchFunction.java
```diff
@@ -0,0 +1,25 @@
...
+
+/**
+ * A {@link ResourceFetchFunction} is any function that takes a {@link URI} and returns the
+ * {@link Resource} at that {@link URI}. The function should return null if the resource cannot be
+ * retrieved.
+ *
+ * @author lcram
+ */
+@FunctionalInterface
+public interface ResourceFetchFunction
```
matthieun: Instead of having an interface for that, could this just be a `java.util.function.Function<URI, Resource>`?

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/SystemTemporaryFileCachingStrategy.java
```diff
@@ -0,0 +1,119 @@
...
+    {
+        return "SystemTemporaryFileCachingStrategy";
+    }
+
+    private void attemptToCacheFileLocally(final File cachedFile,
+            final ResourceFetchFunction defaultFetcher, final URI resourceURI)
+    {
+        if (!cachedFile.exists())
+        {
+            logger.info("Attempting to cache resource {} in temporary file {}", resourceURI,
```
matthieun: This being a fairly low level operation, I would log it as debug or even trace, and only bubble up the errors as warn or error.

src/main/java/org/openstreetmap/atlas/utilities/caching/ResourceFetchFunction.java
snb3300: Instead of creating a separate interface why not use the default Function interface ```Function<URI, Resource>```

src/main/java/org/openstreetmap/atlas/utilities/caching/ResourceCache.java
```diff
@@ -0,0 +1,208 @@
...
+ *
+ * @author lcram
+ */
+public class ResourceCache
+{
+    private static final Logger logger = LoggerFactory.getLogger(ResourceCache.class);
+
+    private URI resourceURI;
+    private CachingStrategy cachingStrategy;
+    private ResourceFetchFunction defaultFetcher;
```
snb3300: type of ```defaultFetcher``` can be ```Function<URI, Resource>``` instead of defining your own type. What do you think about it?

src/main/java/org/openstreetmap/atlas/utilities/caching/LocalFileInMemoryCache.java
```diff
@@ -0,0 +1,77 @@
...
+
+    /**
+     * Create a new {@link LocalFileInMemoryCache} with a given resource URI.
+     *
+     * @param resourceURI
+     *            the default resource {@link URI}
+     */
+    public LocalFileInMemoryCache(final URI resourceURI)
+    {
+        super(resourceURI, new ByteArrayCachingStrategy(), null);
```
snb3300: Any reason you are setting ```null```  ```defaultFetcher``` in the super call and then using the setter?

src/main/java/org/openstreetmap/atlas/utilities/caching/ResourceFetchFunction.java
snb3300: Oops, I guess @matthieun suggested the same

src/main/java/org/openstreetmap/atlas/utilities/caching/ResourceCache.java
lucaspcram: Good idea. @matthieun suggested this as well.

src/main/java/org/openstreetmap/atlas/utilities/caching/LocalFileInMemoryCache.java
lucaspcram: I think this was just an oversight on my part. Good catch.

lucaspcram:
This PR is ready for merge.

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/SystemTemporaryFileCachingStrategy.java
```diff
@@ -0,0 +1,123 @@
...
+                            StandardCopyOption.ATOMIC_MOVE);
+                }
+                catch (final FileAlreadyExistsException exception)
+                {
+                    logger.trace("File {} is already cached", cachedFile.toString());
+                    return;
+                }
+                catch (final Exception exception)
+                {
+                    logger.error("Something went wrong moving {} to {}",
```
MikeGost: Let's add the exception as the last parameter to the logger call, so we don't lose the stack trace if this ever happens.

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/SystemTemporaryFileCachingStrategy.java
```diff
@@ -0,0 +1,123 @@
...
+
+            final Resource resourceFromDefaultFetcher = defaultFetcher.apply(resourceURI);
+            final File temporaryLocalFile = File.temporary();
+            try
+            {
+                resourceFromDefaultFetcher.copyTo(temporaryLocalFile);
+            }
+            catch (final Exception exception)
+            {
+                logger.error("Something went wrong copying {} to temporary local file {}",
```
MikeGost: Same comment here.

src/test/java/org/openstreetmap/atlas/utilities/caching/CachingTests.java
```diff
@@ -0,0 +1,213 @@
...
+import org.openstreetmap.atlas.utilities.caching.strategies.SystemTemporaryFileCachingStrategy;
+import org.openstreetmap.atlas.utilities.scalars.Duration;
+import org.openstreetmap.atlas.utilities.time.Time;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author lcram
+ */
+public class CachingTests
```
MikeGost: Nice tests!

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/SystemTemporaryFileCachingStrategy.java
```diff
@@ -0,0 +1,123 @@
...
+        return "SystemTemporaryFileCachingStrategy";
+    }
+
+    @Override
+    public void invalidate()
+    {
+        throw new UnsupportedOperationException("Operation not supported at this time.");
+    }
+
+    private void attemptToCacheFileLocally(final File cachedFile,
```
MikeGost: If this code was copy/pasted from existing caching file, make sure to move author tags as well.

lucaspcram: It's loosely based off of part my version of `fetchAtlas` from `AtlasGeneratorHelper`, so I will add both you and Matthieu's author tags to this one.

MikeGost:
Great work here, @lucaspcram! Merging.

PR #184, opened 2018-08-01
ShardBucketCollection

jklamer:
### Description:
This is a new abstract collection implementation that allows for easy sorting of located items, like AtlasEntities or CheckFlags,  into different buckets as defined by shards. This collection acts as wrapper for the underlying bucket collections.  The bucket collection type is determined by the implementer. There are a number of control functions that need to be implemented to allow for simple and flexible use. 

Rectangle and Counter changes were made to enable the testing and creation of this class. 

Details: 
The collection lazily constructs collections per bucket. It uses synchronized blocks to ensure only one collection is initialized per shard and no information is lost. This is to allow for multithreaded insertion/sorting, but it does not guarantee the wrapped collection is thread safe. 
The bucket collections are stored in an array and the initializing shards are stored in an immutable R-tree with the index of its corresponding collection in the array. In the figure below I show the SlippyTile shards I used to initialize a number of test collections, each with the corresponding collection array index. 

shardbucketcollectiontest 002

All collection functions interface with the collection index and/or collection array. 

The collection does not support safe iterating while inserting. 

### Potential Impact:

This will change no current implementations but it will enable a number of different located sorting tasks by enabling quick and simple implementation. 

### Unit Test Approach:

I created the most basic test collection possible and tested all the main collection functionalities on that as well as some of the unique functionalities. I then created more test collections that take advantage of the flexibility of the control functions and tested that they worked as intended. 

There are multiple tests checking for the same thing in different ways to gain code coverage. 

### Test Results:

As expected. 

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
@@ -1,0 +1,454 @@
+package org.openstreetmap.atlas.utilities.collections;
+
+import java.io.Serializable;
+import java.lang.reflect.Array;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.apache.commons.lang3.ArrayUtils;
+import org.openstreetmap.atlas.geography.Located;
+import org.openstreetmap.atlas.geography.Rectangle;
+import org.openstreetmap.atlas.geography.index.RTree;
+import org.openstreetmap.atlas.geography.sharding.Shard;
+import org.openstreetmap.atlas.geography.sharding.Sharding;
+import org.openstreetmap.atlas.geography.sharding.SlippyTile;
+import org.openstreetmap.atlas.utilities.scalars.Counter;
+
+/**
+ * A collection wrapper for a set of collections associated with shards containing located items,
+ * such as CheckFlags or AtlasEntities. Supports concurrent add, contains, remove. The term bucket
+ * is used because of the usability of the collections mostly as storage for various location based
+ * sorting tasks. Does not support safe iteration while modifying.
+ *
+ * @param <LocatedType>
+ *            Located item type
+ * @param <CollectionType>
+ *            Type of the Collection of LocatedType associated with shards
+ * @author jklamer
+ */
+public abstract class ShardBucketCollection<LocatedType extends Located & Serializable, CollectionType extends Collection<LocatedType> & Serializable>
+        implements Collection<LocatedType>, Serializable
+{
+
+    /**
+     * A helper class that associates a shard with the index that its collection is at in the
+     * collectionBuckets array.
+     */
+    private static class ShardToCollectionIndex implements Located
+    {
+        private static final long serialVersionUID = 4050100671815503794L;
+        private final int index;
+        private final Shard shard;
+
+        ShardToCollectionIndex(final int index, final Shard shard)
+        {
+            this.index = index;
+            this.shard = shard;
+        }
+
+        @Override
+        public Rectangle bounds()
+        {
+            return this.getShard().bounds();
+        }
+
+        public int getIndex()
+        {
+            return this.index;
+        }
+
+        public Shard getShard()
+        {
+            return this.shard;
+        }
+    }
+
+    private static final long serialVersionUID = -7892704554302160820L;
+
+    private final CollectionType[] collectionBuckets;
+    private final RTree<ShardToCollectionIndex> collectionIndex;
+    private final HashMap<Shard, ShardToCollectionIndex> initializedShards = new HashMap<>();
+    private Rectangle maximumBounds;
+
+    public ShardBucketCollection(final Rectangle maximumBounds, final Integer zoomLevel)
+    {
+        this(maximumBounds, SlippyTile.allTiles(zoomLevel, maximumBounds));
+    }
+
+    public ShardBucketCollection(final Rectangle maximumBounds, final Sharding sharding)
+    {
+        this(maximumBounds, sharding.shards(maximumBounds));
+    }
+
+    /**
+     * Construct the collection by allocation space for a collection for each of the shards and
+     * assigning the index back.
+     * 
+     * @param maximumBounds
+     *            maximum bound of the collection
+     * @param shards
+     *            shards to use
+     */
+    @SuppressWarnings("unchecked")
+    private ShardBucketCollection(final Rectangle maximumBounds,
+            final Iterable<? extends Shard> shards)
+    {
+        this.maximumBounds = maximumBounds;
+        this.collectionIndex = new RTree<>();
+        final Counter counter = new Counter();
+        shards.forEach(shardBucket ->
+        {
+            final ShardToCollectionIndex shardToCollectionIndex = new ShardToCollectionIndex(
+                    (int) counter.getValueAndIncrement(), shardBucket);
+            this.collectionIndex.add(shardToCollectionIndex.bounds(), shardToCollectionIndex);
+        });
+        this.collectionBuckets = (CollectionType[]) Array.newInstance(
+                this.initializeBucketCollection().getClass(), (int) counter.getValue());
+    }
+
+    @Override
+    public final boolean add(final LocatedType item)
+    {
+        if (Objects.nonNull(item) && item.bounds().overlaps(this.maximumBounds))
+        {
+            final List<ShardToCollectionIndex> indexes = this.collectionIndex.get(item.bounds());
+            if (indexes.size() == 1)
+            {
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+        return false;
+    }
+
+    @Override
+    public boolean addAll(final Collection<? extends LocatedType> collection)
+    {
+        if (Objects.nonNull(collection))
+        {
+            final long addCount = collection.stream().filter(this::add).count();
+            return addCount > 0;
```
lucaspcram: Very slick.

src/test/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollectionTestClasses.java
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * Class container for test classes
+ *
+ * @author jklamer
+ */
+public class ShardBucketCollectionTestClasses
+{
+    /**
+     * Collection that will handle splitting multipolygons into buckets by overrideing the add
+     * function
```
lucaspcram: nit: overriding

src/test/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollectionTest.java
```diff
@@ -0,0 +1,297 @@
...
+        polylineBuckets2.addAll(Arrays.asList(firstBucketPolyline, lastBucketPolyline));
+        final Object[] firstCollectionArray = polylineBuckets1.toArray();
+        final PolyLine[] array1 = new PolyLine[] { firstBucketPolyline };
+        Assert.assertArrayEquals(array1, firstCollectionArray);
+        final PolyLine[] array2 = new PolyLine[] { firstBucketPolyline, firstBucketPolyline,
+                lastBucketPolyline };
+        final PolyLine[] allArray = polylineBuckets2.toArray(array1);
+        Assert.assertArrayEquals(array2, allArray);
+
+        // get code coverage
```
MikeGost: What does get code coverage mean?

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+    }
+
+    public ShardBucketCollection(final Rectangle maximumBounds, final Sharding sharding)
+    {
+        this(maximumBounds, sharding.shards(maximumBounds));
+    }
+
+    /**
+     * Construct the collection by allocation space for a collection for each of the shards and
+     * assigning the index back.
```
lucaspcram: nit: allocating

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+     * must be deterministic.
+     *
+     * @param item
+     *            to add
+     * @param collection
+     *            to add to
+     * @param shard
+     *            shard associated with the collection
+     * @return true if the collection has been added to successful and changed as a result, false
+     *         otherwise
```
lucaspcram: nit: successfully

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+            this.initializedShards.clear();
+        }
+    }
+
+    @Override
+    @SuppressWarnings("unchecked")
+    public boolean contains(final Object object)
+    {
+        final Optional<LocatedType> typedItem = this.castToLocatedType(object);
+        if (typedItem.isPresent())
```
lucaspcram: I like this pattern a lot actually. I will be stealing it ;)

lucaspcram:
Looks good @jklamer . Left a few minor documentation nits.

So help me to understand this a bit better. Is this sort of like a way to partition a collection of `Located`s based on some sharding?

Based on the logic in the `add` method, it seems to me like you give it a `Located` and it figures out (using an RTree) which collection (which is tied to a shard, hence the RTree and the `ShardToCollectionIndex` class) that `Located` belongs in.

And then from there, it looks like you can ask for all collections that overlap some arbitrary bounds (based on the `Shard` boundary with which that collection is associated), which it can compute efficiently due to having a spatial index.

Am I getting it?

What types of problems does this facilitate solving? (I am sure there are many, since you clearly put some serious effort into creating this.)

src/test/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollectionTest.java
```diff
@@ -0,0 +1,297 @@
...
+        try
+        {
+            unsupportedOperation.add(testPolyLine1);
+        }
+        catch (final UnsupportedOperationException exception)
+        {
+            caught = true;
+        }
+
+        Assert.assertTrue("Resolve shard default method is an unsupported operation", caught);
```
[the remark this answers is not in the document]
lucaspcram: +1 for this. I use this patten in `ProtoIntegerStringDictionaryAdapterTest`. Check that for an example in action.

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
jklamer: credit due to @matthieun

src/test/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollectionTest.java
jklamer: I want to use all the methods to get good coverage score for sonar.

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
jklamer: :)

jklamer:
@lucaspcram Yes exactly you are getting it. 
Problems this will help facilitate solving: sorting and helping limiting the scope of other locality related tasks.

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+    private final RTree<ShardToCollectionIndex> collectionIndex;
+    private final HashMap<Shard, ShardToCollectionIndex> initializedShards = new HashMap<>();
+    private Rectangle maximumBounds;
+
+    public ShardBucketCollection(final Rectangle maximumBounds, final Integer zoomLevel)
+    {
+        this(maximumBounds, SlippyTile.allTiles(zoomLevel, maximumBounds));
+    }
+
+    public ShardBucketCollection(final Rectangle maximumBounds, final Sharding sharding)
```
matthieun: Sharding implementations are already behaving like a spatial index. I believe here you could gain some performance and simplification by using the sharding itself to query the intersecting bounds instead of a RTree, as the scope here is much smaller that what RTrees can handle...

jklamer: Well, if I query the sharing directly I would also need a map of of shard -> index, so that would require two look-ups instead of just the one now, even if the second look up is constant time it is more steps. I would also need to store both sharding and the map of shard ->index, instead of just the RTree.

 In terms of what RTree's can handle, this allows for initializations for slippy tiles up to the SlippyTile max zoom, which is 30. At zoom 19 there are [274.9 billion tiles](https://wiki.openstreetmap.org/wiki/Slippy_map_tilenames#Zoom_levels). So even if the max bounds isn't the whole world the Rtree needs the ability to hold a lot even if my tests didn't work it that hard.

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
```diff
@@ -0,0 +1,454 @@
...
+ * is used because of the usability of the collections mostly as storage for various location based
+ * sorting tasks. Does not support safe iteration while modifying.
+ *
+ * @param <LocatedType>
+ *            Located item type
+ * @param <CollectionType>
+ *            Type of the Collection of LocatedType associated with shards
+ * @author jklamer
+ */
+public abstract class ShardBucketCollection<LocatedType extends Located & Serializable, CollectionType extends Collection<LocatedType> & Serializable>
```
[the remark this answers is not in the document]
jklamer: It looks much scarier than it is

src/main/java/org/openstreetmap/atlas/utilities/collections/ShardBucketCollection.java
matthieun: I see, if it works for you now, we can merge it!

jklamer: 😄

PR #198, opened 2018-08-16
Added an always true filter for atlas loading

lucaspcram:
### Description:

Currently, the `AtlasResourceLoader` automatically and silently skips loading an atlas resource if it does not have a `.atlas` extension. This PR provides a simple way to disable this behavior if required.

```java
new AtlasResourceLoader().withAtlasFileExtensionFilterSetTo(false).load(myAtlasResource);
```

I have some upcoming changes to `atlas-generator` that add a `HadoopAtlasFileCache`, but these are blocked until the changes in this PR are integrated and I can update `atlas-generator` to depend on them.

### Potential Impact:

This will have no downstream impact, as it is completely opt-in. All existing code will continue to function identically.

### Unit Test Approach:

I have added a unit test to `AtlasResourceLoaderTest` called `testAtlasFileExtensionFilter` that demonstrates proper functionality.

### Test Results:

All tests in `AtlasResourceLoaderTest` pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
@@ -59,10 +59,7 @@ public class AtlasResourceLoader
 
     private static final Logger logger = LoggerFactory.getLogger(AtlasResourceLoader.class);
 
-    private final Predicate<Resource> alwaysTrueAtlasFilter = resource ->
-    {
-        return true;
-    };
+    private final Predicate<Resource> alwaysTrueAtlasFilter = resource -> true;
 
     private Predicate<Resource> resourceFilter;
     private Predicate<AtlasEntity> atlasEntityFilter;
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
@@ -54,22 +54,15 @@ public class AtlasResourceLoader
         }
     }
 
-    /**
-     * @author lcram
-     */
-    private class AlwaysTrueAtlasFilter implements Predicate<Resource>
-    {
-        @Override
-        public boolean test(final Resource resource)
-        {
-            return true;
-        }
-    }
-
     public static final Predicate<Resource> IS_ATLAS = FileSuffix.resourceFilter(FileSuffix.ATLAS)
             .or(FileSuffix.resourceFilter(FileSuffix.ATLAS, FileSuffix.GZIP));
 
     private static final Logger logger = LoggerFactory.getLogger(AtlasResourceLoader.class);
+
+    private final Predicate<Resource> alwaysTrueAtlasFilter = resource ->
+    {
+        return true;
+    };
 
     private Predicate<Resource> resourceFilter;
     private Predicate<AtlasEntity> atlasEntityFilter;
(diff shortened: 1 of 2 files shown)


lucaspcram:
The `SystemTemporaryFileCachingStrategy` caches files as the UUID (computed using an MD5 sum) of their URI. So if you do something like:

```java
ConcurrentResourceCache cache = new ConcurrentResourceCache(new SystemTemporaryFileCachingStrategy(), this::fetcher);

Optional<Resource> atlasResourceOptional = cache.get(atlasURI);

Atlas atlas = new AtlasResourceLoader().load(atlasResourceOptional.get())
atlas.metaData();
```
 you end up getting a `NullPointerException` on the last line, since the `Resource` name won't have a `.atlas` file extension.

src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
```diff
@@ -54,6 +54,18 @@
         }
     }
 
+    /**
+     * @author lcram
+     */
+    private class AlwaysTrueAtlasFilter implements Predicate<Resource>
```
MikeGost: Nit: Instead of a class, why not just a `private final Predicate<Resource> alwaysTrueAtlasFilter = resource -> return true;`

lucaspcram:
Thinking about it, even this isn't the best solution. Because now, anyone using a cache for atlas `Resources` has to remember to call `withAtlasFileExtensionFilterSetTo(false)` every time they want to load a resource they got from the cache. Not really sure of a way around this though.

I'm not sure that outright removing the filtering by default would be the best idea. Although I'm also not sure why the class filters to begin with. I can't see why you would ever have an `Iterable<Resource>` where only some of the resources in the `Iterable` are atlases. Especially because when receiving an `Iterable<Resource>`, the `AtlasResourceLoader` loads them as a `MultiAtlas`.

I feel like it makes more sense to just assume the users of `AtlasResourceLoader` know what they are doing, and that any files they give are a valid atlas. If one is not, then you would get an exception thrown by the `PackedAtlasSerializer`, instead of the silent fail and null return that you currently get.

Please chime in if I am missing something, or if there is a use case I am not considering.

MikeGost:
That use case makes sense, thanks for the code snippet :)

MikeGost:
I think removing the filter functionality may lead to other types of issues. There's a lot of times when I'm working with a directory that contains non-Atlas files (think success/failure output from our Spark jobs). If the `AtlasResourceLoader` would throw an exception for this use case, it could get fairly cumbersome to use the functionality.

lucaspcram:
Ah I see. That makes sense.

lucaspcram:
So basically it provides a convenient way to load up a `MultiAtlas`, effectively by specifying a whole directory of atlas files, and then it just skips anything in that directory that is not an atlas?

MikeGost:
That's right!

lucaspcram:
Cool, now I am understanding. In that case, having the `withAtlasFileExtensionFilterSetTo(false)` is the best we can do I suppose. You can always reuse the same loader to avoid having to call that long method over and over. This code below seems reasonable.

```java
AtlasResourceLoader myLoader = new AtlasResourceLoader().withAtlasFileExtensionFilterSetTo(false);

Atlas atlas1 = myLoader.load(resource1);
Atlas atlas2 = myLoader.load(resource2);
Atlas atlas3 = myLoader.load(resource3);
```

src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
lucaspcram: Ah ok this is the syntax. I couldn't figure out how to make the compiler stop complaining and got frustrated so I just made a class :P

lucaspcram: Also slight addendum
`private final Predicate<Resource> alwaysTrueAtlasFilter = resource -> return true;` actually doesn't seem to be working. It looks like it needs to be
`private final Predicate<Resource> alwaysTrueAtlasFilter = resource -> {return true; };`

MikeGost:
Agreed, I think the current solution is an optimal middle ground. It may be worthwhile putting some of the content we covered in this PR discussion into the javadoc, to help users properly navigate which options to toggle.

src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java
MikeGost: No IDE to help me on github :) - glad you were able to make it work!

PR #195, opened 2018-08-14
AtlasDelta and AtlasDeltaGenerator Revamp

hallahan:
### Description:

This PR re-implements AtlasDeltaGenerator, as the old implementation was not working. Now, this tool will generate GeoJSON both for the features as well as the relations. It will also output the text-based "diff view friendly string".

### Potential Impact:

AtlasDelta is mainly used for testing and data verification. 

### Unit Test Approach:

Because this PR is focused on the GeoJSON generation, I have `AtlasDeltaGeoJsonIntegrationTest` generate GeoJSON and assert certain things about it such as:
* length of serialized JSON
* feature count
* if the features have the necessary diff related properties

### Test Results:

In addition, I have been running `AtlasDeltaGenerator` on various atlases (just shards), and then take the generated GeoJSON and look at it in JOSM.

Note that I'm continuing with the existing approach where we separate relations from nodes and ways in the delta GeoJSONs. This helps reduce noise, because typically when a relation is modified, a lot of underlying geometry is affected.

PR #199, opened 2018-08-16
Atlas Delta Generator Sharded

hallahan:
### Description:

This is an extension on PR #195 . 

Now the tool allows you to process diffs of directories of atlas shards, instead of single-file atlases. You can now delta before atlases that are sharded with non-sharded afters. You can also, more importantly, delta shard-by-shard before and after sharded atlases. Lastly, single atlas compares also still work. This delineation will be automatic based on if your before / after CLI args are files or directories.

### Potential Impact:

This is non-impacting PR in that I am only improving the AtlasDeltaGenerator CLI, so even AtlasDelta should remain the same.

### Unit Test Approach:

It's a command line tool, so my testing has been manual.

### Test Results:

I have manually tested several countries' worth of data, comparing various atlases, both sharded and non-sharded.


src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -1,28 +1,44 @@
...
+ * This CLI will allow you to point to two atlases and generate a diff of the two. You can point to
+ * single atlas files or directories of atlas shards. If you are diffing shards for both your input
+ * and your output, we will process diffs for each shard individually in parallel.
+ *
  * @author matthieun
  * @author hallahan
  */
 public class AtlasDeltaGenerator extends Command
 {
+    private static final int THREADS = 8; // Tweak this if desired.
```
mgcuthbert: I would make this configurable through a switch, and then just make this the default value. Why did you choose 8 by the way?

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -55,44 +71,118 @@ protected SwitchList switches()
...
+     *
+     * @param path
+     *            An atlas shard directory or a single atlas.
+     * @return An atlas object.
+     * @throws IOException
+     *             Exception if loading the atlas directory does not work.
+     */
+    private Atlas load(final Path path)
+    {
+        return Files.isDirectory(path) ? loadAtlasDirectory(path) : loadSingleAtlas(path);
```
mgcuthbert: nit: "this." before local function calls.

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
hallahan: Basically, that's what my laptop has, and the amount of memory used is not too bad. This is so when you're on a beefy server with many cores, you don't run out of memory. I can make this as a switch, good idea.

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
hallahan: Right, will do.

hallahan:
Anybody know what's up with Travis?

```
The command ".travis/build.sh" exited with 1.
0.00s$ .travis/build-pyatlas-gate.sh
```

lucaspcram:
Huzzah for command line tools! I'll take a look at this tomorrow!

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -44,55 +70,134 @@ protected int onRun(final CommandMap command)
...
+            // You need to have the before dir be a dir of shards too for this to work.
+            if (!Files.isDirectory(before))
+            {
+                logger.error("Your -before parameter must point to a directory of atlas shards if "
+                        + "you want to compare shard by shard with an -after directory also of shards!");
+                System.exit(COMMAND_LINE_USAGE_ERROR_EXIT);
+            }
+
+            // Execute in a pool of threads so we limit how many atlases get loaded in parallel.
+            final ForkJoinPool customThreadPool = new ForkJoinPool(threads);
```
MikeGost: Why not use the `Pool` class from Atlas?

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -44,55 +70,134 @@ protected int onRun(final CommandMap command)
...
+                        .submit(() -> this.compareShardByShard(before, after, outputDirectory))
+                        .get();
+            }
+            catch (final InterruptedException interrupt)
+            {
+                logger.error("The shard diff workers were interrupted.", interrupt);
+            }
+            catch (final ExecutionException exec)
+            {
+                logger.error("There was an execution exception on the shard diff workers.", exec);
```
MikeGost: `customThreadPool` is never shutdown, recommend doing it in a finally block. Possibly a try-with-resources block.

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -44,55 +70,134 @@ protected int onRun(final CommandMap command)
...
+            {
+                customThreadPool
+                        .submit(() -> this.compareShardByShard(before, after, outputDirectory))
+                        .get();
+            }
+            catch (final InterruptedException interrupt)
+            {
+                logger.error("The shard diff workers were interrupted.", interrupt);
+            }
+            catch (final ExecutionException exec)
```
MikeGost: Nit: exec --> execution

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -44,55 +70,134 @@ protected int onRun(final CommandMap command)
...
-
-    private void run(final Path before, final Path after, final Path outputDir)
+    private static List<File> fetchAtlasFilesInDirectory(final Path directory)
     {
-        this.logger.info("Comparing {} and {}", before, after);
-        final Atlas beforeAtlas = load(before);
-        final Atlas afterAtlas = load(after);
-        compare(beforeAtlas, afterAtlas, outputDir);
+        return new File(directory.toFile()).listFilesRecursively().stream()
+                .filter(file -> "atlas".equals(FilenameUtils.getExtension(file.getName())))
```
MikeGost: Recommend reusing the `AtlasResourceLoader` `IS_ATLAS` predicate here, this will also allow for gzipped files to be compared. [Reference](https://github.com/osmlab/atlas/blob/053eed9d3520b5791faafbd8af1aeb9ca33806ca/src/main/java/org/openstreetmap/atlas/geography/atlas/AtlasResourceLoader.java#L57).

If that doesn't work, let's rely on `FileSuffix` instead of a raw string.

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
```diff
@@ -44,55 +70,134 @@ protected int onRun(final CommandMap command)
...
-    private void compare(final Atlas beforeAtlas, final Atlas afterAtlas, final Path outputDir)
+    private Atlas loadSingleAtlas(final Path path)
     {
-        final String name = beforeAtlas.getName().split(".atlas")[0];
+        return new AtlasResourceLoader().load(new File(path.toFile()));
+    }
+
+    private Atlas loadAtlasDirectory(final Path path)
+    {
+        return new AtlasResourceLoader().load(fetchAtlasFilesInDirectory(path));
```
MikeGost: It almost feels like this functionality should be part of `AtlasResourceLoader` : given a path, load the file or files there. Seems very reusable and common logic. If others agree, it may be worthwhile to add it there.

lucaspcram: I am fairly certain `AtlasResourceLoader` already does this. Do:

```java
new AtlasResourceLoader().load(new File(path.toAbsolutePath().toString()))
```

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
hallahan: Well, this is easy and built into Java? It works.

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
hallahan: Cool, will try it out!

src/main/java/org/openstreetmap/atlas/geography/atlas/delta/AtlasDeltaGenerator.java
MikeGost: Yep, remember writing this haha! Looks like we can get rid of some duplicated code here.

hallahan:
Thank you for your suggestions @MikeGost . Ready for review or merge.

PR #203, opened 2018-08-24
Fixing Way Section Exceeding Limit Bug

MikeGost:
### Description:

Couple of changes here:

1. A test case which reproduces the issue. The issue here is in the rare corner case that way-sectioning may run out of identifiers when creating `Edge`s from an OSM Way. This can happen in a couple of cases - when a really (emphasis on really) long way has more than 1000 intersections or (more-likely) in a case of bad data, when a really long way (with more than 1000 shape points) is duplicated and overlayed on top of itself - resulting in sectioning at each shape point - causing us to run out of identifiers. The latter case is what the test case is based off of.
2. The fix in the section logic to address the case. The basic premise is we are going to way-section as usual up to `Edge` 998.  As soon as we see that we only have a single identifier left, we will create an `Edge` 999 with the rest of the un-sectioned way and add a `SyntheticInvalidWaySectionTag` to that `Edge` to indicate it isn't properly way-sectioned. This is done in case we want to add an atlas-check to catch these types of issues in the future.
3. Minor update to way-sectioning logging to give context for fetch and sub-atlas operations when building a `DynamicAtlas`. Previously, we would only log the shard we were starting from in all logging statements, which led to duplicated logging when processing multiple shards. After the change, we have more granularity and know which shard we're starting from and what other shards we're working with.

### Potential Impact:

This will no longer break way-sectioning for bad data. However, it will produce un-way-sectioned `Edge`s for any downstream consumers.

### Unit Test Approach:

Added an explicit unit test that broke prior to the fix.

### Test Results:

Ran a full suite of integration tests and local testing for edge case analysis - all passed.

------


MikeGost:
Good point, let me see if I can trim that down to only the bare minimum!

PR #209, opened 2018-08-30
Find By Atlas Identifier Sub Command

Bentleysb:
### Description:
This adds an `AtlasFindByAtlasIdentifierSubCommand`. The command has two required and one optional parameter. Required are an input directory of Atlas shards, and a list of Atlas identifiers that are to be found. Optional is a output file. This command will output information to the console about each Atlas identifier found. This information includes the item type, containing shard, and tags. Additionally, the shards that contain the matched Atlas identifiers will be joined and saved if an output file is specified. 

This command is based on the `AtlasFindByFeatureIdentifierLocatorSubCommand`, but differs in 3 mayor ways. First, the new command takes Atlas identifiers, instead of feature identifiers. This means that you need not know an entities type to search for it. Second, the feature identifier search only uses the OSM part of the numeric identifier. The Atlas identifier search uses all of the identifier, allowing for specific edges to be searched for (as opposed to all edges with that OSM id). Finally, the new command has an option to merge the shards where matching entities were found and output them as a single atlas. This is very useful when analyzing large areas that have been sharded, and that otherwise would be un-viewable due to memory constraints. 

To accommodate the combining of the matched shards, the `AtlasJoinerSubCommand` is updated to have an optional parameter where shard names can be specified. When present, the shard names are used as a filter whereby only those shards from the specified input directory are joined.  

### Potential Impact:
Adds a useful command.
The `AtlasJoinerSubCommand` update is fully backwards compatible. 

### Unit Test Approach:
The `AtlasFindByAtlasIdentifierSubCommand` has its console output tested by capturing the output stream (using an new helper class) and searching for the expected Atlas shard names. 

The original function of `AtlasJoinerSubCommand` and the new use in combination with `AtlasFindByAtlasIdentifierSubCommand` are tested by joining 3 and 2 tiny Atlas shards, respectively. The results are saved to temporary files and checked for any contents. 

### Test Results:
Verification of search and joining actions were verified using known shard names and visual inspection of new Atlas files in JOSM. All test results were positive. 

------


Bentleysb:
Closing and reopening to rerun Travis.

src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommand.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    @Override
+    protected int finish(final CommandMap command)
+    {
+        final Optional output = command.getOption(JOINED_OUTPUT_PARAMETER);
+        // If joining is requested and there are shards to join...
+        if (output.isPresent() && !this.shardNames.isEmpty())
+        {
+            System.out.printf("Joining...");
```
MikeGost: Nit: Maybe something more descriptive like "Stitching shards and saving to output {}"

src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommand.java
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    /**
+     * Predicate to check an {@link AtlasObject} against the list of ids.
+     *
+     * @param <T>
+     *            Object type
+     * @return {@link Predicate}
+     */
+    private <T extends AtlasObject> Predicate<T> idCheck()
```
MikeGost: Nit: id --> identifier to stay consistent with lack of abbreviations

src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommand.java
```diff
@@ -0,0 +1,125 @@
...
+    {
+        return object -> this.ids.contains(((Long) object.getIdentifier()).toString());
```
MikeGost: Why do we need to cast as `Long` and do a `toString()`? Why not just see if the `Set` of type `Long` contains the object identifier?

src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommand.java
```diff
@@ -0,0 +1,125 @@
...
+     * Creates an informative {@link String} for an atlas {@link AtlasEntity} and the {@link Atlas}
+     * that contains it.
+     *
+     * @param atlas
+     *            Containing {@link Atlas}
+     * @param entity
+     *            {@link AtlasEntity} to create the string for
+     * @return formatted string
+     */
+    private String formatAtlasObject(final Atlas atlas, final AtlasEntity entity)
```
MikeGost: We might be able to get away without passing in the `Atlas`, I believe every `AtlasEntity` should have a `getAtlas() `method, which will return the `Atlas` it belongs to.

src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommandTest.java
```diff
@@ -0,0 +1,54 @@
...
+        final CaptureOutputStream captureStream = new CaptureOutputStream(originalOut);
+        System.setOut(captureStream);
+
+        // Run AtlasFindByAtlasIdentifierSubCommand
+        final String[] args = { "find-atlas-id", String.format("-input=%1$s", shardPath),
+                "-id=546649246000001,575954012000000" };
+        new AtlasReader(args).runWithoutQuitting(args);
+
+        // Reset System.out
+        System.setOut(originalOut);
```
MikeGost: Any chance something goes wrong on lines 29-31 and we never reset System.out? Possible solution could be a try/catch/finally block.

Bentleysb:
@matthieun, unfortunately I don't think text atlases can be used here. `AtlasFindByAtlasIdentifierSubCommand` extends `AbstractAtlasSubCommand` and uses its `execute` method to load the atlas files from the `-input` flag. This method calls `AtlasResourceLoader().load()` to load the atlas files, and this only takes binary atlas files. 
I would just recreate the `execute` method in the test code and change it to accept a text atlas, but all the methods it calls are protected. If you have any ideas about how to get around this (without changing `AbstractAtlasSubCommand` and all of its sub classes) that would be great, because even making those tiny binary atlases was a bit of a pain.

matthieun:
The issue here is we cannot have tests rely on the current binary representation of Atlas files. Once default Atlas binary relies on protobuf arrays for example, those tests would fail. I see the problem you have here, those Atlas files need to be binary in a folder. What I'd suggest is still store them as text in Github, but have your test create a temporary folder in which to save the binary version from text, then run your test, and finally do a cleanup. Does that sound ok?

Bentleysb:
That makes sense and sounds like a plan. I will make that change.

src/test/java/org/openstreetmap/atlas/geography/atlas/command/AtlasJoinerSubCommandTest.java
```diff
@@ -0,0 +1,53 @@
...
+    public static void createBinaryAtlases()
+    {
+        final List<String> shardList = Arrays.asList(TXT_PATH + "/DNK_1.txt",
+                TXT_PATH + "/DNK_2.txt", TXT_PATH + "/DNK_3.txt");
+        shardList.forEach(shard -> new TextAtlasBuilder().read(new File(shard))
+                .save(new File(SHARD_PATH.getPath() + shard.replace("txt", "atlas"))));
+    }
+
+    @AfterClass
+    public static void deleteBianaryAtlases()
```
MikeGost: Nit: Bianary --> Binary

src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasFindByAtlasIdentifierSubCommand.java
```diff
@@ -0,0 +1,124 @@
...
+            possibleMultipleOSMIdentifier -> Stream.of(possibleMultipleOSMIdentifier.split(","))
+                    .map(Long::parseLong).collect(Collectors.toSet()),
+            Command.Optionality.REQUIRED);
+
+    private static final Command.Switch<String> JOINED_OUTPUT_PARAMETER = new Command.Switch<>(
+            "joinedOutput",
+            "The Atlas file to save the joined output to (optional). If not passed the found shards will not be joined and only appear in the console.",
+            String::toString, Command.Optionality.OPTIONAL);
+
+    private final Set<Long> ids = new HashSet<>();
```
MikeGost: Nit: ids --> identifiers, to follow repo convention of no abbreviations.

matthieun:
Owww, the binary Atlas files stayed in the PR :( Will have to remove them later.

MikeGost:
That was my mistake @matthieun - saw the txt files commit, but didn't check if the binary files were removed. Looks like @Bentleysb has followed up with a PR to remove them here: https://github.com/osmlab/atlas/pull/222. Thanks!

PR #210, opened 2018-08-31
WaySectionProcessor performance improvements

matthieun:
### Description:

- When the WaySectionProcessor expands to neighboring shards, using subAtlas was being slow for a minor gain in memory improvement. This PR disables it.
- Fixes an issue when after way sectioning, the data was not always cut along the shard borders, sometimes resulting extra load time
- Changed the dynamic atlas policy's expansion filter to remove a spatial index query on every point on the slicedRawAtlas files.

### Potential Impact:

- Slightly larger memory usage with WaySectionProcessor
- Faster way sectioning

### Unit Test Approach:

Added a new unit test for ensuring of the cut at the border.

image

All other tests pass.

### Test Results:

The proper number of edges are excluded in the new unit test.

image

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
@@ -513,22 +513,16 @@ public class WaySectionProcessor
     {
         try
         {
-            if (this.loadedShards.size() > 1)
-            {
-                // The first shard is always the initial one. Use its bounds to build the atlas.
-                final Rectangle originalShardBounds = this.loadedShards.get(0).bounds();
-                return atlas.subAtlas(originalShardBounds).get();
-            }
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -742,9 +749,16 @@ private boolean isAtlasNode(final WaySectionChangeSet changeSet, final Point poi
     private boolean isAtlasPoint(final WaySectionChangeSet changeSet, final Point point)
     {
         final boolean hasExplicitOsmTags = pointHasExplicitOsmTags(point);
+
+        // We use the presence of explicit OSM tagging to determine if it's a point
+        if (hasExplicitOsmTags)
```
MikeGost: Nice cleanup, this makes more sense logically and likely a performance boost.

PR #212, opened 2018-09-04
Fix single nested relations issue in RawAtlasGenerator

matthieun:
### Description:

In case a relation (A) has a single member **and** that member is a point that is exactly duplicated by another point, **and** that relation's (A) point is the one marked for "duplicate removal" by the `RawAtlasGenerator`, the `PackedAtlasBuilder` would throw an `AtlasIntegrityException` complaining that referencing the relation (A) (which is removed) as a member of relation (B) was illegal. This PR keeps track of this case by making sure relation (A) is also removed as a member from relation (B).

### Potential Impact:

No `AtlasIntegrityException` in that use case.

### Unit Test Approach:

Created a fake PBF file with 2 overlapping nodes, both only members single relations which are only members of another relation each. Checking that the resulting RawAtlas contains only 2 relations and not 4, and that no `AtlasIntegrityException` is thrown.

Also added `.osm` file for reference.

### Test Results:

There are only two relations, and no `AtlasIntegrityException` is thrown.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/RawAtlasGenerator.java
```diff
@@ -444,8 +455,9 @@ private Atlas rebuildAtlas(final Atlas atlas, final Set<Long> pointsToRemove,
             }
             else
             {
-                logger.debug("Relation {} bean is empty, dropping from Atlas",
-                        relation.getIdentifier());
+                final long relationIdentifier = relation.getIdentifier();
+                logger.debug("Relation {} bean is empty, dropping from Atlas", relationIdentifier);
```
jwpgage: bean?

matthieun: RelationBean is an internal PackedAtlas term referring to the relation's collection of members.

PR #213, opened 2018-09-06
Double precision in Distance

matthieun:
### Description:

Underpin `Distance` with a `double` instead of a `long` when storing the value as millimeters.

### Potential Impact:

This changes the `asMillimeters` API to return a `double`. This would allow distance calculations to not lose precision due to rounding to the closest millimeter.

### Unit Test Approach:

Updated unit tests and also ran all integration tests. Also tried generating Atlas files using `AtlasGenerator` in https://github.com/osmlab/atlas-generator and those look good.

### Test Results:

Atlas files look good and all unit tests and integration tests pass.

------


jklamer:
👍

PR #215, opened 2018-09-12
Fixed a resource leak in PackedAtlasSerializer

lucaspcram:
### Description:

Previously, the `determineAtlasLoadFormat` method would cause a resource leak when it failed to get the format correct on the first try. This was because `determineAtlasLoadFormat` tried reading the atlas metadata, but then failed fast and did not close the resource it used to read that metadata. This has been remedied by adding a try-with-resources to the `deserializeSingleField` method, which gets called in all cases of metadata deserialization. The resource will now be closed properly even on failures.

### Potential Impact:

This has no downstream performance or API impacts.

### Unit Test Approach:

The test `testResourceClosureOnInvalidLoadFormat` has been added to `PackedAtlasSerializerTest`. It confirms that the correct number of resource closures occurs.

### Test Results:

Unit test now passes.

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializerTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializerTest.java
@@ -1,5 +1,8 @@
 package org.openstreetmap.atlas.geography.atlas.packed;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.OutputStream;
 import java.lang.reflect.Field;
 import java.util.Map;
 import java.util.Set;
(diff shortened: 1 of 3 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializerTest.java
```diff
@@ -31,6 +41,116 @@
  */
 public class PackedAtlasSerializerTest
 {
+    /**
+     * @author lcram
+     */
+    private static class TraceableByteArrayResource extends AbstractWritableResource
+    {
+        private static final Logger logger = LoggerFactory.getLogger(ByteArrayResource.class);
```
MikeGost: Should this be `TraceableByteArrayResource.class` here?

src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasSerializerTest.java
```diff
@@ -31,6 +41,116 @@
...
+        private static final int BYTE_MASK = 0xFF;
+
+        private int numberStreamsClosed = 0;
+
+        private final ByteArray array;
+
+        TraceableByteArrayResource()
+        {
+            this.array = new ByteArray(Long.MAX_VALUE);
+            this.array.setName("ByteArrayResource");
```
MikeGost: Same here.

PR #216, opened 2018-09-13
Fixed a bug in DynamicRelation.members() that broke BareAtlas.relationsLowerOrderFirst()

lucaspcram:
### Description:

Previously, the interaction between `BareAtlas`, `AtlasEntity`, and `DynamicAtlas`/`DynamicRelation` was broken in the following way:

If you had a `DynamicAtlas` containing some relation `R` that contained at least one other relation `R'` as a member, calling `DynamicAtlas.relationsLowerOrderFirst()` (which used the implementation in `BareAtlas`) would drop relation `R` from the returned set.

This was due to a faulty `AtlasEntity` equality assumption made by `relationsLowerOrderFirst()` which did not hold in the case of `DynamicAtlas`.

This has now been fixed in `DynamicRelation`. The `members` function now returns entities of the correct type, so that the equality assumptions made in `relationsLowerOrderFirst()` are correct.

### Potential Impact:

The broken version of `DynamicAtlas.members()` (and `BareAtlas.relationsLowerOrderFirst()`) was negatively impacting the functionality of `WaySectionProcessor` (the raw atlas version), which depends on `DynamicAtlas` and proper functionality of this method to work. Before the bug fix, the `WaySectionProcessor` would drop any relation which had at least one relation as a member.

### Unit Test Approach:

I have added `DynamicAtlasTest.testRelationsLowerOrderFirstConsistency()`, which demonstrates the case in which the bug would occur. It utilizes a new test atlas, `DynamicAtlasTestRule.atlasForRelationsTest`, which is designed to trigger the bug. With the fix, this unit test now passes.

`RawAtlasIntegrationTest.testSectioningFromShard()` was also affected. In the previous code version, it was dropping two relations from the build, and the test was assuming that those two relations did not exist (when in reality they had actually just been dropped by the `WaySectionProcessor`). The test is updated, and passes with the bug fix in place.

### Test Results:

I am planning on doing more extensive testing. Please reach out to me to discuss.

------

### DIFF
--- a/src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/RawAtlasIntegrationTest.java
+++ b/src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/RawAtlasIntegrationTest.java
@@ -287,7 +287,15 @@ public class RawAtlasIntegrationTest
         Assert.assertEquals(5126, finalAtlas.numberOfAreas());
         Assert.assertEquals(184, finalAtlas.numberOfPoints());
         Assert.assertEquals(271, finalAtlas.numberOfLines());
-        Assert.assertEquals(14, finalAtlas.numberOfRelations());
+
+        /*
+         * This has been updated to 16 instead of 14. This is because two of the relations in
+         * 8-122-122-trimmed.osm.pbf have subrelations, and so the WaySectionProcessor was simply
+         * dropping them when using the old BareAtlas.relationsLowerOrderFirstCode (We have now
+         * fixed BareAtlas.relationsLowerOrderFirst to not drop relations when the BareAtlas is a
+         * DynamicAtlas). In reality, we should process them and include then in the final atlas.
+         */
+        Assert.assertEquals(16, finalAtlas.numberOfRelations());
     }
 
     @Test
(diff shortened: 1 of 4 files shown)


src/integrationTest/java/org/openstreetmap/atlas/geography/atlas/raw/RawAtlasIntegrationTest.java
```diff
@@ -287,7 +287,15 @@ public void testSectioningFromShard()
         Assert.assertEquals(5126, finalAtlas.numberOfAreas());
         Assert.assertEquals(184, finalAtlas.numberOfPoints());
         Assert.assertEquals(271, finalAtlas.numberOfLines());
-        Assert.assertEquals(14, finalAtlas.numberOfRelations());
+
+        /*
+         * This has been updated to 16 instead of 14. This is because two of the relations in
```
MikeGost: Not sure this comment is necessary. The check here was on the data, simply changing the number is fine. Thanks for the detailed explanation!

lucaspcram: I honestly left it just for the purposes of explaining the change in the PR, so you guys didn't think I was just cheating to make a test pass :)

It could have been removed but we merged so fast that I didn't get a chance.

PR #219, opened 2018-09-19
Initialize in Construction of AtlasEntityPolygonsFilter

jklamer:
### Description:
Initialize the underlying STRtree in the Rtree class, which lazily constructs its index, when first queried. 
This means without querying the Rtree during construction, the filter isn't thread safe despite being read only. 

### Potential Impact:

Slightly higher memory usage when the filter is never used. 

### Unit Test Approach:

Standard unit tests should cover because the problem is itself a race condition and cannot be reliably reproduced.

### Test Results:

Passes all unit tests.

------


jklamer:
@MikeGost multiple threads using the same filter. Caused an NPE in the STRtree code, where the stack trace lead right to the issue.

Issue #88, opened 2018-02-24
Simple Command Line Interface for PBF -> atlas


hallahan:
Sometimes, all I need to do is make an atlas file with an OSM PBF file. Right now, I write some Java code like so:

```java
final File pbfFile = new File(pbf);
final OsmPbfLoader loader = new OsmPbfLoader(pbfFile, MultiPolygon.MAXIMUM,
                AtlasLoadingOption.withNoFilter());
final Atlas atlas = loader.read();
final File atlasFile = new File("output.atlas");
atlas.save(atlasFile);
```

Instead, it would be nice to just do this from the command line and not have to write any code...

This might be already there. If so, I'm happy to add instructions to the readme.

matthieun:
@hallahan Would a pre-baked script do the trick?

hallahan:
Basically, a simple CLI where you can give it an input PBF and a boundary file, and out comes atlas shards.

lucaspcram:
Closing this as it is addressed by command `pbf2atlas` added in #223

PR #223, opened 2018-09-24
PBF To Atlas Sub Commmand

Bentleysb:
### Description:
This adds a simple command to convert an OSM PBF to an Atlas file (as per https://github.com/osmlab/atlas/issues/88). It requires only the path to a PBF file, and a path for an output file. It has a number of optional parameters that can be passed to customize the Atlas generation. 

### Potential Impact:

No impacts for downstream libraries. 

### Unit Test Approach:
Unit test `testDefaultConversion` checks that that the command functions as intended when no optional parameters are passed.  
Unit test `testFiltersTextMapCountryCodesConversion` checks that all the filter parameters, country code parameter, and the country boundary map parameter with a text file work as expected.  
Unit test `testNoSectioningNoSlicingNoRelationsShapeMapConversion` checks that the way section, country slicing, load relation, and country boundary map (with a shape file) parameters work as expected.  
Unit test `testNoWaysConversion` checks that the load ways parameter works as expected. 

### Test Results:
Local tests were performed on multiple PBFs, using many combinations of default and optional parameters. All worked as expected, with the exception of using a large shape file as a country boundary map. This scenario caused the command to take a long time to load.  It is recommended to use text country boundary maps when possible. 

------


src/main/java/org/openstreetmap/atlas/geography/atlas/command/OsmPbfToAtlasSubCommand.java
```diff
@@ -0,0 +1,207 @@
...
+            "Path to a json filter for OSM relations", File::new, Optionality.OPTIONAL);
+    private static final Switch<File> WAY_FILTER_PARAMETER = new Switch<>("way-filter",
+            "Path to a json filter for OSM ways", File::new, Optionality.OPTIONAL);
+    private static final Switch<File> WAY_SECTION_FILTER_PARAMETER = new Switch<>(
+            "way-section-filter", "Path to a json filter for determining where to way section",
+            File::new, Optionality.OPTIONAL);
+
+    // Load Parameters
+    private static final Switch<Boolean> LOAD_RELATIONS_PARAMETER = new Switch<>("load-relations",
+            "Whether to load Relations (boolean)", Boolean::new, Optionality.OPTIONAL);
```
matthieun: How about `Boolean::parseBoolean` and add a default value of `"true"` at the end? That would make the switches well defined for the end user!

src/main/java/org/openstreetmap/atlas/geography/atlas/command/OsmPbfToAtlasSubCommand.java
```diff
@@ -0,0 +1,207 @@
...
+
+    // Required parameters
+    private static final Switch<File> INPUT_PARAMETER = new Switch<>("pbf", "Input PBF path",
+            File::new, Optionality.REQUIRED);
+    private static final Switch<File> OUTPUT_PARAMETER = new Switch<>("output",
+            "Output Atlas file path", File::new, Optionality.REQUIRED);
+
+    // Filter parameters
+    private static final Switch<File> EDGE_FILTER_PARAMETER = new Switch<>("edge-filter",
+            "Path to a json filter for determining Edges", File::new, Optionality.OPTIONAL);
```
matthieun: I would mention that it has to be a local File here in the description.

src/main/java/org/openstreetmap/atlas/geography/atlas/command/OsmPbfToAtlasSubCommand.java
```diff
@@ -0,0 +1,207 @@
...
+        writer.println(
+                "-way-sectioning=boolean : whether to perform way sectioning; defaults to true");
+        writer.println(
+                "-way-section-filter=/path/to/json/way/section/filter : json filter to determine where to way section");
+    }
+
+    @Override
+    public int execute(final CommandMap map)
+    {
+        new OsmPbfLoader((File) map.get(INPUT_PARAMETER), this.getAtlasLoadingOption(map))
```
matthieun: Here, I think we should use the latest ingest, that goes through raw Atlas. I think this is the way (@MikeGost can correct me):
```java
File pbf, output;
AtlasLoadingOption option;

Atlas atlas = new RawAtlasGenerator(pbf, option).build();
if (countrySlicing)
{
    atlas = new RawAtlasCountrySlicer(countryList, countryBoundaries).slice(atlas);
}
atlas = new WaySectionProcessor(atlas, option).run();
atlas.save(output);
```
@MikeGost Maybe we should have an easy one-liner available that does all that somewhere, when the pbf->atlas loading use case is simple like that.

src/test/resources/org/openstreetmap/atlas/geography/atlas/command/continent_map.cpg
```diff
@@ -0,0 +1 @@
+UTF-8
```
matthieun: Can we remove all the shapefiles? I think the text boundary below should be enough..

src/main/java/org/openstreetmap/atlas/geography/atlas/command/OsmPbfToAtlasSubCommand.java
MikeGost: Agreed, this should flow through the raw atlas flow instead of the legacy atlas creation mechanism. This is also strikingly similar to the work that @lucaspcram did here: https://github.com/osmlab/atlas-generator/pull/53. We may want to consolidate the tools into a single one.

MikeGost: See `RawAtlasIntegrationTest` for some examples of the flow.

MikeGost: @matthieun - I have a backlog item to refactor AtlasLoadingOption that would automatically allow you to build complete atlases, without having to perform each step (create, slice, section) yourself. Definitely valuable functionality that can be reused here.

Bentleysb: It looks easy enough to change this over to the raw atlas flow. The bigger questions seems to be if we want to consolidate commands. I did not realize the atlas generator command existed, and I agree it has very similar functionality. The one big difference I see is that the atlas generator command is built around the sharded pbf/atlas workflow, whereas this command is made to be a more general flow.

PR #225, opened 2018-09-24
File extension is now preserved by the temp file caching strategy

lucaspcram:
### Description:

The file extension of a cached resource is now preserved by the temporary file strategy. This makes [the changes in this PR unnecessary.](https://github.com/osmlab/atlas/pull/198)

### Potential Impact:

This will technically impact any users of `SystemTemporaryFileCachingStrategy`. But the impact is all internal and will go unnoticed by users of the class.

### Unit Test Approach:

Added some additional testing to `CachingTests.testBaseCacheWithGivenStrategy` to make sure the file extension saving is working.

### Test Results:

Test results passed. Additionally, I manually verified that file extensions are saved correctly by checking my local tmp file system.

------


src/test/java/org/openstreetmap/atlas/utilities/caching/CachingTests.java
```diff
@@ -78,6 +80,10 @@ private void compareFetchedResource(final URI resourceURI, final Resource golden
             .get("src/test/resources/org/openstreetmap/atlas/utilities/configuration/feature.json")
             .toAbsolutePath();
     private static final URI LOCAL_TEST_FILE_2_URI = LOCAL_TEST_FILE_2.toUri();
+    private static final Path LOCAL_TEST_FILE_3 = Paths
+            .get("src/test/resources/org/openstreetmap/atlas/utilities/caching/fileNoExt")
```
matthieun: It might not be the right place to say this, but, it would be better to not have hardcoded paths there (in case we change package names in the future, this _will_ be missed).

How about doing this in the tests below:
```
ByteArrayResource originalFileBytes = new ByteArrayResource();
originalFileBytes.copyFrom(new InputStreamresource(() ->
    CachingTests.class.getResourceAsStream("fileNoExt")));
```

lucaspcram: This is a good call. I think there are a lot of places in the codebase that could use this change.

src/test/java/org/openstreetmap/atlas/utilities/caching/CachingTests.java
```diff
@@ -101,85 +49,37 @@ public void testBaseCacheWithNoStrategy()
     public void testLocalFileInMemoryCache()
     {
         final LocalFileInMemoryCache cache = new LocalFileInMemoryCache();
+        URI resourceUri;
+        try
+        {
+            resourceUri = CachingTests.class.getResource(LOG4J_PROPERTIES).toURI();
```
matthieun: Is log4j.properties used as a random file to test the cache, or did you need it for configuring log4j here?

lucaspcram: It's just a random, non-empty file to test the caching. Should I change it? (I assume it may have logging implications.)

PR #226, opened 2018-09-24
Relation Bean from relation

jklamer:
### Description:
Add the ability to get a RelationBean representation from a Relation.

### Potential Impact:

None unless used

### Unit Test Approach:

N/A

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
@@ -36,7 +36,7 @@ public class RelationBean implements Serializable
     @Override
     public boolean equals(final Object other)
     {
-        if (Objects.nonNull(other) && other instanceof RelationBean)
+        if (other instanceof RelationBean)
         {
             final RelationBean that = (RelationBean) other;
             return Iterables.equals(this.getMemberIdentifiers(), that.getMemberIdentifiers())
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -93,6 +94,20 @@ public String configurableString(final String betweenEachMemberAndRelation,
...
 
+    /**
+     * @return the {@link RelationBean} representation of the Relation
+     */
+    public RelationBean getBean()
+    {
+        final RelationBean bean = new RelationBean();
+        for (final RelationMember member : this)
+        {
+            bean.addItem(member.getEntity().getIdentifier(), member.getRole(),
```
MikeGost: Nit: for a large Relation - any performance impact on pulling out the entity just once and saving as a temporary variable, and then getting the identifier and type instead of fetching the entity each time.

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
```diff
@@ -31,6 +33,26 @@ public void addItem(final Long identifier, final String role, final ItemType ite
         this.memberTypes.add(itemType);
     }
 
+    @Override
+    public boolean equals(final Object other)
+    {
+        if (Objects.nonNull(other) && other instanceof RelationBean)
```
MikeGost: Nit: nonNull is not needed - `instanceof` will perform the null check under the covers.

PR #230, opened 2018-10-01
Fixing way sectioning bug in raw atlas

MikeGost:
### Description:

In the case of invalid data (see previous version of https://www.openstreetmap.org/way/621043891) - we were getting a way-sectioning error. The existing logic keeps track of the number of times a given `Location` has been the used as the start `Node` and the number of times that `Location` is duplicated between the start and end `Node`s. This is used to calculate the occurrence of the start and end `Node`s. The bug manifested in the case where a duplicated `Location` was an end `Node` more than a single time. The old logic would take the number of seen occurrences so far and add the number of duplicates to that to calculate the index of the end `Node`. However, when this happened more than once, the count broke down, since we would be adding the number of duplicates more than once. The fix is to instead use the duplicate count plus the number of occurrences seen so far, without doing the incrementation. I've also updated the code to be consistent across both ring and non-ring lines. 

A shoutout to @lucaspcram for providing the test resources used to reproduce the issue, thank you!

### Potential Impact:

More robust way-sectioning for dealing with invalid data.

### Unit Test Approach:

Created a new unit test that was failing pre-fix and working post fix. 

### Test Results:

Existing unit and integration tests are consistent and passing both pre and post fix.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -1220,30 +1213,28 @@ private boolean shouldSectionAtLocation(final Location location)
...
+                            final int startOccurrence = nodesToSectionAt
+                                    .getOccurrence(startNode.get()) - 1;
+                            nodesToSectionAt.incrementOccurrence(startNode.get());
+
+                            // Update end point occurrence to factor in any duplicates in the
+                            // polyline
+                            final int duplicates = duplicateLocations.getOrDefault(currentLocation,
+                                    0);
+                            final int endOccurrence = duplicates
+                                    + nodesToSectionAt.getOccurrence(end) - 1;
```
adahn6: elegant! 💯 given the comment explaining the calculation, and the fact that `duplicates` is disposable (single-use), might be best to just do this calculation in one line? i.e. 
```
final int endOccurrence = duplicateLocations.getOrDefault(currentLocation, 0)
    + nodesToSectionAt.getOccurrence(end) - 1;
```

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
```diff
@@ -1063,23 +1063,16 @@ private boolean shouldSectionAtLocation(final Location location)
...
                     // since PolyLine uses zero-based numbering. We are incrementing only the
                     // start node occurrence, since the end node will either be used as a future
                     // start node or be the end of the way, in which case we don't care.
                     final int startOccurrence = nodesToSectionAt.getOccurrence(startNode.get()) - 1;
                     nodesToSectionAt.incrementOccurrence(startNode.get());
-                    final int endOccurrence = nodesToSectionAt.getOccurrence(end) - 1;
+
+                    // Update end point occurrence to factor in any duplicates in the polyline
+                    final int duplicates = duplicateLocations.getOrDefault(end.getIdentifier(), 0);
+                    final int endOccurrence = duplicates + nodesToSectionAt.getOccurrence(end) - 1;
```
adahn6: same comment as below re: one-line calculation

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
MikeGost: Good point, let me update!

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessorTestRule.java
```diff
@@ -11,6 +11,9 @@
  */
 public class WaySectionProcessorTestRule extends CoreTestRule
 {
+    @TestAtlas(loadFromTextResource = "malformedPolyLine.atlas.txt")
+    private Atlas malformedPolyLineAtlas;
+
```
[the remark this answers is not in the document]
MikeGost: Yeah, it can load from JOSM resource, OSM XML or text. Very powerful! It can be expanded to load from geoJson and other formats as well.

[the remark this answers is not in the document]
matthieun: Here it has to be loaded from a text, as the josm/osm->atlas use the raw atlas loader that we are testing here :)

PR #232, opened 2018-10-03
Initial SQLite Atlas file format

lucaspcram:
### Description:

These are the beginnings of a SQLite atlas file format. Things are still very much WIP, but I would like to get them merged in to encourage collaboration.

### Potential Impact:

No downstream impact.

### Unit Test Approach:

This is an area we still need to flesh out. Testing has been ad-hoc, using the various creation commands found in the PR.

### Test Results:

Results look good so far, but we initially have only implemented the basics.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
@@ -676,6 +676,29 @@ public interface Atlas extends Located, Iterable<AtlasEntity>, Serializable
     void saveAsGeoJson(WritableResource resource, Predicate<AtlasEntity> matcher);
 
     /**
+     * Save as a SQLite database
+     *
+     * @param resource
+     *            The SQLite file database to write to
+     * @param buildIndices
+     *            Whether the database should have the indices built for you.
+     */
+    void saveAsSQLite(WritableResource resource, boolean buildIndices);
+
+    /**
+     * Save as a SQLite database with matcher
+     *
+     * @param resource
+     *            The SQLite file database to write to
+     * @param buildIndices
+     *            * Whether the database should have the indices built for you.
+     * @param matcher
+     *            The matcher to consider
+     */
+    void saveAsSQLite(WritableResource resource, boolean buildIndices,
+            Predicate<AtlasEntity> matcher);
+
+    /**
      * Save as list of items
      *
      * @param resource
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
@@ -1,5 +1,7 @@
 package org.openstreetmap.atlas.geography.atlas.packed;
 
+import java.io.IOException;
+import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -10,6 +12,7 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.TreeSet;
+import java.util.function.Predicate;
 
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.Location;
@@ -32,6 +35,9 @@ import org.openstreetmap.atlas.geography.atlas.items.Point;
 import org.openstreetmap.atlas.geography.atlas.items.Relation;
 import org.openstreetmap.atlas.geography.atlas.items.RelationMember;
 import org.openstreetmap.atlas.geography.atlas.items.RelationMemberList;
+import org.openstreetmap.atlas.geography.atlas.sqlite.PackedAtlasFieldDelegator;
+import org.openstreetmap.atlas.geography.atlas.sqlite.SQLiteWriter;
+import org.openstreetmap.atlas.streaming.resource.File;
 import org.openstreetmap.atlas.streaming.resource.Resource;
 import org.openstreetmap.atlas.streaming.resource.WritableResource;
 import org.openstreetmap.atlas.utilities.arrays.ByteArrayOfArrays;
@@ -685,6 +691,45 @@ public final class PackedAtlas extends AbstractAtlas
     public void save(final WritableResource writableResource)
     {
         new PackedAtlasSerializer(this, writableResource).save();
+    }
+
+    @Override
+    public void saveAsSQLite(final WritableResource resource, final boolean buildIndices,
+            final Predicate<AtlasEntity> matcher)
+    {
+        try
+        {
+            final File file;
+            if (resource instanceof File)
+            {
+                file = (File) resource;
+            }
+            else
+            {
+                file = new File(java.io.File.createTempFile(resource.getName(), ".db"));
+            }
+
+            // TODO see note in the PackedAtlasFieldDelegator class
+            PackedAtlasFieldDelegator delegator = new PackedAtlasFieldDelegator()
+                    .withDictionary(dictionary()).withNodeTags(nodeTags()).withEdgeTags(edgeTags())
+                    .withPointTags(pointTags()).withLineTags(lineTags()).withAreaTags(areaTags());
+            delegator = delegator.withNodeIdentifierMap(nodeIdentifierToNodeArrayIndex())
+                    .withEdgeIdentifierMap(edgeIdentifierToEdgeArrayIndex())
+                    .withPointIdentifierMap(pointIdentifierToPointArrayIndex())
+                    .withLineIdentifierMap(lineIdentifierToLineArrayIndex())
+                    .withAreaIdentifierMap(areaIdentifierToAreaArrayIndex());
+
+            final SQLiteWriter writer = new SQLiteWriter(file, false, delegator);
+            writer.write(this, matcher);
+        }
+        catch (final IOException ioex)
+        {
+            logger.error("Unable to create SQLite database file.", ioex);
+        }
+        catch (final SQLException sqlex)
+        {
+            logger.error("JDBC unable to build Atlas SQLite database.", sqlex);
+        }
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/AtlasDirectorySQLiteConverter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/AtlasDirectorySQLiteConverter.java
@@ -1,0 +1,102 @@
+package org.openstreetmap.atlas.geography.atlas.sqlite;
+
+import static org.openstreetmap.atlas.geography.atlas.AtlasResourceLoader.IS_ATLAS;
+
+import java.io.IOException;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.apache.commons.io.FileUtils;
+import org.openstreetmap.atlas.geography.atlas.Atlas;
+import org.openstreetmap.atlas.geography.atlas.AtlasResourceLoader;
+import org.openstreetmap.atlas.streaming.resource.File;
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
+import org.openstreetmap.atlas.utilities.runtime.Command;
+import org.openstreetmap.atlas.utilities.runtime.CommandMap;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * This CLI is similar to SQLiteConverter, except here you give it a directory to search for
+ * atlases. You also give it a SQLite directory where those atlas shards will then be converted into
+ * their equivalent SQLite database shard.
+ *
+ * @author hallahan
+ */
+public class AtlasDirectorySQLiteConverter extends Command
+{
+
+    private static final Logger logger = LoggerFactory
+            .getLogger(AtlasDirectorySQLiteConverter.class);
+
+    private static final Switch<Path> ATLAS_DIRECTORY = new Switch<>("atlasDirectory",
+            "The directory of atlases to convert.", Paths::get, Optionality.REQUIRED);
+
+    private static final Switch<Path> SQLITE_DIRECTORY = new Switch<>("sqliteDirectory",
+            "The directory to write SQLite databases.", Paths::get, Optionality.REQUIRED);
+
+    private static final Switch<Boolean> OVERWRITE = new Switch<>("overwrite",
+            "Choose to automatically overwrite a database if it exists at the given path.",
+            Boolean::new, Optionality.OPTIONAL, "false");
+
+    private static final Switch<Boolean> BUILD_INDICES = new Switch<>("indices",
+            "Choose to build column and spatial indices automatically.", Boolean::new,
+            Optionality.OPTIONAL, "false");
+
+    public static void main(final String[] args)
+    {
+        new AtlasDirectorySQLiteConverter().run(args);
+    }
+
+    private static List<File> fetchAtlasFilesInDirectory(final Path directory)
+    {
+        return new File(directory.toFile()).listFilesRecursively().stream().filter(IS_ATLAS)
+                .collect(Collectors.toList());
+    }
+
+    @Override
+    protected int onRun(final CommandMap command)
+    {
+        final Path atlasDirectory = (Path) command.get(ATLAS_DIRECTORY);
+        final Path sqliteDirectory = (Path) command.get(SQLITE_DIRECTORY);
+        final Boolean overwrite = (Boolean) command.get(OVERWRITE);
+        final Boolean buildIndices = (Boolean) command.get(BUILD_INDICES);
+
+        if (overwrite)
+        {
+            try
+            {
+                FileUtils.deleteDirectory(sqliteDirectory.toFile());
+            }
+            catch (final IOException noDelete)
+            {
+                logger.warn(
+                        "Tried to delete SQLite database output directory {} for overwrite, but unable.",
+                        sqliteDirectory);
+            }
+        }
+
+        final List<File> atlases = fetchAtlasFilesInDirectory(atlasDirectory);
+        logger.info("About to convert {} atlas shards into SQLite databases...", atlases.size());
+        for (final File atlasFile : atlases)
+        {
+            final Atlas atlas = new AtlasResourceLoader().load(atlasFile);
+            final String name = atlasFile.getName().split(FileSuffix.ATLAS.toString())[0];
+            final String dbName = name + FileSuffix.SQLITE.toString();
+            final File sqliteFile = new File(sqliteDirectory.resolve(dbName).toFile());
+            atlas.saveAsSQLite(sqliteFile, buildIndices);
+            logger.info("Saved Atlas SQLite database shard to {}", sqliteFile);
+        }
+
+        logger.info("Finished converting directory of atlases into SQLite database shards!");
+        return 0;
+    }
+
+    @Override
+    protected SwitchList switches()
+    {
+        return new SwitchList().with(ATLAS_DIRECTORY, SQLITE_DIRECTORY, OVERWRITE, BUILD_INDICES);
+    }
+}
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
@@ -1,0 +1,152 @@
+package org.openstreetmap.atlas.geography.atlas.sqlite;
+
+import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas;
+import org.openstreetmap.atlas.geography.atlas.packed.PackedTagStore;
+import org.openstreetmap.atlas.utilities.compression.IntegerDictionary;
+import org.openstreetmap.atlas.utilities.maps.LongToLongMap;
+
+/**
+ * This delegator allows the {@link SQLiteWriter} to access internal {@link PackedAtlas} fields it
+ * needs to efficiently build the database.
+ *
+ * @author lcram
+ */
+public class PackedAtlasFieldDelegator
+{
+    // TODO this class has a lot of fields. It may be worth refactoring it out and just using
+    // reflection. Alternatively, we could also just make a ton of PackedAtlas fields public. Let's
+    // discuss these options please.
+
+    private IntegerDictionary<String> dictionaryHandle;
+    private PackedTagStore nodeTagsHandle;
+    private PackedTagStore edgeTagsHandle;
+    private PackedTagStore pointTagsHandle;
+    private PackedTagStore lineTagsHandle;
+    private PackedTagStore areaTagsHandle;
+    private LongToLongMap nodeIdentifierMapHandle;
+    private LongToLongMap edgeIdentifierMapHandle;
+    private LongToLongMap pointIdentifierMapHandle;
+    private LongToLongMap lineIdentifierMapHandle;
+    private LongToLongMap areaIdentifierMapHandle;
+
+    public LongToLongMap areaIdentifierMapHandle()
+    {
+        return this.areaIdentifierMapHandle;
+    }
+
+    public PackedTagStore areaTagsHandle()
+    {
+        return this.areaTagsHandle;
+    }
+
+    public IntegerDictionary<String> dictionaryHandle()
+    {
+        return this.dictionaryHandle;
+    }
+
+    public LongToLongMap edgeIdentifierMapHandle()
+    {
+        return this.edgeIdentifierMapHandle;
+    }
+
+    public PackedTagStore edgeTagsHandle()
+    {
+        return this.edgeTagsHandle;
+    }
+
+    public LongToLongMap lineIdentifierMapHandle()
+    {
+        return this.lineIdentifierMapHandle;
+    }
+
+    public PackedTagStore lineTagsHandle()
+    {
+        return this.lineTagsHandle;
+    }
+
+    public LongToLongMap nodeIdentifierMapHandle()
+    {
+        return this.nodeIdentifierMapHandle;
+    }
+
+    public PackedTagStore nodeTagsHandle()
+    {
+        return this.nodeTagsHandle;
+    }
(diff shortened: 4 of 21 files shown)


lucaspcram:
@hallahan The PR is up.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -687,6 +693,45 @@ public void save(final WritableResource writableResource)
...
+            {
+                file = (File) resource;
+            }
+            else
+            {
+                file = new File(java.io.File.createTempFile(resource.getName(), ".db"));
+            }
+
+            // TODO see note in the PackedAtlasFieldDelegator class
+            PackedAtlasFieldDelegator delegator = new PackedAtlasFieldDelegator()
```
hallahan: Nice : )

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
```diff
@@ -0,0 +1,152 @@
...
+import org.openstreetmap.atlas.utilities.compression.IntegerDictionary;
+import org.openstreetmap.atlas.utilities.maps.LongToLongMap;
+
+/**
+ * This delegator allows the {@link SQLiteWriter} to access internal {@link PackedAtlas} fields it
+ * needs to efficiently build the database.
+ *
+ * @author lcram
+ */
+public class PackedAtlasFieldDelegator
```
hallahan: Nit. Typically they say "Delegate", not "Delegator".

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
```diff
@@ -0,0 +1,152 @@
...
+{
+    // TODO this class has a lot of fields. It may be worth refactoring it out and just using
+    // reflection. Alternatively, we could also just make a ton of PackedAtlas fields public. Let's
+    // discuss these options please.
+
```
hallahan: I think the reason the getters are not public in PackedAtlas is that in theory, you could mutate what gets returned. But, we're not... However, making a copy is also a bad idea, because we don't want to abuse our memory usage.

Personally, I'd just make the getters public in packed atlas, but I'm fine with whatever, as long as we don't do a copy.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+ * To save all of the data within an atlas, we need to get each {@link AtlasItem} type and iterate
+ * though them with inserts into the database. This class declares that a given writer needs to get
+ * the atlas to do its job.
+ *
+ * @author hallahan
+ * @author lcram
+ */
+public class SQLiteItemWriter
+{
+    private static final String INSERT_SQL = "INSERT INTO %s VALUES (?, ?, ?, ?)";
```
hallahan: I like this

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+    private static final String INSERT_TAGS_SQL = "INSERT INTO %s_tags VALUES (?, ?, ?)";
+    private static final String INSERT_RTREE_SQL = "INSERT INTO %s_rtree VALUES (?, ?, ?, ?, ?)";
+
+    // database column constants
+    private static final int FIRST_COLUMN = 1;
+    private static final int SECOND_COLUMN = 2;
+    private static final int THIRD_COLUMN = 3;
+    private static final int FOURTH_COLUMN = 4;
+    private static final int FIFTH_COLUMN = 5;
```
hallahan: I used to do that, but it get's a bit hard to maintain eventually. ++column is easier. But, I also am fine with whatever.

lucaspcram: Would it be preferable to just add a checkstyle exception for this class, specifically for the magic constants rule? That way we could just use integer literals. Or do you still think `++column` is easier to maintain? I did find that you have to be careful about resetting it to `0` all over the place.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
lucaspcram: Fair enough. Honestly considering changing it to `bean` or `container` or something, since it seems to be straying a bit from the delegate pattern (since it's not delegating so much as it is simply packaging a bunch of references).

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/PackedAtlasFieldDelegator.java
lucaspcram: Yeah, no copies for sure. If this were C/C++, we would have `const` here to help us out, and all parties would be satisfied...

I'm not particularly committed to any of the particular solutions, but I do sympathize with @matthieun 's desire to keep `PackedAtlas` as locked down as possible.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
hallahan: I have no preference.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
```diff
@@ -114,6 +114,7 @@ public ProtoSerializable deserialize(final byte[] byteArray)
         Field indexField = null;
         Long index = -1L;
 
+        // Consider using getters, says nick
```
hallahan: This is just a note, should we do this in another PR?

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteDictionaryWriter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public SQLiteDictionaryWriter(final Connection connection,
+            final PackedAtlasFieldDelegator delegator) throws SQLException
+    {
+        this.connection = connection;
+        this.insertStatement = this.connection.prepareStatement(INSERT_SQL);
+        this.delegator = delegator;
+    }
+
+    public void write(final PackedAtlas atlas) throws SQLException
```
adahn6: why does this have an unused parameter?

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+        // Performance is very bad without doing this!
+        this.connection.setAutoCommit(false);
+
+        final Iterable<? extends AtlasItem> items = getItemIterableOfCorrectType(atlas,
+                this.itemType);
+
+        for (final AtlasItem item : items)
+        {
+            // INSERT feature data
+            final long atlasId = item.getIdentifier();
```
adahn6: this may not be possible and/or efficient, but I think you could add a `getItemSQL()` to each item class, then you could just maintain it with the data itself instead of in a separate set of switches and generic logic. e.g., inside of Node.java, 
```Java
public PreparedStatement getItemSql(Connection connection, String shardName) {
    String itemStatement = connection.prepareStatement(
                 String.format(SQLiteItemWriter.INSERT_SQL, this.itemType.toString().toLowerCase()));
    itemStatement.setLong(SQLiteItemWriter.FIRST_COLUMN, this.getIdentifier());
    itemStatement.setString(SQLiteItemWriter.SECOND_COLUMN, "md5sum_here");
    itemStatement.setString(SQLiteItemWriter.THIRD_COLUMN, shardName);
    itemStatement.setBytes(SQLiteItemWriter.FOURTH_COLUMN,
        new WkbLocationConverter().convert(this.getLocation()));
    return itemStatement;
}
```
Same thing for tag statement, which could be in AtlasEntity.java. Then, you could just iterate over each atlas entity directly, skipping over unrequested / unsupported types, like 
```
for (final AtlasItem item : entitiesIterator) {
    if( checkEntityTypeSupported(item)) {
        item.getItemSQL(connection, shardName).execute(); 
    } 
}
```
Obvs this is scratch code and just a thought! May not be worth the refactor or even actually work.

src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
```diff
@@ -675,6 +675,29 @@
...
+     */
+    void saveAsSQLite(WritableResource resource, boolean buildIndices);
+
+    /**
+     * Save as a SQLite database with matcher
+     *
+     * @param resource
+     *            The SQLite file database to write to
+     * @param buildIndices
+     *            * Whether the database should have the indices built for you.
```
MikeGost: Nit: extra *

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/AtlasDirectorySQLiteConverter.java
```diff
@@ -0,0 +1,102 @@
...
+
+        if (overwrite)
+        {
+            try
+            {
+                FileUtils.deleteDirectory(sqliteDirectory.toFile());
+            }
+            catch (final IOException noDelete)
+            {
+                logger.warn(
```
MikeGost: Log the `noDelete`, just in case.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/AtlasDirectorySQLiteConverter.java
```diff
@@ -0,0 +1,102 @@
...
+/**
+ * This CLI is similar to SQLiteConverter, except here you give it a directory to search for
+ * atlases. You also give it a SQLite directory where those atlas shards will then be converted into
+ * their equivalent SQLite database shard.
+ *
+ * @author hallahan
+ */
+public class AtlasDirectorySQLiteConverter extends Command
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteDictionaryWriter.java
```diff
@@ -0,0 +1,52 @@
...
+        for (int key = 0; key < size; ++key)
+        {
+            int column = 0;
+            final String value = dictionary.word(key);
+            this.insertStatement.setInt(++column, key);
+            this.insertStatement.setString(++column, value);
+            this.insertStatement.execute();
+        }
+
+        this.connection.commit();
```
MikeGost: Are we leaking connections? This is never closed anywhere. Possibly implement `AutoCloseable` here.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+        if (item.getType() != this.itemType)
+        {
+            throw new CoreException(
+                    "ItemType {} of AtlasItem {} did not match expected ItemType {}",
+                    item.getType(), item, this.itemType);
+        }
+
+        switch (item.getType())
+        {
+            case NODE:
```
MikeGost: Nit: To simplify, consider breaking this down into `LineItem` and `LocationItem` since `Edge`/`Line` processing is the identical, as is `Point`/`Node` processing.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteMergeCommand.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+            logger.info("Took {} to merge the databases.", mergeTime.elapsedSince());
+
+            if (buildIndices)
+            {
+                logger.info("Building indices in output database...");
+                final Time indicesTime = Time.now();
+                merger.buildIndices();
+                logger.info("Took {} to build indices.", indicesTime.elapsedSince());
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteMergeCommand.java
```diff
@@ -0,0 +1,127 @@
...
+            }
+
```
MikeGost: Same

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteMerger.java
```diff
@@ -0,0 +1,114 @@
...
+
+/**
+ * SQLiteMerger takes a path to an output database file that it will create. It then can attach
+ * Atlas SQLite shard databases and merge those into this output database.
+ *
+ * @author hallahan
+ */
+public class SQLiteMerger
+{
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteMerger.java
```diff
@@ -0,0 +1,114 @@
...
+    public void buildIndices() throws SQLException
+    {
+        try
+        {
+            this.schema.buildIndices(this.connection);
+        }
+        catch (final SQLException sqlException)
+        {
+            logger.error("Unable to build indices on merge database.");
```
MikeGost: Consider logging the error, might help in debugging

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteSchema.java
```diff
@@ -0,0 +1,110 @@
...
+    private static final String USER_VERSION_TEMPLATE = "PRAGMA user_version = %s";
+    private static final String CACHE_SIZE_TEMPLATE = "PRAGMA cache_size = %s";
+
+    private final String tables;
+    private final String indices;
+
+    /**
+     * The constructor loads the schema file of the matching schema version we are using.
+     */
+    SQLiteSchema() throws IOException, NullPointerException
```
MikeGost: This is pretty rare to see - throwing NPE. Thoughts on IllegalArgument instead - NPE may lead people down wrong path.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteSchema.java
```diff
@@ -0,0 +1,110 @@
...
+                    .getFile();
+            this.tables = FileUtils.readFileToString(new java.io.File(fileName),
+                    StandardCharsets.UTF_8);
+
+            resourceName = String.format(INDICES_FILE_NAME_TEMPLATE, SCHEMA_VERSION);
+            fileName = Objects.requireNonNull(getClass().getClassLoader().getResource(resourceName))
+                    .getFile();
+            this.indices = FileUtils.readFileToString(new java.io.File(fileName),
+                    StandardCharsets.UTF_8);
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteSchema.java
```diff
@@ -0,0 +1,110 @@
...
+        catch (final IOException noSchema)
+        {
+            logger.error("Unable to read SQLite schema file in resources.", noSchema);
+            throw noSchema;
+        }
+    }
+
+    void buildIndices(final Connection connection) throws SQLException
+    {
+        try (Statement statement = connection.createStatement())
```
MikeGost: Nice use of try with resources.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteWriter.java
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * @author hallahan
+ */
+public class SQLiteWriter
+{
+    private static final Logger logger = LoggerFactory.getLogger(SQLiteWriter.class);
+
+    private static final String CONNECTION_URL_TEMPLATE = "jdbc:sqlite:%s";
+
+    private final Connection connection;
```
MikeGost: Connection never closed.

src/main/java/org/openstreetmap/atlas/proto/adapters/ProtoPackedTagStoreAdapter.java
MikeGost: "says Nick" :) :)

src/main/java/org/openstreetmap/atlas/streaming/resource/FileSuffix.java
```diff
@@ -26,7 +26,8 @@
     TEXT(".txt"),
     ZIP(".zip"),
     WKT(".wkt"),
-    WKB(".wkb");
+    WKB(".wkb"),
```
MikeGost: Love it.

adahn6:
@MikeGost changes addressed!

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteDictionaryWriter.java
mgcuthbert: To add to this comment, I would be very wary of using connections the way you are here by including it as a class variable. I am not sure if there is a need for a connection pool, however you should have a class (like a connection pool) to handle the connections and dole them out to functions as requested. Otherwise you are holding onto that connection the whole time this class is in memory.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+    private final ItemType itemType;
+
+    SQLiteItemWriter(final Connection connection, final boolean buildIndices,
+            final PackedTagStore tagStore, final LongToLongMap identifierToTagStoreIndex,
+            final ItemType itemType) throws SQLException
+    {
+        this.tagStore = tagStore;
+        this.identifierToTagStoreIndex = identifierToTagStoreIndex;
+        this.itemType = itemType;
+        this.connection = connection;
```
mgcuthbert: Again generally not best practice to store connections as class variables.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+                unwrappedStatement.setDouble(THIRD_COLUMN, maxLon);
+                unwrappedStatement.setDouble(FOURTH_COLUMN, minLat);
+                unwrappedStatement.setDouble(FIFTH_COLUMN, maxLat);
+
+                unwrappedStatement.execute();
+            }
+        }
+
+        this.connection.commit();
+        this.connection.setAutoCommit(true);
```
mgcuthbert: All statements would need to be closed after they have been executed/used.

mgcuthbert: Generally best approach is to use them in a try block like 
`try (PreparedStatement unwrappedStatement = this.rtreeStatementOptional.get())
{
  //do something with the statement
}
`
Because the PreparedStatement is AutoCloseable it will clean up everything for you.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
```diff
@@ -0,0 +1,216 @@
...
+
+            if (keyIndices.length != valueIndices.length)
+            {
+                throw new CoreException(
+                        "key and value array length mismatch, something went wrong");
+            }
+
+            for (int index = 0; index < keyIndices.length; index++)
+            {
+                this.tagsStatement.setLong(FIRST_COLUMN, atlasId);
```
mgcuthbert: This is one of the cases where I would argue that Magic Numbers should be ignored in the check style.

src/main/resources/org/openstreetmap/atlas/sqlite/tables_v1.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE metadata (key text NOT NULL, value);
+CREATE TABLE dictionary (key integer PRIMARY KEY, value);
+CREATE TABLE node (id integer PRIMARY KEY, checksum text, shard text, geom blob NOT NULL);
+CREATE TABLE node_tags (id integer NOT NULL, key integer NOT NULL, value integer NOT NULL);
+CREATE TABLE edge (id integer PRIMARY KEY, checksum text, shard text, geom blob NOT NULL);
```
mgcuthbert: I would recommend change the shard column to integer. You can generate shard id's by taking Zoom-X-Y and just removing the dashes and making it an integer. This will give you a performance gain if using it in your where clause especially if you have a index on it.

mgcuthbert: Also for edges I would suggest storing start and end node identifiers.

src/main/resources/org/openstreetmap/atlas/sqlite/tables_v1.sql
```diff
@@ -0,0 +1,13 @@
+CREATE TABLE metadata (key text NOT NULL, value);
+CREATE TABLE dictionary (key integer PRIMARY KEY, value);
+CREATE TABLE node (id integer PRIMARY KEY, checksum text, shard text, geom blob NOT NULL);
```
mgcuthbert: You might want to considering storing incoming and outgoing edge identifiers for a node either in an array column or in another table. Although if you store it in a separate table that other table could get very large very quickly.

mgcuthbert:
I added a couple of comments, one thing I would strongly suggest is using a connection pool library like HikariCP to handle your connections for you. I think it would make your life a lot easier.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteMerger.java
```diff
@@ -0,0 +1,114 @@
...
+        try (Statement statement = this.connection.createStatement())
+        {
+            // Attach merge database.
+            final String attachSql = String.format(ATTACH_SQL, mergeDatabase.getAbsolutePath());
+            statement.execute(attachSql);
+            logger.info("Attached database {} for merge.", databaseName);
+
+            // Do the merge!
+            statement.executeUpdate(MERGE_EDGES_SQL);
+            statement.executeUpdate(MERGE_EDGES_TAGS_SQL);
```
mgcuthbert: Is there a reason why we don't merge other types? I don't think I follow exactly how this is merging 2 sqlite database.

lucaspcram:
Thanks for the notes @mgcuthbert . The connections stuff could definitely use some cleanup. However, there have been some scope changes and so we are closing this PR.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteItemWriter.java
hallahan: In the past I've done this, but for these purposes, we know we only have one connection to the DB ever. Explicitly closing the connection has no effect on the state of the underlying database file.

src/main/resources/org/openstreetmap/atlas/sqlite/tables_v1.sql
hallahan: Yeah, totally. I was thinking about doing both things.

src/main/resources/org/openstreetmap/atlas/sqlite/tables_v1.sql
hallahan: For rendering purposes, this can just go in the properties. If this becomes an underlying format for atlas, absolutely! There's a ton of things we can do to optimize the schema and the encoding. For one, the database gets big fast, and some of the compression techniques you see in the current atlas format are good. Like, the Mapquest LineString compression, and the lat lons packed into int64s. SQLite gives you the indexing, but how you store the underlying data being indexed is up to us.

src/main/java/org/openstreetmap/atlas/geography/atlas/sqlite/SQLiteDictionaryWriter.java
mgcuthbert: For something like SQLite it is a little different because it is in-memory. However it is generally considered best practice to retrieve from a connection pool and then close or return connections as soon as you are done with it. For a shared database the benefits of this kind of practice are very obvious. Connection leaks are much more dangerous and can cause your database to run out of open connections making not only your local application start to fail, but also other applications connecting to the database. A connection pool allows you to limit the connections your application makes and shares an open connection among threads within you application. This gives you a bit of performance so that the application doesn't have to continue opening and closing connections to the database. All these things are not as problematic with something like an in-memory database like SQLite. However if your program grows or even if you start using multiple threads within your application, the same principles apply. 

To answer your "Dumb Question" (and it is in quotes because it really is not a dumb question at all, once the JVM exist all resources associated with the JVM will "close". In this case the socket resource to the database will automatically be shut down. However the connection will not be closed properly. A database can react unexpectedly in that scenario and usually the session in the database will be shutdown in a couple of seconds after the socket closes but there may be cases where the session is kept alive for longer than you would ordinarily want. One such case would be if a SQL query is running at the time the JVM exits, it could cause the session to remain alive for a lot longer than expected, especially if you get into a hairy deadlock kind of situation.

Anyway I understand that in this specific instance connection handling won't have as much impact as it would if you were connecting to a shared database resource, but I would still strongly recommend following best practices so that you won't find yourself in a difficult spot if the code outgrows the initial use cases.

PR #233, opened 2018-10-04
Added in Relation flattening

adahn6:
### Description:

This PR adds in a function to Relation.java that "flattens" the Relation into a set of its non-Relation child members. 
### Potential Impact:

Probably nothing, giving it's a new API addition.

### Unit Test Approach:

Unit tests cover basic relation (one node), nested relations (several nodes and one nested relation), and recursive relations (relation with itself as a child).

### Test Results:

```
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:62 - Relation 6: [Relation: id=6, [Members: 
				{Member: ID = 1, Type = NODE, Role = outside}
			], [Tags: ]]
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:64 - Flattened: [[Node: id=1, location=POINT (-122.009566 37.33531), inEdges=[], outEdges=[], [Tags: ]]]
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:43 - Relation 8: [Relation: id=8, [Members: 
				{Member: ID = 4, Type = NODE, Role = outside}, 
				{Member: ID = 5, Type = NODE, Role = outside}, 
				{Member: ID = 8, Type = RELATION, Role = outside}
			], [Tags: ]]
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:45 - Flattened: [[Node: id=4, location=POINT (-122.009566 37.33531), inEdges=[], outEdges=[], [Tags: ]], [Node: id=5, location=POINT (-122.009566 37.33531), inEdges=[], outEdges=[], [Tags: ]]]
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:24 - Relation 7: [Relation: id=7, [Members: 
				{Member: ID = 1, Type = NODE, Role = outside}, 
				{Member: ID = 2, Type = NODE, Role = outside}
			], [Tags: ]]
2018-10-04 13:48:28 INFO  [main] RelationFlatteningTest:26 - Flattened: [[Node: id=1, location=POINT (-122.009566 37.33531), inEdges=[], outEdges=[], [Tags: ]], [Node: id=2, location=POINT (-122.009566 37.33531), inEdges=[], outEdges=[], [Tags: ]]]
```

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
@@ -127,12 +127,12 @@ public abstract class Relation extends AtlasEntity implements Iterable<RelationM
             {
                 ((Relation) polledMember).members()
                         .forEach(member -> toProcess.add(member.getEntity()));
+                relationsSeen.add(polledMember.getIdentifier());
             }
             else
             {
                 relationMembers.add(polledMember);
             }
-            relationsSeen.add(this.getIdentifier());
         }
         return relationMembers;
     }
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
@@ -118,13 +118,12 @@ public abstract class Relation extends AtlasEntity implements Iterable<RelationM
         while (!toProcess.isEmpty())
         {
             polledMember = toProcess.poll();
-            if (relationsSeen.contains(polledMember.getIdentifier()))
-            {
-                continue;
-            }
-
             if (polledMember instanceof Relation)
             {
+                if (relationsSeen.contains(polledMember.getIdentifier()))
+                {
+                    continue;
+                }
                 ((Relation) polledMember).members()
                         .forEach(member -> toProcess.add(member.getEntity()));
                 relationsSeen.add(polledMember.getIdentifier());
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
@@ -45,8 +45,8 @@ public abstract class Relation extends AtlasEntity implements Iterable<RelationM
      */
     public enum Ring
     {
-    OUTER,
-    INNER
+        OUTER,
+        INNER
     }
 
     private static final Logger logger = LoggerFactory.getLogger(Relation.class);
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
@@ -2,8 +2,11 @@ package org.openstreetmap.atlas.geography.atlas.items;
 
 import java.util.ArrayList;
 import java.util.Comparator;
+import java.util.Deque;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
@@ -24,6 +27,8 @@ import org.openstreetmap.atlas.tags.RelationTypeTag;
 import org.openstreetmap.atlas.tags.annotations.validation.Validators;
 import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.StringList;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * An OSM relation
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -93,6 +99,44 @@ public String configurableString(final String betweenEachMemberAndRelation,
...
+        final Set<AtlasObject> relationMembers = new HashSet<>();
+        final Deque<AtlasObject> toProcess = new LinkedList<>();
+        final Set<Long> relationsSeen = new HashSet<>();
+        AtlasObject polledMember;
+
+        toProcess.add(this);
+        while (!toProcess.isEmpty())
+        {
+            polledMember = toProcess.poll();
+            if (relationsSeen.contains(polledMember.getIdentifier()))
```
MikeGost: This would be a really rare case, but OSM identifiers are only unique within their type - meaning a Relation and Way can have the same identifier, but two Relations cannot. Because of this, it's technically possible for two different `AtlasEntity`s to have the same identifier. To avoid this, we also want to check the type. I would add an explicit check that the `polledMember` is a `Relation`, in addition to having been seen. You can add this check in at line 127 when you're checking for Relations.

src/test/java/org/openstreetmap/atlas/geography/atlas/items/RelationFlatteningRule.java
```diff
@@ -0,0 +1,41 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Node;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Relation;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Relation.Member;
+
+/**
+ * @author samuelgass
+ */
+public class RelationFlatteningRule extends CoreTestRule
+{
+
```
MikeGost: Nit: extra line

src/test/java/org/openstreetmap/atlas/geography/atlas/items/RelationFlatteningRule.java
```diff
@@ -0,0 +1,41 @@
...
+                    @Relation(id = "7", members = {
+                            @Member(id = "1", role = "outside", type = "node"),
+                            @Member(id = "2", role = "outside", type = "node") }),
+                    @Relation(id = "8", members = {
+                            @Member(id = "4", role = "outside", type = "node"),
+                            @Member(id = "5", role = "outside", type = "node"),
+                            @Member(id = "8", role = "outside", type = "relation") }) }
+
+    )
+    private Atlas atlas1;
```
MikeGost: Nit: there is only one atlas - no need for 1.

src/test/java/org/openstreetmap/atlas/geography/atlas/items/RelationFlatteningRule.java
```diff
@@ -0,0 +1,41 @@
...
+
+    public Atlas getAtlas1()
```
MikeGost: Nit: likewise here - getAtlas?

src/test/java/org/openstreetmap/atlas/geography/atlas/items/RelationFlatteningRule.java
```diff
@@ -0,0 +1,41 @@
...
+                    @Node(id = "5") },
+
+            relations = {
+
+                    @Relation(id = "6", members = {
+                            @Member(id = "1", role = "outside", type = "node") }),
+                    @Relation(id = "7", members = {
+                            @Member(id = "1", role = "outside", type = "node"),
+                            @Member(id = "2", role = "outside", type = "node") }),
+                    @Relation(id = "8", members = {
```
MikeGost: How about testing a valid relation with one or more sub-relations? Currently this is only testing an invalid relationship. It may be good to separate these into individual tests and test atlases. Something like below. Thoughts?

- testShallowRelation
- testNestedRelation
- testLoopingRelation

adahn6:
@MikeGost changes addressed!

src/test/java/org/openstreetmap/atlas/geography/atlas/items/RelationFlatteningRule.java
adahn6: good idea

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
MikeGost: I'm probably reading it wrong - won't the issue still be present? If we have a `polledMember`, that's a `Node`, with the same identifier as an already seen `Relation` - that `Node` wouldn't make it into the final return set, right? May be worth adding a unit test case for this too :)

adahn6: oh, I see, this is my mistake-- I had initially thought you were just referring to the adding of an ID as seen. but, obviously, it should apply to the check as well. I'll move it in, and also make a unit test.

PR #236, opened 2018-10-11
Pool can have a companion ticker thread

matthieun:
### Description:

This change brings the ability to schedule a companion ticker thread that allows time-based reporting on the initial Runnable or Callable execution.

### Potential Impact:

No impact, it is a new feature.

### Unit Test Approach:

Tested the new option with Callable and Runnable, and made sure they still report errors properly.

### Test Results:

Companion thread has the same lifespan as other thread.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/threads/Pool.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/threads/Pool.java
@@ -126,9 +126,43 @@ public class Pool implements Closeable
         return doWithTheOutput.apply(item);
     }
 
+    public <T> Result<T> queue(final Callable<T> task, final Ticker ticker)
+    {
+        final Callable<T> taskWrapper = () ->
+        {
+            try
+            {
+                return task.call();
+            }
+            finally
+            {
+                ticker.close();
+            }
+        };
+        this.queue(ticker);
+        return new Result<>(this.pool.submit(taskWrapper), this, ticker);
+    }
+
     public void queue(final Runnable command)
     {
         this.pool.execute(command);
+    }
+
+    public void queue(final Runnable command, final Ticker ticker)
+    {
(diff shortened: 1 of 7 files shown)


src/main/java/org/openstreetmap/atlas/utilities/threads/Pool.java
```diff
@@ -126,11 +126,45 @@ public boolean isDead()
...
+            try
+            {
+                command.run();
+            }
+            finally
+            {
+                ticker.close();
+            }
+        };
+        this.pool.execute(ticker);
```
lucaspcram: One thing to note here. When using the `Pool` with a `Runnable`, it fails fast if the `Runnable` throws an exception. So you could have a case where the `ticker` throws an exception before the `queue` method even gets to the next line.

matthieun: Actually you are right, let me add ticker error handling!

matthieun:
Rhaaaaaa!

src/test/java/org/openstreetmap/atlas/utilities/threads/PoolTest.java
```diff
@@ -138,6 +138,48 @@ public void testTickerCallableWithErrors()
...
+                    final Ticker rogueTicker = new Ticker("Ticker " + idx,
+                            Duration.milliseconds(50))
+                    {
+                        @Override
+                        protected void tickAction(final Duration sinceStart)
+                        {
+                            logger.info("{}: {}", getName(), sinceStart);
+                            if ("Ticker 3".equals(getName()))
+                            {
+                                throw new CoreException("I am rogue Ticker 3");
```
lucaspcram: Hahaha "rogue ticker" I love this :D

PR #237, opened 2018-10-11
Added Boundary deduplication logic

jwpgage:
### Description:

This PR adds filtering to the Complex Boundaries when creating a country boundary map from an atlas file. Duplicate outer polygons are eliminated, meaning that only one country can claim a given outer polygon, such as an island. This implementation required adding a method of removing outers from a `ComplexBoundary`.

### Potential Impact:

Enforces a 1:1 mapping of Outer Boundary Polygon -> Country Code.

### Unit Test Approach:

~~Unit test forthcoming.~~

Unit test is on an island tagged by USA and HTI. Checks that only one of the countries is assigned the island in the boundary map.

### Test Results:

Tested on Central America, resolved boundary deduplication and produced a valid OSM border file.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/boundaries/ComplexBoundary.java
```diff
@@ -120,6 +122,23 @@ public boolean hasCountryCode()
...
+    public void removeOuter(final Polygon outerToRemove)
+    {
+        final MultiMap<Polygon, Polygon> outersToInners = new MultiMap<>();
+        this.outline.outers().forEach(outer ->
+        {
+            final List<Polygon> innersForThisOuter = this.outline.innersOf(outer);
+            outersToInners.put(outer, innersForThisOuter);
+        });
+        outersToInners.remove(outerToRemove);
+        this.outline = new MultiPolygon(outersToInners);
```
matthieun: Use setOutline here?

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -482,7 +483,9 @@ STRtree getRawIndex()
      */
     void readFromAtlas(final Atlas atlas)
     {
-        for (final ComplexBoundary complexBoundary : new ComplexBoundaryFinder().find(atlas))
+        final Iterable<ComplexBoundary> dedupedComplexBoundaries = resolveOverlappingBorders(
```
matthieun: Avoid abbreviations here: `deduplicatedComplexBoundaries`

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1118,6 +1121,55 @@ public void initializeGridIndex(final Set<String> countries)
...
+     *            The {@link ComplexBoundary}'s to filter
+     * @return An iterable of {@link ComplexBoundary} with no duplicate outer polygons
+     */
+    public Iterable<ComplexBoundary> resolveOverlappingBorders(
+            final Iterable<ComplexBoundary> complexBoundaries)
+    {
+        final List<ComplexBoundary> dedupedComplexBoundaries = new ArrayList<>();
+        final Set<org.openstreetmap.atlas.geography.Polygon> processedOuters = new HashSet<>();
+        final List<ComplexBoundary> complexBoundariesList = Lists.newArrayList(complexBoundaries);
+        Collections.sort(complexBoundariesList, (firstBoundary, secondBoundary) ->
```
matthieun: Wouldn't it be better to use a SortedSet here? That way it would be sorted at insertion time. Can this contain duplicates? If yes, then Set won't work.

adahn6: the sorted collection here doesn't seem used for by the deduplication-- does it have a purpose that I'm not seeing? or should it be removed?

jwpgage: Since we pass in an `Iterable`, we would still need to convert from an `Iterable` to a `Collection` before being able to insert into a `SortedSet`, right? Therefore we'd still need to an insertion prior to the sorting.

Also, to my mind this list could contain duplicates, as this is the list of `ComplexBoundary` objects, not the set of outer polygons.

jwpgage: Good catch! Meant to loop over the resulting `complexBoundariesList`. Fixing now.

matthieun:
Also I would rename USA_boundary to USA_boundary_reduced so there is no confusion here :)

PR #243, opened 2018-10-15
AtlasStatistics: protect from malformed multipolygons

matthieun:
### Description:

When multipolygons were malformed, the AtlasStatistics computation would fail on `IllegalArgumentException` in their area computation.

### Potential Impact:

This will allow skipping malformed multipolygons.

### Unit Test Approach:

Added a malformed multipolygon in the water test.
image

### Test Results:

Error is logged and area is not counted.

------


MikeGost:
Looks like a spotless needs to be applied.

PR #251, opened 2018-10-22
subAtlas(Polygon) missing reverse edge and performance

matthieun:
### Description:

This fixes two issues:
- `subAtlas(Polygon)` would in some cases include a reverse edge without its master edge, in a very corner case where JTS would match the reverse edge to the `Polygon`, but not the master edge.
```
org.openstreetmap.atlas.geography.atlas.exception.AtlasIntegrityException: Cannot build an Atlas with a negative edge without its positive counterpart: -XXXXXXXXX
	at org.openstreetmap.atlas.geography.atlas.packed.PackedAtlasBuilder.lambda$verifyNegativeEdgesHaveMasterEdge$1(PackedAtlasBuilder.java:319)
	at java.lang.Iterable.forEach(Iterable.java:75)
	at org.openstreetmap.atlas.geography.atlas.packed.PackedAtlasBuilder.verifyNegativeEdgesHaveMasterEdge(PackedAtlasBuilder.java:312)
	at org.openstreetmap.atlas.geography.atlas.packed.PackedAtlasBuilder.get(PackedAtlasBuilder.java:211)
	at org.openstreetmap.atlas.geography.atlas.BareAtlas.subAtlas(BareAtlas.java:629)
```
- Improve `subAtlas` performance by caching the edges intersecting the polygon, instead of re-computing them 3 times.

### Potential Impact:

A more accurate `subAtlas` in very corner cases. A faster `subAtlas` in most cases.

### Unit Test Approach:

Added one unit test that exposes the corner case, and verifies that the master edge gets added.

### Test Results:

Tests succeed.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
@@ -953,7 +953,7 @@ public abstract class BareAtlas implements Atlas
     private <Member extends AtlasEntity> Supplier<Iterable<Member>> getCachingSupplier(
             final Iterable<Member> source, final ItemType type)
     {
-        final List<Set<Long>> memberIdentifiersIntersecting = new ArrayList<>();
+        final Set<Long> memberIdentifiersIntersecting = new HashSet<>();
         // A supplier that will effectively cache all the members intersecting that polygon. This is
         // thread safe because the cache is internal to the supplier's scope.
         @SuppressWarnings("unchecked")
@@ -961,19 +961,17 @@ public abstract class BareAtlas implements Atlas
         {
             if (memberIdentifiersIntersecting.isEmpty())
             {
-                final Set<Long> intersecting = new HashSet<>();
-                memberIdentifiersIntersecting.add(intersecting);
                 // Here using a map instead of forEach to pipe the edge identifiers to the cache
                 // list without having to re-open the iterable.
                 return Iterables.stream(source).map(member ->
                 {
-                    intersecting.add(member.getIdentifier());
+                    memberIdentifiersIntersecting.add(member.getIdentifier());
                     return member;
                 }).collect();
             }
             else
             {
-                return (Iterable<Member>) Iterables.stream(memberIdentifiersIntersecting.get(0))
+                return (Iterable<Member>) Iterables.stream(memberIdentifiersIntersecting)
                         .map(identifier -> this.entity(identifier, type)).collect();
             }
         };
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
@@ -579,6 +579,8 @@ public abstract class BareAtlas implements Atlas
         {
             if (builder.peek().edge(edge.getIdentifier()) == null)
             {
+                // Here, making sure that edge identifiers are not 0 to work around an issue in unit
+                // tests: https://github.com/osmlab/atlas/issues/252
                 if (edge.getIdentifier() != 0 && edge.hasReverseEdge())
                 {
                     final Edge reverse = edge.reversed().get();
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
@@ -573,13 +573,13 @@ public abstract class BareAtlas implements Atlas
         {
             if (builder.peek().edge(edge.getIdentifier()) == null)
             {
-                if (!edge.isMasterEdge())
+                if (edge.hasReverseEdge())
                 {
-                    final Edge master = edge.reversed().get();
-                    if (builder.peek().edge(master.getIdentifier()) == null)
+                    final Edge reverse = edge.reversed().get();
+                    if (builder.peek().edge(reverse.getIdentifier()) == null)
                     {
-                        builder.addEdge(master.getIdentifier(), master.asPolyLine(),
-                                master.getTags());
+                        builder.addEdge(reverse.getIdentifier(), reverse.asPolyLine(),
+                                reverse.getTags());
                     }
                 }
                 builder.addEdge(edge.getIdentifier(), edge.asPolyLine(), edge.getTags());
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
@@ -526,8 +526,6 @@ public abstract class BareAtlas implements Atlas
                 linesIntersecting(boundary), ItemType.LINE);
         final Supplier<Iterable<Point>> pointsWithin = getCachingSupplier(pointsWithin(boundary),
                 ItemType.POINT);
-        final Supplier<Iterable<Relation>> relationsWithEntitiesIntersecting = getCachingSupplier(
-                relationsWithEntitiesIntersecting(boundary), ItemType.RELATION);
 
         // Generate the size estimates, then the builder.
         // Nodes estimating is a bit tricky. We want to include all the nodes within the polygon,
@@ -541,7 +539,7 @@ public abstract class BareAtlas implements Atlas
         final long lineNumber = Math.round(Iterables.size(linesIntersecting.get()) * ratioBuffer);
         final long pointNumber = Math.round(Iterables.size(pointsWithin.get()) * ratioBuffer);
         final long relationNumber = Math
-                .round(Iterables.size(relationsWithEntitiesIntersecting.get()) * ratioBuffer);
+                .round(Iterables.size(relationsWithEntitiesIntersecting(boundary)) * ratioBuffer);
         final AtlasSize size = new AtlasSize(edgeNumber, nodeNumber, areaNumber, lineNumber,
                 pointNumber, relationNumber);
         final PackedAtlasBuilder builder = new PackedAtlasBuilder().withSizeEstimates(size)
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
@@ -516,8 +516,18 @@ public abstract class BareAtlas implements Atlas
     {
         final Time begin = Time.now();
 
+        final Supplier<Iterable<Node>> nodesWithin = getCachingSupplier(nodesWithin(boundary),
+                ItemType.NODE);
         final Supplier<Iterable<Edge>> edgesIntersecting = getCachingSupplier(
                 edgesIntersecting(boundary), ItemType.EDGE);
+        final Supplier<Iterable<Area>> areasIntersecting = getCachingSupplier(
+                areasIntersecting(boundary), ItemType.AREA);
+        final Supplier<Iterable<Line>> linesIntersecting = getCachingSupplier(
+                linesIntersecting(boundary), ItemType.LINE);
+        final Supplier<Iterable<Point>> pointsWithin = getCachingSupplier(pointsWithin(boundary),
+                ItemType.POINT);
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -914,6 +942,34 @@ private void checkRelationMembersAndIndexRelation(final Relation relation,
...
+                memberIdentifiersIntersecting.add(intersecting);
+                return Iterables.stream(source).map(member ->
+                {
+                    intersecting.add(member.getIdentifier());
+                    return member;
+                }).collect();
+            }
+            else
+            {
+                return (Iterable<Member>) Iterables.stream(memberIdentifiersIntersecting.get(0))
```
adahn6: I'm confused about the purpose of the List wrapper around the Set here. Since this pattern limits the List to fetching only one set, why not just use the Set directly?

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -914,6 +942,34 @@ private void checkRelationMembersAndIndexRelation(final Relation relation,
...
+        // A supplier that will effectively cache all the members intersecting that polygon. This is
+        // thread safe because the cache is internal to the supplier's scope.
+        @SuppressWarnings("unchecked")
+        final Supplier<Iterable<Member>> result = () ->
+        {
+            if (memberIdentifiersIntersecting.isEmpty())
+            {
+                final Set<Long> intersecting = new HashSet<>();
+                memberIdentifiersIntersecting.add(intersecting);
+                return Iterables.stream(source).map(member ->
```
lucaspcram: I find it a little confusing that the call to `map` does not actually perform a mapping operation, and instead produces a side effect on some external global state. I think you are doing this to avoid calling `Iterables.stream()` twice (Once for a `forEach` and once for the `collect`). So here, you are using `map` as a modified `forEach` that returns an identical stream (so that `collect` can be called in the same line)?

If this is the case, maybe leave a comment that explains this, since this use of `map` is bit non-idiomatic :)

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -914,6 +942,34 @@ private void checkRelationMembersAndIndexRelation(final Relation relation,
         }
     }
 
+    private <Member extends AtlasEntity> Supplier<Iterable<Member>> getCachingSupplier(
+            final Iterable<Member> source, final ItemType type)
+    {
+        final List<Set<Long>> memberIdentifiersIntersecting = new ArrayList<>();
```
lucaspcram: Could this also work as an `Optional`? Might make things a little clearer, since it's a singleton `List` anyway.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
adahn6: +1 to this

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -560,9 +568,24 @@ public SnappedEdge snapped(final Location point, final Distance threshold)
...
 
-        // Add the edges
-        edgesIntersecting(boundary).forEach(
-                edge -> builder.addEdge(edge.getIdentifier(), edge.asPolyLine(), edge.getTags()));
+        // Add the edges. Use a consumer that makes sure master edges are always added first.
+        final Consumer<Edge> edgeAdder = edge ->
+        {
+            if (builder.peek().edge(edge.getIdentifier()) == null)
+            {
+                if (edge.getIdentifier() != 0 && edge.hasReverseEdge())
```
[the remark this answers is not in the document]
lucaspcram: I agree.

It seems to me that swapping `edge.getIdentifier() != 0` with `!isMasterEdge()` would correct this code segment.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
matthieun: Because the object that goes from outside to inside the lambda has to be final. I cannot make the set final as I use the fact it might not be there as a cause for initializing it. So wrapping it in a final list is a trick that might be ugly but works :) I am open to any other more elegant suggestion!

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -560,9 +568,24 @@ public SnappedEdge snapped(final Location point, final Distance threshold)
         nodesWithin(boundary, hasNode.negate()).forEach(
                 node -> builder.addNode(node.getIdentifier(), node.getLocation(), node.getTags()));
 
-        // Add the edges
-        edgesIntersecting(boundary).forEach(
-                edge -> builder.addEdge(edge.getIdentifier(), edge.asPolyLine(), edge.getTags()));
+        // Add the edges. Use a consumer that makes sure master edges are always added first.
```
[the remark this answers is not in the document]
matthieun: So if the `1` and `-1` have not been added yet there are multiple use cases:
- The consumer is told to add `1` first.
  - If `1` has a reverse edge, it will add `-1` (if not added already), then `1` (if not added already)
  - If `1` does not have a reverse edge, it will just add `1` (if not added already).
- The consumer is told to add `-1` first.
  - It will add `1` first (if not added already) then `-1` (if not added already)

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
matthieun: We have a lot of unit tests that unfortunately create edges with id 0 using `@TestAtlas`. The whole idea of having a `+` and `-` for forward and reverse edges breaks down unfortunately for edge 0, and all the tests break. This might be worth creating an issue to track it, however the impact is pretty small. Thoughts?

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
matthieun: Well it would not work as an optional given the fact that you cannot swap the internals of an Optional object without changing the optional handle... or can you?
Basically, like in https://github.com/osmlab/atlas/pull/251/files#r227476980, the issue is that I can share objects within the lambda only when they are final. Open to other ideas though!

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
matthieun: That works :) I will add a comment.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
MikeGost: That makes sense, it's impossible to know that without you explaining - I was trying to wrap my head around why 0 was there - it's worth putting that as a comment here. Otherwise, is it possible to clean up tests now? If not, creating an issue like you said to clean up the test and update this piece of code to not rely on 0 might be worthwhile.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -514,18 +516,24 @@ public SnappedEdge snapped(final Location point, final Distance threshold)
...
         // buffer on top of the nodes inside the polygon. This mostly avoids resizing.
-        final double nodeRatioBuffer = 1.2;
-        final long nodeNumber = Math.round(Iterables.size(nodesWithin(boundary)) * nodeRatioBuffer);
-        final long edgeNumber = Iterables.size(edgesIntersecting(boundary));
-        final long areaNumber = Iterables.size(areasIntersecting(boundary));
-        final long lineNumber = Iterables.size(linesIntersecting(boundary));
-        final long pointNumber = Iterables.size(pointsWithin(boundary));
-        final long relationNumber = Iterables.size(relationsWithEntitiesIntersecting(boundary));
+        final double ratioBuffer = 1.2;
+        final long nodeNumber = Math.round(Iterables.size(nodesWithin(boundary)) * ratioBuffer);
```
matthieun: Maybe I should use the caching for all the other types as well!

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
lucaspcram: So in sub-bullet point 1 under bullet point 1, you still add the reverse edge before the master edge. Is this OK in this case? I guess I am just confused by the discrepancy between the comment and what actually happens.

Also, I am curious why we need to make sure the master gets added first. Sorry to be probing, I just want to make sure I understand!

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
lucaspcram: Yeah you can't, so that wouldn't work.... Darn

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
matthieun: Cleaning up the tests will be a full task. Created an issue #252, will add comments too.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
adahn6: hmm... wouldn't it be possible to just use the `isEmpty()` check for initializing? a final `Set` can still be modified via `addAll()`, so you could do something like
```
final Set<Long> intersecting = new HashSet<>();
@SuppressWarnings("unchecked")
final Supplier<Iterable<Member>> result = () -> {
     if(intersecting.isEmpty()) {
           intersecting.addAll(Iterables.asSet(source));
     }
     return (Iterable<Member>) Iterables.stream(intersecting);
}
```
Not sure that code compiles or works, just spitballing here

lucaspcram: I am also curious to know if what @adahn6 suggests works or not.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -514,18 +516,30 @@ public SnappedEdge snapped(final Location point, final Distance threshold)
     {
         final Time begin = Time.now();
 
+        final Supplier<Iterable<Node>> nodesWithin = getCachingSupplier(nodesWithin(boundary),
+                ItemType.NODE);
+        final Supplier<Iterable<Edge>> edgesIntersecting = getCachingSupplier(
+                edgesIntersecting(boundary), ItemType.EDGE);
+        final Supplier<Iterable<Area>> areasIntersecting = getCachingSupplier(
+                areasIntersecting(boundary), ItemType.AREA);
+        final Supplier<Iterable<Line>> linesIntersecting = getCachingSupplier(
```
MikeGost: Wondering if we can add the same caching strategy to the Predicate subAtlas call.

matthieun: If the filtering is expensive and we re-filter many times I'd say yes. But we'd want to run the profiler on a big atlas first to make sure it is a worthwhile change.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
lucaspcram: Oh now I understand. The problem was that the `Iterable` returned by `edgesIntersecting` is sometimes missing the master edge. So it doesn't actually matter which is added first, just that we always add the reverse edge/master edge together if both exist.

PR #253, opened 2018-10-23
TaggableFilter old constructor for backwards compatibility

matthieun:
### Description:

atlas-checks is making extensive use of new `TaggableFilter(definition)`. Adding this back until it is all migrated to `TaggableFilter.forDefinition`.

### Potential Impact:

One new deprecated constructor

### Unit Test Approach:

Same unit tests

### Test Results:

Same unit tests

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/tags/filters/TaggableFilter.java
+++ b/src/main/java/org/openstreetmap/atlas/tags/filters/TaggableFilter.java
@@ -80,6 +80,17 @@ public class TaggableFilter implements Predicate<Taggable>, Serializable
         return new LineFilterConverter().convert(definition);
     }
 
+    /**
+     * @param definition
+     *            The {@link String} definition of the filter
+     * @deprecated Use {@code TaggableFilter.forDefinition(definition)} instead.
+     */
+    @Deprecated()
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/tags/filters/TaggableFilter.java
```diff
@@ -80,6 +80,17 @@ public static TaggableFilter forDefinition(final String definition)
...
 
+    /**
+     * @param definition
+     *            The {@link String} definition of the filter
+     * @deprecated Use {@code TaggableFilter.forDefinition(definition)} instead.
+     */
+    @Deprecated()
+    public TaggableFilter(final String definition)
+    {
+        this(TaggableFilter.forDefinition(definition));
```
yiqingj: in previous behavior if definition is "", it'll return an instance that accepts everything.  with TaggableFilter.forDefinition(""), it will throw exception `org.openstreetmap.atlas.exception.CoreException: Taggable filter definition "" is invalid.`

matthieun: Fixed.

PR #255, opened 2018-10-24
PackedAtlas is now full protobuf by default. 

lucaspcram:
Additionally, we now guarantee a successful first load for proto atlases.

### Description:

`PackedAtlas` now defaults to proto for the save and load format. The logger for load format selection is now `debug`. Also, we now declare `PROTOBUF` first in the enum so that it is the first format that `determineLoadFormat` tries.

### Potential Impact:

N/A

### Unit Test Approach:

Check all proto related unit tests. Updated `testResourceClosureOnInvalidLoadFormat` to reflect the change.

### Test Results:

All tests pass.

------


lucaspcram:
@MikeGost It should work provided your JOSM is up-to-date (ie. is using a version of `atlas` that can understand protobuf atlases). @matthieun and @jwpgage can weigh in here.

matthieun:
@MikeGost Yes, it works if the plugin is updated on your side!

PR #258, opened 2018-10-26
ResourceCache API overhaul

lucaspcram:
### Request From Reviewers:
Please carefully look at the logging changes I have made, and try to anticipate how these will impact the debugging experience. I am trying to balance log verbosity/usefulness. If you see something that should be logged and isn't, something that is logged and shouldn't be, or something that is logged at an inappropriate level, *leave a comment.*

There are also quite a few thorny issues involving concurrency. These manifest in different ways at the various layers of the caching abstractions (see the class javadoc comments). I did my best to synchronize things properly, but extra eyes are always appreciated. Concurrency bugs down the road are something we want to avoid.

### Description:

TL;DR skip to the **Potential Impact** section.

This is a non-trivial refactor of some internals within the `ResourceCache` and related code. This is mostly to support more advanced/complex use cases, and provide more granular control over the caches.

Change Summary:

1) Much better logging. Lots of `info` noise has been moved to `trace`, and more useful `warn` messages have been added. Additionally, caches and strategies now generate a random `UUID` at creation time to tag their logs. This should help in environments with multiple caches.

2) `SystemTemporaryFileCachingStrategy` has been decomposed into `NamespaceCachingStrategy` and `GlobalNamespaceCachingStrategy`. Using `GlobalNamespaceCachingStrategy` will yield mostly identical behaviour to `SystemTemporaryFileCachingStrategy`. However, `NamespaceCachingStrategy` allows callers to provided a namespace at creation time, which is used to give more granular control over misses, hits, and invalidates.

3) Added a `ResourceCache` interface to super the `ConcurrentResourceCache`. This is mostly to allow for some future flexibility in designing new caches.

4) Internally, the strategy fetchers now return `Optional<Resource>` instead of a bare `Resource`. This gives the fetcher functions more control over how to handle cases when a fetched `Resource` is invalid in some way. It also allows the strategies to provide more relevant logging.

5) Invalidation is now fully supported by all cache and strategy implementations. Additionally, we now support invalidation of individual `URI`s, for more selective control over cache contents.

### Potential Impact:

1) Note that the changes in this PR will break `HadoopAtlasFileCache` in `atlas-generator` if `atlas-generator` is updated to depend on an `atlas` with these changes, without itself making the necessary corrections to `HadoopAtlasFileCache`. A part 2 PR in `atlas-generator` with the necessary fixes is incoming.

2) `SystemTemporaryFileCachingStrategy` has been removed in favor of a more modular design. If anyone downstream was using this class directly, they will need to update their code. However, anyone downstream using `HadoopAtlasFileCache` should be sheltered from breakage.

### Unit Test Approach:

I have updated `CachingTests` to check the new `invalidate(URI)` feature: `testIndividualResourceInvalidation`

### Test Results:

All tests continue to pass.

------


lucaspcram:
Umm I am not sure what happened with this `travis` failure. Anyone seen this before? It looks like a test that relies on some random behaviour. Maybe we just got unlucky?

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
```diff
@@ -72,16 +60,51 @@ public ConcurrentResourceCache(final CachingStrategy cachingStrategy,
 
         if (!cachedResource.isPresent())
         {
-            logger.warn("Cache fetch failed, falling back to default fetcher...");
+            logger.warn("CacheID {}: cache fetch failed, falling back to default fetcher...",
```
matthieun: Like the cache id inclusion here. Very searchable in mountains of logs.

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/GlobalNamespaceCachingStrategy.java
```diff
@@ -0,0 +1,65 @@
...
+ *
+ * @author lcram
+ */
+public class GlobalNamespaceCachingStrategy extends NamespaceCachingStrategy
+{
+    /*
+     * This is a random SHA256 hash. Collisions with this namespace are astronomically unlikely (due
+     * to the 256 bits of entropy).
+     */
+    private static final String GLOBAL_NAMESPACE = "3707740A818531237051A0F1E086CF701E2C38483675FCD1AAD8F5C5C33F19BC";
```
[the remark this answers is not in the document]
matthieun: In the Spark use-case, we would want to not do that, right? So multiple JVMs on the same node can still share the same cache?

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/NamespaceCachingStrategy.java
```diff
@@ -0,0 +1,223 @@
...
+            getCachedFile(resourceURI).delete();
+        }
+        catch (final Exception exception)
+        {
+            logger.warn("StrategyID {}: invalidate of resource {} failed due to {}",
+                    this.getStrategyID(), resourceURI, exception);
+        }
+    }
+
+    private void attemptToCacheFileLocally(final File cachedFile,
```
matthieun: Any point this method should throw an exception instead of just logging errors?

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/GlobalNamespaceCachingStrategy.java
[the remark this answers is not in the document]
lucaspcram: Well, it's a tradeoff worth discussing. Multiple JVMs sharing the same underlying cache would be good for performance (since it avoids downloading duplicate `Resources`). However, once you allow that you have no way to guarantee that a bad interleaving of `invalidate(my_uri)` in one JVM and `get(my_uri)` in another JVM won't occur, since at that point you are at the mercy of the OS scheduler.

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/NamespaceCachingStrategy.java
lucaspcram: I actually don't think it should, since an exception would needlessly halt execution. If `attemptToCacheFileLocally` fails, then what ultimately happens is that the `NamespaceCachingStrategy.attemptFetch(URI)` returns an empty `Optional`. From there, the `ConcurrentResourceCache` object containing that `NamespaceCachingStrategy` will simply fall back to the default fetcher.

I am open to having my mind changed on this.

lucaspcram:
As to your bullet points:

1) >A fetcher randomly fails (retry strategy?)

This may be worth adding in the `ConcurrentResourceCache`. We can discuss.

2) >There is a File permission mismatch?

This will only affect the `NamespaceCachingStrategy`. What would happen is that `NamespaceCachingStrategy.attemptToCacheFileLocally` would fail, log an `ERROR` detailing the problem (see `NamespaceCachingStrategy` line 147/173), and then the parent `ConcurrentResourceCache` would fall back on the default fetcher.

3) >A JVM invalidates the cache while another is fetching the resource?

[See my comment here.](https://github.com/osmlab/atlas/pull/258#discussion_r228984876)

matthieun:
Sounds good, all this makes sense. Do you want to add more tests?

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
```diff
@@ -34,31 +39,14 @@
      *            the default fetcher
      */
     public ConcurrentResourceCache(final CachingStrategy cachingStrategy,
-            final Function<URI, Resource> fetcher)
+            final Function<URI, Optional<Resource>> fetcher)
```
jklamer: 👍

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
```diff
@@ -34,31 +39,14 @@
...
     {
         this.cachingStrategy = cachingStrategy;
         this.fetcher = fetcher;
+        this.cacheID = UUID.randomUUID();
```
jklamer: If this is used just for logging it might be helpful to have the name of the Cache with a unique ID. 
Something to the effective of "ConcurrentResourceCache instance {}: <message here>". Theoretically you should know what caching you're using but in the rare case of using two different types of caching in a job, it would be hard to differentiate without the cache name.

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/AbstractCachingStrategy.java
```diff
@@ -24,20 +20,26 @@
      */
     private final Map<String, UUID> uriStringToUUIDCache;
 
+    private final UUID strategyID;
+
     public AbstractCachingStrategy()
     {
-        this.uriStringToUUIDCache = new HashMap<>();
+        this.uriStringToUUIDCache = new ConcurrentHashMap<>();
+        this.strategyID = UUID.randomUUID();
```
jklamer: same with strategies

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
lucaspcram: So the logging framework will already show you the class and the line of code from which each message originates, eg.
```
2018-10-29 07:43:21 WARN  [main] ConcurrentResourceCache:63 - CacheID e18be71c-21f5-46fa-99cc-ca70bdf9fded: cache fetch failed, falling back to default fetcher...
```
That being said, it won't show the actual subclass implementation, should the log be triggered by a class that extended `ConcurrentResourceCache`. That could be remedied by storing `this.getClass().getName()` as a member variable, but it would also add more noise to the logs. Thoughts?

adahn6: I'll put in my vote for leaving as-is; subclasses that extend `ConcurrentResourceCache` can always add in class-specific logs for clarity if needed.

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
```diff
@@ -72,16 +61,51 @@ public ConcurrentResourceCache(final CachingStrategy cachingStrategy,
...
+        // Synchronize invalidation with the same lock used to fetch and cache. This prevents
+        // invalidation corruption.
+        synchronized (this)
+        {
+            this.cachingStrategy.invalidate();
+        }
+    }
+
+    @Override
+    public void invalidate(final URI resourceURI)
```
adahn6: 👍 to URI-specific invalidate!

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/GlobalNamespaceCachingStrategy.java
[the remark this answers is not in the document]
adahn6: IMO the name "global" should apply to as a large a space as is reasonably expectable, in this case system-wide. If we want a JVM specific cache, then that's worth either a toggle in the `GlobalNamespaceCachingStrategy` constructor or another subclass explicitly dedicated to that.

src/main/java/org/openstreetmap/atlas/utilities/caching/ConcurrentResourceCache.java
lucaspcram: I agree. I ended up adding this to the constructor of `ConcurrentResourceCache`:
```java
logger.info("Initialized cache {} with ID {}", this.getClass().getName(), this.cacheID);
```

It will print the subclass name if this constructor was being invoked thru a `super`. Should give @jklamer the search info he needs without adding more noise to every log message.

lucaspcram:
Still looking for ideas about what additional components can be tested!

Also, should there be some sort of retry code in `ConcurrentResourceCache.get(URI)`?
```java
public Optional<Resource> get(final URI resourceURI)
{
.
.
.
         // apply default fetcher if resource not in cache
         if (!cachedResource.isPresent())
         {
             synchronized (this)
             {
                 // Should this retry N times?
                 cachedResource = this.fetcher.apply(resourceURI);
             }
         }

         return cachedResource;
     }
```

src/main/java/org/openstreetmap/atlas/utilities/caching/strategies/GlobalNamespaceCachingStrategy.java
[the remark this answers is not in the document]
lucaspcram: Fair enough - that makes sense. I really don't like the intrinsically unsafe nature of this code, but I am not sure how we could make it safe given the design constraints.

PR #260, opened 2018-10-29
Relation member consistency when an OSM Node is both a Node and Point

MikeGost:
### Description:

In the case of a `Relation` that contains an OSM Node that is translated into both an `Atlas Point` and `Atlas Node`, we only added the `Atlas Point` as the member. This change ensures that both the `Point` and `Node` are added as members. This allows for backward compatibility and consistency with the `Atlas` data model. Thanks to @danielduhh and @Bentleysb for point this issue out!

~~**DNM**: Want to test at scale before signing off on the change~~

### Potential Impact:

This will have an impact on various atlas-checks that rely on `Relation`s or potentially `BigNode`s.

### Unit Test Approach:

I've created a unit test that verifies that behavior change. It fails before the change and passes after. Here's an example of the changes for [relation 578254](https://www.openstreetmap.org/relation/578254):

**Before**:
```
Relation: id=578254000000, [Members: 
				{Member: ID = -125508462000000, Type = EDGE, Role = from}, 
				{Member: ID = -125508462000000, Type = EDGE, Role = to}, 
				{Member: ID = 125508462000000, Type = EDGE, Role = from}, 
				{Member: ID = 125508462000000, Type = EDGE, Role = to}, 
				{Member: ID = 556595133000000, Type = POINT, Role = via}
```

**After**:
```
Relation: id=578254000000, [Members: 
				{Member: ID = 556595133000000, Type = NODE, Role = via}, 
				{Member: ID = -125508462000000, Type = EDGE, Role = from}, 
				{Member: ID = -125508462000000, Type = EDGE, Role = to}, 
				{Member: ID = 125508462000000, Type = EDGE, Role = from}, 
				{Member: ID = 125508462000000, Type = EDGE, Role = to}, 
				{Member: ID = 556595133000000, Type = POINT, Role = via}
```

Notice the extra `Node`, with the via role.

### Test Results:

Tested at-scale, with no impact on performance, BigNode creation or build functionality. 

------


danielduhh:
Sorry for the late response. I've verified these updates have fixed the previously broken `InvalidTurnRestriction` & `AddressPointMatch` integrity checks. Thanks the swift updates!
@MikeGost

MikeGost:
Awesome - thanks for following up @danielduhh!

PR #269, opened 2018-11-01
Tippecanoe & Line Delimited GeoJSON

hallahan:
### Description:

This pull request re-vamps how GeoJSON is made in atlas. This new line-delimited GeoJSON functionality serializes entity-by-entity, and provides a predicate and mutator for you to filter and modify features as they are serialized.

You can run `LineDelimitedGeoJsonConverter` to convert a directory full of atlases into their corresponding line-delimited GeoJSON. The CLI also concatenates that output into a single file.

You can run `TippecanoeExporter`, which subclasses `LineDelimitedGeoJsonConverter`. This will create your line-delimited GeoJSON, and then it will execute [tippecanoe](https://github.com/mapbox/tippecanoe) to ultimately create an MBTiles container file with MapboxVectorTiles. You can then use [mbview](https://github.com/mapbox/mbview) or [tileserver-gl](https://github.com/klokantech/tileserver-gl) to serve and visualize the data.

Another key part of this pull request is the ability to add the tippecanoe GeoJSON extension. This adds an extra object into your GeoJSON features called `tippecanoe`. In there, you can declare what layer the feature can be in, as well as the minimum zoom the feature should be in for the vector tiles. The `MinZoom` class provides some defaults, and ultimately we will want to tweak it further. Once we dial in what the various min zooms should be, I don't anticipate having to change that very often, so I haven't gone in and made a GeoJSON configuration, though that's a potential direction we could go in.

#### TODO

- [x] Unit Tests.
- [x] Include relation metadata in relation members' properties.

Note that the `LineDelimitedGeoJsonConverter` by default only includes positive ID'd features, as well as only relations that are multipolygons. The rationale is that you will be able to get a hold of the needed relation metadata from the members for everything but multipolygons. Also, edges that are one way can be derived by the `oneway=yes` tag, so finding the positive and negative feature is redundant and wastes space. You can change this predicate if you need these things.

### Potential Impact:

This PR does not change the functionality of the legacy GeoJSON serialization. We should do a future PR where we remove the old non-line-delimited GeoJSON generator, and simply adapt this code to also generate non-line-delimited GeoJSON.

The memory footprint of this approach is dramatically lower, as we only create `JsonObject`s on a feature-by-feature basis. The legacy GeoJSON generator would create a `JsonObject` of all of the entity in an atlas first, and this would then explode in size in memory, as this has no compression.

The new GeoJSON code handles relations a little bit better. Before, all relations were just GeoJSON features with point geometries pointing at null island. Now, if they are multipolygons, it properly encodes the multipolygon if it has all of the geometries it needs. Otherwise, it will encode the relation as a feature, with properties, and a polygon geometry that is the bounding box of that given relation. This way, you can see where it is on the map.

### Instructions

https://github.com/hallahan/atlas/blob/tippecanoe/src/main/java/org/openstreetmap/atlas/utilities/vectortiles/README.md

### Unit Test Approach:

Unit tests on most of the classes and critical touched pieces.

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtilsTest.java

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/geography/MultiPolygonTest.java

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/geography/PolyLineTest.java

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/geography/PolygonTest.java

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/geography/atlas/BareAtlasTest.java

https://github.com/hallahan/atlas/blob/tippecanoe/src/test/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoomTest.java

### Test Results:

Unit and Integration tests pass, unit tests are run by Travis.

I have also manually tested by running both CLIs and reviewed the output. For `TippecanoeExporter`, I have generated MBTiles from atlas files and explored the data in [mbview](https://github.com/mapbox/mbview).


src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -194,6 +196,32 @@ public String toDiffViewFriendlyString()
      */
     public abstract LocationIterableProperties toGeoJsonBuildingBlock();
 
+    /**
+     * A base method that creates a GeoJSON feature with properties from the tags. Note that it does
```
matthieun: So that sub-classing this is fool-proof, I would go the following way instead (names non-binding):
1. Create an abstract method called `asGeojsonFeature` that sub-classes will have to implement to compile
1. Have this `asGeoJsonFeature` method `protected`, and renamed to `asGeoJsonFeatureWithoutGeometry`, so it is only used by the sub classes and not used by customers by mistake.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LocationItem.java
```diff
@@ -25,6 +28,17 @@ protected LocationItem(final Atlas atlas)
         super(atlas);
     }
 
+    @Override
+    public JsonObject asGeoJsonFeature()
```
matthieun: Same type of comment here: Have a local `asGeojsonFeatureWithoutConnectivity` or something similar be protected, and that will force the sub-classes to use it when they implement the proper `asGeojsonFeature` that is mandated by `AtlasEntity`.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -57,6 +64,8 @@
             final Relation relation2) -> Long.compare(relation1.getIdentifier(),
                     relation2.getIdentifier());
 
+    private static final RelationOrAreaToMultiPolygonConverter MULTI_POLYGON_CONVERTER = new RelationOrAreaToMultiPolygonConverter();
```
matthieun: YaY!

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
```diff
@@ -1,5 +1,7 @@
 package org.openstreetmap.atlas.geography.geojson;
 
+import static org.openstreetmap.atlas.geography.geojson.GeoJsonUtils.locationsToCoordinates;
```
matthieun: Same comment here.

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
```diff
@@ -0,0 +1,102 @@
...
+import com.google.gson.JsonPrimitive;
+
+/**
+ * These are utility functions that well help you create GeoJSON!
+ *
+ * @author hallahan
+ */
+public final class GeoJsonUtils
+{
+
```
matthieun: whitespace

src/main/java/org/openstreetmap/atlas/vectortiles/MinZoom.java
```diff
@@ -0,0 +1,166 @@
+package org.openstreetmap.atlas.vectortiles;
```
matthieun: How about putting this in the `utilities` package?
```
org.openstreetmap.atlas.utilities.vectortiles
```

src/main/java/org/openstreetmap/atlas/vectortiles/MinZoom.java
```diff
@@ -0,0 +1,166 @@
...
+/**
+ * This utility class is where you set the minimum zoom for features based on their tags. The values
+ * you see here for various min zooms is inspired for what you see in the standard OpenStreetMap
+ * carto style. This is very loosely based off of the minimum zooms you see for various types of
+ * features. Note that there is definitely more work that needs to be done to refine our min zooms.
+ * https://github.com/gravitystorm/openstreetmap-carto
+ *
+ * @author hallahan
+ */
+final class MinZoom
```
matthieun: Avoid abbreviations: `MinimumZoom`

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
```diff
@@ -0,0 +1,148 @@
...
+import org.slf4j.LoggerFactory;
+
+/**
+ * This CLI will take a directory full of atlases and export it into an MBTiles file full of Mapbox
+ * vector tiles. It creates intermediary line-delimited GeoJSON, and then it ultimately drives
+ * tippecanoe to do the vector tile creation.
+ *
+ * @author hallahan
+ */
+public final class TippecanoeExporter extends LineDelimitedGeoJsonConverter
```
matthieun: Same comment here as for the Super class. How can we have this be split into two, one that runs the actual logic, and can save to a WritableResource, and another one that runs it command-style?

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
```diff
@@ -0,0 +1,148 @@
...
+
+    @Override
+    protected SwitchList switches()
+    {
+        return super.switches().with(MBTILES);
+    }
+
+    private boolean hasValidTippecanoe()
+    {
+        final CommandLine commandLine = CommandLine.parse("tippecanoe").addArgument("--version");
```
matthieun: Same, use `RunScript` here.

matthieun: Actually even better, the tippecanoe check should live in a shell script in the resource folder: `src/main/resources/path/to/package`

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeSettings.java
```diff
@@ -0,0 +1,50 @@
...
+        // Utility Class
+    }
+
+    static final DefaultArtifactVersion MIN_VERSION = new DefaultArtifactVersion("1.32.1");
+    static final String GEOJSON = "EVERYTHING.geojson";
+
+    static final String[] ARGS = new String[] { "-Z6", "-z14", "--generate-ids", "--read-parallel",
+            "--no-tile-size-limit", "--no-feature-limit" };
+
+    static final BiConsumer<AtlasEntity, JsonObject> JSON_MUTATOR = (atlasEntity, feature) ->
```
matthieun: This should be done using a Json configuration file in Resources, and a `ConfiguredTaggableFilter`. [Documentation here](https://github.com/osmlab/atlas/tree/4f20c144cdf5f46f59a4947410d1aa50d4be5211/src/main/java/org/openstreetmap/atlas/tags/filters#tag-filtering)!

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -740,7 +752,7 @@ public Angle maximumAngle()
 
     public Location middle()
     {
-        return offsetFromStart(Ratio.HALF);
+        return this.offsetFromStart(Ratio.HALF);
```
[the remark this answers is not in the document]
hallahan: Hi @matthieun , I don't use save actions at all. I think that's from @lucaspcram commit. Is it ok to keep this, or do you want it removed?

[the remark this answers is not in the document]
lucaspcram: Lol I knew this would come up. I screwed up a save action config at some point and those `this` keywords ended up everywhere without me noticing. The problem is that you can only remove them manually....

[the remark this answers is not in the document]
lucaspcram: Looks like there are about 30 places this happened. I don't mind fixing it myself (since it's my fault), but I am unsure how I can edit @hallahan 's PR. Ideas?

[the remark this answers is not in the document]
hallahan: @lucaspcram I added you as a collaborator to my fork, you can commit to the tippecanoe branch, which will update this pr.

https://github.com/hallahan/atlas/invitations

https://github.com/hallahan/atlas/tree/tippecanoe

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
hallahan: As we chatted in person, you can actually already save a given atlas as ld geojson to a WritableResource, which this CLI calls. the predicate and mutator in the settings class is now exposed as public, that way you can set this same thing up in Spark.

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeSettings.java
hallahan: We decided to leave this and make it public. Making this a JSON config totally makes sense though, could be a good optimization.

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -494,14 +497,22 @@ public Location first()
...
-                    + ", which is not between 0 and " + size());
+                    + ", which is not between 0 and " + this.size());
         }
         return this.points.get(index);
     }
 
+    public JsonObject asGeoJsonGeometry()
+    {
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty("type", "LineString");
```
MikeGost: Same

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -280,6 +283,20 @@ public boolean fullyGeometricallyEncloses(final Segment segment)
         return this.fullyGeometricallyEncloses(segment.middle());
     }
 
+    @Override
+    public JsonObject asGeoJsonGeometry()
+    {
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty("type", "Polygon");
```
MikeGost: Same

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -484,6 +487,32 @@ public void saveAsGeoJson(final WritableResource resource, final Predicate<Atlas
...
+    {
+        saveAsLineDelimitedGeoJsonFeatures(resource, item -> true, jsonMutator);
+    }
+
+    @Override
+    public void saveAsLineDelimitedGeoJsonFeatures(final WritableResource resource,
+            final Predicate<AtlasEntity> matcher,
+            final BiConsumer<AtlasEntity, JsonObject> jsonMutator)
+    {
+        final JsonWriter writer = new JsonWriter(resource);
```
MikeGost: How about putting this into a try-with-resources block?

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
...
+            "The number of threads to work on processing atlas shards.", Integer::valueOf,
+            Optionality.OPTIONAL, String.valueOf(DEFAULT_THREADS));
+
+    /**
+     * We only want positive edges, because the negative edge can be derived at the application
+     * level, and this encodes extraneous data that can be easily derived by the map viewer. For
+     * relations, we only want multipolygon relations, as the rest can be derived from their
+     * members.
+     */
+    private static final Predicate<AtlasEntity> PREDICATE = atlasEntity ->
```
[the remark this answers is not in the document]
MikeGost: +1

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
...
+    protected static final Switch<Boolean> OVERWRITE = new Switch<>("overwrite",
+            "Choose to automatically overwrite a GeoJSON file if it exists at the given path.",
+            Boolean::new, Optionality.OPTIONAL, "false");
+
+    private static final Switch<Integer> THREADS = new Switch<>("threads",
+            "The number of threads to work on processing atlas shards.", Integer::valueOf,
+            Optionality.OPTIONAL, String.valueOf(DEFAULT_THREADS));
+
+    /**
+     * We only want positive edges, because the negative edge can be derived at the application
```
MikeGost: Positive edges --> master edges :)

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
...
+     * level, and this encodes extraneous data that can be easily derived by the map viewer. For
+     * relations, we only want multipolygon relations, as the rest can be derived from their
+     * members.
+     */
+    private static final Predicate<AtlasEntity> PREDICATE = atlasEntity ->
+    {
+        // We only want positive atlas entities. No negative ids.
+        if (atlasEntity.getIdentifier() < 0)
```
MikeGost: I suggest relying on functionality such as check for type edge and use `isMasterEdgeIdentifier` instead of < 0 check

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -1,5 +1,7 @@
 package org.openstreetmap.atlas.geography.atlas.items;
 
+import static org.openstreetmap.atlas.geography.geojson.GeoJsonUtils.boundsToPolygonGeometry;
```
MikeGost: One more import to clean up

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -248,9 +257,55 @@ public LocationIterableProperties toGeoJsonBuildingBlock()
...
+        // themselves fine.
+        // The isMultiPolygon() method also includes boundaries, which we do not want.
+        if (Validators.isOfType(this, RelationTypeTag.class, RelationTypeTag.MULTIPOLYGON))
+        {
+            try
+            {
+                final MultiPolygon multiPolygon = MULTI_POLYGON_CONVERTER.convert(this);
+                geometry = multiPolygon.asGeoJsonGeometry();
+            }
+            // It seems like we get caught in this exception a lot! We don't ingest coastline
```
MikeGost: Nice commenting!

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * Creates a GeoJSON Polygon geometry from a bounds.
+     *
+     * @param bounds
+     *            A bounds.
+     * @return A GeoJSON Polygon geometry JsonObject.
+     */
+    public static JsonObject boundsToPolygonGeometry(final Rectangle bounds)
```
MikeGost: This looks like a possible use case for a `Converter`

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
```diff
@@ -0,0 +1,166 @@
...
+                return THIRTEEN;
+        }
+    }
+
+    static int building(final String tagValue)
+    {
+        return THIRTEEN;
+    }
+
+    static int highway(final String tagValue)
```
MikeGost: Will this change in the future? It almost makes sense to read this from a configuration file (with this implementation being the default) in case users want to override it..

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
```diff
@@ -0,0 +1,166 @@
...
+            default:
+                return THIRTEEN;
+        }
+    }
+
+    static int landuse(final String tagValue)
+    {
+        switch (tagValue)
+        {
+            case "basin":
```
MikeGost: If we're going to keep this in the code, might make sense to leverage the specific tag class enums here instead of hard-coding the values.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
hallahan: Yeah, I got started making a JSON resource file which I would then parse, but I found that this approach was nice having the switch with fallthroughts. I think you're right that this should be explored. What do you think about leaving it this way for this PR and then having that be part of the next sprint? Or would you rather have me re-do this now?

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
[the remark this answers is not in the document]
lucaspcram: Mostly fixed. Looks like in some places of `PolyLine`, there were already `this` prefixes before this PR, since now the diff is showing some as removed (Please just ignore the commit noise, I don't want to spend time figuring out which `this` prefixes I added and which were already present before the PR). Also, there are a couple other spots in this PR where you used a `this` prefix where it was not necessary. I have left those alone. I think it's fine.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
MikeGost: That sounds like a fine plan, maybe create a github issue to track this for some time in the future (next sprint or later).

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
hallahan: good idea, done!

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -56,6 +57,8 @@
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.google.gson.JsonObject;
+
 /**
  * @author matthieun
  * @author tony
```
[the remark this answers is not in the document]
hallahan: 😄 Done.

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
+package org.openstreetmap.atlas.geography.atlas.geojson;
+
+import static org.openstreetmap.atlas.geography.atlas.AtlasResourceLoader.IS_ATLAS;
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
...
+import org.slf4j.LoggerFactory;
+
+import com.google.gson.JsonObject;
+
+/**
+ * This CLI takes a directory of atlas files and turns them into line-delimited GeoJSON.
+ *
+ * @author hallahan
+ */
+public class LineDelimitedGeoJsonConverter extends Command
```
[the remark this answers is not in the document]
hallahan: As we chatted yesterday, you can directly call `saveAsLineDelimitedGeoJsonFeatures` in atlas. You don't have to use this command. This just makes it convenient to turn a directory full of atlases into ld GeoJSON

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
hallahan: Good call, done.

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilder.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -194,6 +196,34 @@ public String toDiffViewFriendlyString()
...
+        final JsonObject properties = new JsonObject();
+        getTags().forEach(properties::addProperty);
+        properties.addProperty("identifier", getIdentifier());
+        properties.addProperty("osmIdentifier", getOsmIdentifier());
+        properties.addProperty("itemType", String.valueOf(getType()));
+
+        final Optional<String> shardName = getAtlas().metaData().getShardName();
+        shardName.ifPresent(shard -> properties.addProperty("shard", shard));
+
+        feature.add("properties", properties);
```
jklamer: constant here too

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -194,6 +196,34 @@ public String toDiffViewFriendlyString()
...
+    public abstract JsonObject asGeoJsonFeature();
+
+    /**
+     * A method that creates a GeoJSON feature with properties from the tags. Note that it does not
+     * contain the geometry, and the subclasses should call this and then fill in the geometry from
+     * within their implementation of asGeoJsonFeature().
+     *
+     * @return A GeoJSON Feature with properties but no geometry.
+     */
+    protected JsonObject asGeoJsonFeatureWithPropertiesOnly()
```
jklamer: A geojson feature without geometry seems incomplete. I think getGeojsonProperties() would be cleaner.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java
```diff
@@ -37,6 +39,18 @@ protected Edge(final Atlas atlas)
         super(atlas);
     }
 
+    @Override
+    public JsonObject asGeoJsonFeature()
+    {
+        final JsonObject feature = super.asGeoJsonFeature();
```
jklamer: I feel like this should override ```getGeojsonProperties()``` referenced above.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LineItem.java
```diff
@@ -125,4 +134,15 @@ public LocationIterableProperties toGeoJsonBuildingBlock()
...
         return new GeoJsonBuilder.LocationIterableProperties(getRawGeometry(), tags);
     }
+
+    @Override
+    public JsonObject asGeoJsonFeature()
+    {
+        final JsonObject feature = asGeoJsonFeatureWithPropertiesOnly();
+
+        final JsonObject geometry = asPolyLine().asGeoJsonGeometry();
+        feature.add("geometry", geometry);
```
jklamer: constant here.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Node.java
```diff
@@ -29,6 +33,34 @@ public long absoluteValence()
         return this.connectedEdges().size();
     }
 
+    @Override
+    public JsonObject asGeoJsonFeature()
+    {
+        final JsonObject feature = super.asGeoJsonFeature();
+        final JsonObject properties = feature.getAsJsonObject("properties");
```
jklamer: same ad before. This function could just be overriding the ```getGeojsonProperties()``` function.

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    {
+        final JsonObject geometry = new JsonObject();
+
+        final Location lowerLeft = bounds.lowerLeft();
```
jklamer: you can replace the following with a bounds.closedLoop()

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    /**
+     * An iterable of locations will turn into a JsonArray of Longitude, Latitude coordinates.
+     *
+     * @param locations
+     *            An iterable of locations
+     * @return A JsonArray of Longitude, Latitude coordinates.
+     */
+    public static JsonArray locationsToCoordinates(final Iterable<Location> locations)
```
jklamer: I'd prefer you have a separate function for Polygons, that does the geojson specific wrapping for the user. You manually wrap the Polygons coordinates in a lot of places. And have to remember to use ```closedLooop()``` all of which can be handled in this class.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -478,9 +482,36 @@ public void saveAsGeoJson(final WritableResource resource)
...
+    {
+        saveAsLineDelimitedGeoJsonFeatures(resource, item -> true, jsonMutator);
+    }
+
+    @Override
+    public void saveAsLineDelimitedGeoJsonFeatures(final WritableResource resource,
+            final Predicate<AtlasEntity> matcher,
+            final BiConsumer<AtlasEntity, JsonObject> jsonMutator)
     {
         final JsonWriter writer = new JsonWriter(resource);
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -280,6 +283,20 @@ public boolean fullyGeometricallyEncloses(final Segment segment)
...
+
+        final JsonArray coordinates = new JsonArray();
+        geometry.add("coordinates", coordinates);
```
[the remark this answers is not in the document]
hallahan: How about I make a function in the GeoJsonUtils that creates a JsonObject that is the geometry?

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
hallahan: I looked at this, and the expression would be several added function calls, because, i'd have to first check that it's edge type, than cast it to edge, then call that function, which really, is just looking at if the id is < 0...

Let me see if it has any impact on performance...

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,220 @@
...
+            logger.info("Concatenated to {} in {}", EVERYTHING, time.elapsedSince());
+        }
+        catch (final IOException ioException)
+        {
+            logger.error("Unable to concatenate the output line-delimited GeoJSON files.",
+                    ioException);
+        }
+    }
+
+    private static List<File> fetchAtlasFilesInDirectory(final Path directory)
```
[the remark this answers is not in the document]
hallahan: Because we don't want to load a MultiAtlas, we're actually doing shard by shard in parallel.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
hallahan: Duh, good point, will do. Then I can have a feature function in GeoJsonUtil that can take a geometries and a properties and make the feature.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java
hallahan: LineItem does indeed call that now. That is super. The super makes most of what we need, here, we're adding the start and end nodes to the properties.

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -85,6 +90,40 @@ public MultiPolygon(final MultiMap<Polygon, Polygon> outerToInners)
...
 
+    /**
+     * Creates a JsonObject with GeoJSON geometry representing this multi-polygon.
+     *
+     * @return A JsonObject with GeoJSON geometry
+     */
+    public JsonObject asGeoJsonGeometry()
+    {
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty("type", "MultiPolygon");
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/Location.java
```diff
@@ -245,6 +249,16 @@ public Distance equirectangularDistanceTo(final Location that)
         return Distance.AVERAGE_EARTH_RADIUS.scaleBy(Math.sqrt(xAxis * xAxis + yAxis * yAxis));
     }
 
+    public JsonObject asGeoJsonGeometry()
+    {
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty("type", "Point");
+
+        geometry.add("coordinates", coordinate(this));
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/MultiPolygon.java
```diff
@@ -85,6 +90,41 @@ public MultiPolygon(final MultiMap<Polygon, Polygon> outerToInners)
...
+        final JsonArray polygons = new JsonArray();
+        geometry.add("coordinates", polygons);
+
+        for (final Map.Entry<Polygon, List<Polygon>> entry : outerToInners.entrySet())
+        {
+            final Polygon outer = entry.getKey();
+            final List<Polygon> inners = entry.getValue();
+
+            final JsonArray polygon = new JsonArray();
+            final JsonArray outerRingCoordinates = locationsToCoordinates(outer.closedLoop());
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/PolyLine.java
```diff
@@ -249,6 +252,14 @@ public GeoJsonObject asGeoJson()
         return asGeoJson(geometries);
     }
 
+    public JsonObject asGeoJsonGeometry()
+    {
+        final JsonObject geometry = new JsonObject();
+        geometry.addProperty("type", "LineString");
```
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LineItem.java
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Node.java
hallahan: Addressed ^^^.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LocationItem.java
hallahan: Revamped that stuff, should be good

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
lucaspcram: @hallahan to solve your problem, try adding the new API we discussed, that calls:
```java
public Process exec(String[] cmdarray, String[] envp, File dir);
```
This will handle tokenizing correctly. @matthieun I think this is a good way to use `RunScript` while still allowing @hallahan to get his code working.

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,222 @@
...
+    private void concatenate(final Path geojsonDirectory)
+    {
+        final Time time = Time.now();
+        final String directory = geojsonDirectory.toString();
+
+        // https://stackoverflow.com/questions/5080109/how-to-execute-bin-sh-with-commons-exec
+        final String cat = String.format("cat '%s/'*.geojson > '%s/'%s", directory, directory,
+                EVERYTHING);
+
+        final CommandLine commandLine = CommandLine.parse("bash").addArgument("-c", false)
```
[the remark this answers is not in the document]
hallahan: I've attempted to use `RunScript`, however, it does not work properly. Here's some very basic usage I'm attempting:

https://github.com/hallahan/atlas/blob/b1587a61dac8d53d938c312e62723dfa70551130/src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java#L221-L227

```java
    private void concatenate2(final Path geojsonDirectory)
    {
        final String directory = geojsonDirectory.toString();
        final String command = String.format("bash -c \"cat '%s/'*.geojson > '%s/'%s\"", directory, directory, EVERYTHING);

        RunScript.run(command);
    }
```

I get the following error:

```
[main] ERROR org.openstreetmap.atlas.utilities.runtime.Command - Command execution failed.
org.openstreetmap.atlas.exception.CoreException: Non-Zero return value 1 when running script "cat '/Users/n/viz-data/ECU/'*.geojson > '/Users/n/viz-data/ECU/'EVERYTHING.geojson".
	at org.openstreetmap.atlas.utilities.runtime.RunScript.run(RunScript.java:93)
	at org.openstreetmap.atlas.utilities.runtime.RunScript.run(RunScript.java:25)
	at org.openstreetmap.atlas.geography.atlas.geojson.LineDelimitedGeoJsonConverter.concatenate2(LineDelimitedGeoJsonConverter.java:226)
	at org.openstreetmap.atlas.geography.atlas.geojson.LineDelimitedGeoJsonConverter.onRun(LineDelimitedGeoJsonConverter.java:150)
	at org.openstreetmap.atlas.utilities.runtime.Command.execute(Command.java:338)
	at org.openstreetmap.atlas.utilities.runtime.Command.run(Command.java:282)
	at org.openstreetmap.atlas.geography.atlas.geojson.LineDelimitedGeoJsonConverter.main(LineDelimitedGeoJsonConverter.java:115)
```

However, when I execute it in my terminal, this command works fine.

```
bash -c cat '/Users/n/viz-data/ECU/'*.geojson > '/Users/n/viz-data/ECU/'EVERYTHING.geojson
```

After debugging a bit with @lucaspcram , we noticed that your code (RunScript) is running:

```
final Process process = Runtime.getRuntime().exec(script, env);
```

If you drill in, you'll notice that exec is using a string tokenizer that is breaking everything into an array or strings from spaces and returns:

```java
public Process exec(String command, String[] envp, File dir)
        throws IOException {
        if (command.length() == 0)
            throw new IllegalArgumentException("Empty command");

        StringTokenizer st = new StringTokenizer(command);
        String[] cmdarray = new String[st.countTokens()];
        for (int i = 0; st.hasMoreTokens(); i++)
            cmdarray[i] = st.nextToken();
        return exec(cmdarray, envp, dir);
    }
```

If we were to make this work, we'd actually want the more exact 

```java
public Process exec(String[] cmdarray, String[] envp, File dir)
        throws IOException {
        return new ProcessBuilder(cmdarray)
            .environment(envp)
            .directory(dir)
            .start();
    }
```

I can try this out, but I still think apache commons exec is a better option. Java's out of the box exec is notorious for being brittle, and I don't think this simple RunScript wrapper is going to serve us well.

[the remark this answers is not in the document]
hallahan: In addition, there is an issue with `/Users/matthieun` hard coded in the unit tests. The only other unit test is calling java -version.

https://github.com/osmlab/atlas/issues/271

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
hallahan: Ok, I'll spend more time trying to make this work.

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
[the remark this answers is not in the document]
lucaspcram: My opinion, regardless of if @hallahan decides to use the apache libraries (I have no skin in this game), I would like to see an API addition to `RunScript` included in this PR that calls the more robust `exec(String[] cmdarray, String[] envp, File dir)`.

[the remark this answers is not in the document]
hallahan: Can this become a problem?

https://github.com/hallahan/atlas/blob/b1587a61dac8d53d938c312e62723dfa70551130/src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java#L73

```java
// Wait for the monitors
            printer.waitForCompletion(Duration.ONE_SECOND);

            if (monitors != null && !monitors.isEmpty())
            {
                for (final RunScriptMonitor monitor : monitors)
                {
                    monitor.waitForCompletion(Duration.ONE_SECOND);
                }

                Streams.close(standardOut);
                Streams.close(standardErr);
            }
```

Does this mean we have to wait a second extra every time we want to exec something?

lucaspcram:
100 🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉🎉

src/main/java/org/openstreetmap/atlas/vectortiles/TippecanoeExporter.java
hallahan: Ok, I am now using only RunScript, no more apache commons exec.

src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
```diff
@@ -20,20 +20,20 @@
 {
     private static final Logger logger = LoggerFactory.getLogger(RunScript.class);
 
-    public static void run(final String script)
+    public static void run(final String[] commandArray)
```
matthieun: Please keep the single String option as well, so as to make it non breaking.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeGeoJsonExtension.java
```diff
@@ -0,0 +1,39 @@
+package org.openstreetmap.atlas.utilities.vectortiles;
+
+import com.google.gson.JsonObject;
+
+/**
+ * TippecanoeGeoJsonExtension fills in a JsonObject called "tippecanoe" to your GeoJSON feature.
+ * Here you can specify things like minzoom and layer name for the given feature.
+ * https://github.com/mapbox/tippecanoe/blob/master/README.md#geojson-extension
```
matthieun: Maybe make that link point to a tag, or a specific commit, to make sure it does not go away.

src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
hallahan: done

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeGeoJsonExtension.java
hallahan: done

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
hallahan: Unfortunately, you need to have Strings as constants at compile time for them to work in switches...

hallahan: For example, say I do this

```java
private static final String HOSPITAL = AmenityTag.HOSPITAL.name().toLowerCase();
    private static final String MOTORWAY = HighwayTag.MOTORWAY.getTagValue();
    private static final String TRUNK = HighwayTag.TRUNK.getTagValue();
    private static final String PRIMARY = HighwayTag.PRIMARY.getTagValue();
```
and then have

```java
    static int amenity(final String tagValue)
    {
        switch (tagValue)
        {
            case HOSPITAL:
                return TWELVE;
            default:
                return THIRTEEN;
        }
    }
```

Well, that doesn't work... Won't compile...

jklamer: Static block?

Also why go to lower case?

hallahan: Because the actual tag values are lower case, and the enum itself is caps. If you look at .getTagValue, it's actually doing that, plus putting it in intern. The only thing with that is, only some of the tag enums have that implemented...

hallahan:
Ok, I've revamped how MinimumZoom works. It now takes a JSON configuration from resources.

Right now we need more test coverage.

cc/ @jklamer

src/main/resources/org/openstreetmap/atlas/utilities/vectortiles/minimum-zooms.json
```diff
@@ -0,0 +1,93 @@
+[
```
lucaspcram: I like this way of doing it.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
```diff
@@ -0,0 +1,116 @@
...
+import java.util.Map;
+
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+/**
+ * This singleton class is where you get the minimum zoom for features' tags based on a JSON
+ * configuration. The values you see in the minimum-zoom.json resourse file is inspired for what you
```
lucaspcram: Super super tiny nit: resourse -> resource

src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
```diff
@@ -20,20 +20,30 @@
...
+        run(new String[] { command });
     }
 
-    public static void run(final String script, final List<RunScriptMonitor> monitors)
+    public static void run(final String command, final List<RunScriptMonitor> monitors)
+    {
+        run(new String[] { command }, monitors);
+    }
+
+    public static void run(final String[] commandArray)
```
lucaspcram: This was a much needed addition. Getting exec code right is tricky - this should help avoid future gotchas.

hallahan:
Ready for merge!

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
```diff
@@ -0,0 +1,221 @@
...
+
+        if (overwrite)
+        {
+            try
+            {
+                FileUtils.deleteDirectory(geojsonDirectory.toFile());
+            }
+            catch (final IOException noDelete)
+            {
+                logger.warn(
```
MikeGost: Nit: please add `noDelete` as the last parameter in the logger, this will ensure the stack trace is available in the logs.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -194,6 +199,66 @@ public String toDiffViewFriendlyString()
...
+                        // We shouldn't get here, but if we do, let's know about it in the data...
+                        memberObject.addProperty(GeoJsonUtils.IDENTIFIER, "MISSING");
+                    }
+
+                    // Sometimes a member doesnt have a role. That's normal.
+                    final String role = member.getRole();
+                    if (role != null)
+                    {
+                        // And sometimes the role is "", but we should keep it that way...
+                        memberObject.addProperty("role", role);
```
MikeGost: Do we want to add the type (Edge, Node, etc.) of member as well or does that not matter here?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LineItem.java
```diff
@@ -24,6 +29,8 @@
 {
     private static final long serialVersionUID = -2053566750957119655L;
 
+    private static final Logger logger = LoggerFactory.getLogger(LineItem.class);
```
MikeGost: Nit: is this used anywhere? If not, suggest removing.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LocationItem.java
```diff
@@ -66,6 +76,9 @@ public LocationIterableProperties toGeoJsonBuildingBlock()
         tags.put("osmIdentifier", String.valueOf(getOsmIdentifier()));
         tags.put("itemType", String.valueOf(getType()));
 
+        final Optional<String> shardName = getAtlas().metaData().getShardName();
```
MikeGost: Why is this done in the `bounds()` call? Doesn't seem consistent with the other classes.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -248,9 +255,53 @@ public LocationIterableProperties toGeoJsonBuildingBlock()
...
+            // features, so polygons that touch coastlines will fail. It's good to include
+            // the exception in the data, along with the bounding box. That way, we can
+            // notice the problem when browsing the map.
+            catch (final CoreException exception)
+            {
+                final String message = String.format("%s - %s",
```
MikeGost: Do we want to log it as well?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
jklamer: ++1

src/main/java/org/openstreetmap/atlas/geography/atlas/geojson/LineDelimitedGeoJsonConverter.java
hallahan: done 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
hallahan: done 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LineItem.java
hallahan: done 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LocationItem.java
hallahan: This was an addition I added to the legacy GeoJSON code. Taking this out, since it doesn't really matter for this PR.

src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
```diff
@@ -688,6 +691,32 @@
...
+     *
+     * @param resource
+     *            The resource to write to
+     * @param matcher
+     *            The matcher to consider
+     * @param jsonMutator
+     *            The callback function that will let you change what is in the Feature's JSON.
+     */
+    void saveAsLineDelimitedGeoJsonFeatures(WritableResource resource,
+            Predicate<AtlasEntity> matcher, BiConsumer<AtlasEntity, JsonObject> jsonMutator);
```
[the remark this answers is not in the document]
hallahan: I disagree on this one. the `JsonObject jsonMutator` object is being mutated, as the name suggests. We pass this object by reference, and we do things to it. This is a very common pattern, both in Java and C++. If you were to also return it in a function, that would imply that you're copying or making a new object based on the input parameters, which we are not. Doing that would be a poor choice, as it would hurt performance and do an unnecessary copy on something that is being used exclusively for this given purpose.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -194,6 +199,66 @@ public String toDiffViewFriendlyString()
...
+            final JsonArray relationsArray = new JsonArray();
+            properties.add("relations", relationsArray);
+            for (final Relation relation : relations)
+            {
+                final JsonObject relationObject = relation.geoJsonProperties();
+                relationsArray.add(relationObject);
+                final RelationMemberList members = relation.members();
+                final JsonArray membersArray = new JsonArray();
+                relationObject.add("members", membersArray);
+                for (final RelationMember member : members)
```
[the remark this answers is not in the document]
hallahan: My thought is that if you select a given relation member, which will have an associated geometry, having this encoded would allow a map viewer to select all of the members of a relation in the map viewer. If you do not do this, you will not have that ability.

We are not actually putting in the entity of the members, just the id, role, and type, as you would see in standard OSM XML. 

So, you do get that trade off.

I have tested out this change, and I'm not seeing a difference in how fast the vector tiles render.

I do see the disadvantage that there's too much stuff in the popup of the properties in mbview, but a custom map viewer would address this. 

I can take this out, but there is that trade off. You will no longer get the data that is a relation in your application.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * This singleton class is where you get the minimum zoom for features' tags based on a JSON
+ * configuration. The values you see in the minimum-zoom.json resource file is inspired for what you
+ * see in the standard OpenStreetMap carto style. This is very loosely based off of the minimum
+ * zooms you see for various types of features. See the README.md in this package for more
+ * details...
+ *
+ * @author hallahan
+ */
+enum MinimumZoom
```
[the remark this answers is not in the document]
hallahan: It is configurable. There is not a way to determine it at runtime, but that would be a trivial change. This feature was requested, so I did it. It's cleaner to deal with this way.

src/main/java/org/openstreetmap/atlas/streaming/writers/JsonWriter.java
```diff
@@ -69,9 +69,14 @@ public void write(final JsonObject object)
     public void writeLine(final JsonObject object)
     {
         final String value = object.toString();
+        writeLine(value);
+    }
+
+    public void writeLine(final String stringValue)
```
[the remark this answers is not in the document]
hallahan: yes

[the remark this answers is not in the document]
hallahan: Since I don't need to do that, taking out...

matthieun:
Looks like `org.openstreetmap.atlas.geography.atlas.BareAtlasTest > testSaveAsLineDelimitedGeoJsonFeatures FAILED`: https://travis-ci.org/osmlab/atlas/builds/454064409

hallahan:
Yay testing! It failed, as it should have, because I have added the entity type to the relation members, as was requested. I will up date the test.

src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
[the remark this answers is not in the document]
jklamer: The BiFunction wouldn't necessitate a new object be made. This isn't a blocker for me but the pattern does make more compact implementations possible and fits with the Java8 stream usage patterns.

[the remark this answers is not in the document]
lucaspcram: I agree with @hallahan here. From the documentation of `BiConsumer`:
```java
/**
 * Represents an operation that accepts two input arguments and returns no
 * result.  This is the two-arity specialization of {@link Consumer}.
 * Unlike most other functional interfaces, {@code BiConsumer} is expected
 * to operate via side-effects.
.
.
.
@FunctionalInterface
public interface BiConsumer<T, U> {
.
.
.
```
When you see a `Consumer`, you know that you're dealing with mutation via pass-by- reference semantics - side effects are a well-documented (and expected) behavior. Since `Atlas` and its entities are immutable, it's implied here that the `JsonObject` is the subject of mutation.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -479,9 +483,37 @@ public void saveAsGeoJson(final WritableResource resource)
...
+    }
+
+    @Override
+    public void saveAsLineDelimitedGeoJsonFeatures(final WritableResource resource,
+            final Predicate<AtlasEntity> matcher,
+            final BiConsumer<AtlasEntity, JsonObject> jsonMutator)
+    {
+        try (JsonWriter writer = new JsonWriter(resource))
+        {
+            for (final AtlasEntity entity : entities())
```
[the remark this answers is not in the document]
lucaspcram: @jklamer is right here. `entities(matcher)` is going to iterate only once, but use the `matcher` predicate to skip over non-matching results of the underlying `Iterator` before they are actually returned. See the `filterTranslate` implementation in `Iterables` for clarification.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/MinimumZoom.java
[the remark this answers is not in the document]
lucaspcram: I requested this approach. It prevents us from needing to hardcode configuration data.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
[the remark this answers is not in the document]
lucaspcram: Considering this is a WIP PR, I think it's fine to leave it in for now - just to see how performance will look at scale, as well as test user experience regarding map viewing. This can always be adjusted/made configurable later.

src/main/java/org/openstreetmap/atlas/geography/geojson/GeoJsonUtils.java
```diff
@@ -0,0 +1,129 @@
...
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
+
+/**
+ * These are utility functions that well help you create GeoJSON!
+ *
+ * @author hallahan
+ */
+public final class GeoJsonUtils
```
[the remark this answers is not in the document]
lucaspcram: What would stop someone from calling `locationsToCoordinates(polygon.closedLoop())` in the future, despite the additional API? Or am I missing something?

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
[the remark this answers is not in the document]
hallahan: done

src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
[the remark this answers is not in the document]
jklamer: So any time I select one member of a relation all the other members will get selected too? Or will the map viewer give me an option?

hallahan:
Aha! I found the bug!

The order of the relations that a given relation a member is a part of is non-determinate. That's because the `.relations()` method on an `AtlasEntity` returns a `Set`. A set may or may not be ordered, and it looks like this set is NOT ordered.

If you look in `PackedAtlas`, you'll see that this is how we get the set of relations:

```java
    private Set<Relation> itemRelations(final long[] relationIndices)
    {
        final Set<Relation> result = new HashSet<>();
        if (relationIndices == null)
        {
            return result;
        }
        for (final long relationIndex : relationIndices)
        {
            result.add(new PackedRelation(this, relationIndex));
        }
        return result;
    }
```

A hash set is not ordered! I think that's ok, because this is a derived relationship. However, it's no good for testing, and non-determinate behavior can confuse people and code in various circumstances (like this one)!

So, how about we just turn that into a `LinkedHashSet`?

lucaspcram:
@hallahan sounds ok to me. Let's get this fixed!

src/main/java/org/openstreetmap/atlas/geography/Location.java
```diff
@@ -507,18 +515,17 @@ public Location shiftAlongGreatCircle(final Heading initialHeading, final Distan
...
-        final double longitude2 = longitude1 + Math.atan2(Math.sin(bearing)
-                * Math.sin((double) distance.asMillimeters()
-                        / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
-                * Math.cos(latitude1), Math
-                        .cos((double) distance.asMillimeters()
+        final double longitude2 = longitude1 + Math.atan2(
+                Math.sin(bearing)
+                        * Math.sin(distance.asMillimeters()
                                 / Distance.AVERAGE_EARTH_RADIUS.asMillimeters())
+                        * Math.cos(latitude1),
```
matthieun: Any reason for that update here?

hallahan:
Before you merge, let me run a larger data set and take a look one last time for sanity.

hallahan:
Ok, I tried out a 2.35GB area of atlas, creating vector tiles with TippecanoeExporter.java. It took 19.5 minutes on my Macbook Pro. The contents of what we're seeing with relations is now what was described above ^^^.

Ready for review or merge.

PR #272, opened 2018-11-14
Fixed inconsistent geometry

adahn6:
### Description:

Under certain conditions, geometry across shards was being constructed in both shards but varied slightly due to precision and rounding logic. This PR updates the slicing operation in a number of ways intended to address this problem. 

First, the logic around point creation during slicing has been updated. If the Atlas contains a point at 7 digits of precision matching the sliced coordinate, that point is used. Otherwise, the point is reduced to 6 digits of prevision, and then either an existing Atlas point at that location is used, or a new TemporaryPoint is constructed. 

Additionally, the cache of coordinates to new points has been reconfigured to maintain a per-OSM-identifier cache at the 6 digit of precision level, with a secondary global 7-digit precision map. 

The result logic tweaks the weight of various components of slicing. Existing Atlas points are now prioritized. If there is an existing Atlas point matching the 7-digit precision coordinate, it is always used; similarly, when scaling down to 6-digit precision for a coordinate, the Atlas points are again checked at that newly scaled location. This limits the number of TemporaryPoints made and prevents excess duplicate Points in the final sliced Atlas.

Additionally, this substantially increases determinism in the final output geometry. Previously, there was a corner case involving two shards processing the same geometry differently caused by a caching collision. Since the cache wasn't organizing by OSM identifier, shapes not shared across shards could contribute their 6-digit precision coordinates to the cache, causing different cache contents during the slicing operations for different shards. 

### Potential Impact:

Should fix geometry related errors in Atlas verifier. Otherwise, will round all SyntheticBoundaryNodes created by slicing to 6-digits of prevision, causing slightly different shapes that previous versions.

### Unit Test Approach:

None. Any suggestions for unit tests are greatly appreciated-- this is a complex issue that's hard to design simple cases around.

### Test Results:

Basic sanity tests passed, Atlas verifier no longer shows geometry issues on my runs. 

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
@@ -98,8 +98,11 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
      * raw Atlas, and adds the points the list of points for the line.
      *
      * @param location
+     *            The location for the point(s)
      * @param rawAtlasPoints
+     *            The points from the raw Atlas
      * @param line
+     *            The List representing the points for the line to add the points to
      */
     private void addRawAtlasPointsToLine(final Location location,
             final Iterable<Point> rawAtlasPoints, final List<Long> line)
@@ -289,12 +292,10 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
 
                             // Check the cache for this coordinate -- if it already exists, we'll
                             // use it. Otherwise, scale to 6-digits and continue
-                            if (getCoordinateToPointMapping().containsCoordinate(coordinate,
-                                    line.getOsmIdentifier()))
+                            if (getCoordinateToPointMapping().containsCoordinate(coordinate))
                             {
-                                newLineShapePoints
-                                        .add(getCoordinateToPointMapping().getPointForCoordinate(
-                                                coordinate, line.getOsmIdentifier()));
+                                newLineShapePoints.add(getCoordinateToPointMapping()
+                                        .getPointForCoordinate(coordinate));
                             }
                             else
                             {
@@ -319,7 +320,7 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
                                                     coordinate),
                                             pointIdentifierFactory, pointTags);
                                     getCoordinateToPointMapping().storeMapping(coordinate,
-                                            line.getOsmIdentifier(), newPoint.getIdentifier());
+                                            newPoint.getIdentifier());
                                     newLineShapePoints.add(newPoint.getIdentifier());
                                     this.slicedPointAndLineChanges.createPoint(newPoint);
                                 }
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
@@ -93,6 +93,28 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
     }
 
     /**
+     * Takes a location, a list of points from the raw Atlas at that location, and a list of point
+     * IDs. Ensures the tags for the point are updated, ensures the points are not removed from the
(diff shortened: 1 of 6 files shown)


jklamer:
😱

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -41,16 +43,20 @@
...
     protected static final JtsPolyLineConverter JTS_POLYLINE_CONVERTER = new JtsPolyLineConverter();
     protected static final JtsLocationConverter JTS_LOCATION_CONVERTER = new JtsLocationConverter();
     protected static final JtsLinearRingConverter JTS_LINEAR_RING_CONVERTER = new JtsLinearRingConverter();
     protected static final MultiplePolyLineToPolygonsConverter MULTIPLE_POLY_LINE_TO_POLYGON_CONVERTER = new MultiplePolyLineToPolygonsConverter();
 
+    // JTS precision handling
+    private static final Integer SEVEN_DIGIT_PRECISION_SCALE = 10_000_000;
+    private static final PrecisionModel PRECISION_MODEL = new PrecisionModel(
+            SEVEN_DIGIT_PRECISION_SCALE);
+    protected static final PrecisionReducerCoordinateOperation PRECISION_REDUCER = new PrecisionReducerCoordinateOperation(
```
MikeGost: Nice cleanup!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -632,6 +641,8 @@ private void mergeOverlappingClosedMembers(final Relation relation,
             final int outerIndex = entry.getKey();
             final List<Integer> innerIndices = entry.getValue();
 
+            final CountrySlicingIdentifierFactory lineIdentifierGenerator = new CountrySlicingIdentifierFactory(
```
MikeGost: Can you add a quick comment here, I think it's important to call out that the outerIndex is now being used for seeding.

adahn6: done!

adahn6:
hold off on merging this, actually, gonna test a simplification that I just realized was possible

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -92,6 +92,28 @@ public Atlas slice()
         return atlasWithSlicedWaysAndPoints;
     }
 
+    /**
+     * Takes a location, a list of points from the raw Atlas at that location, and a list of point
+     * IDs. Ensures the tags for the point are updated, ensures the points are not removed from the
+     * raw Atlas, and adds the points the list of points for the line.
+     *
+     * @param location
```
matthieun: I believe those need descriptions too. I thought Checkstyle caught this... will have to look into that.

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RelationSlicingTest.java
```diff
@@ -130,7 +130,7 @@ public void testSimpleMultiPolygonWithHoleSpanningTwoCountries()
         final Atlas slicedAtlas = RAW_ATLAS_SLICER.slice(rawAtlas);
         new ComplexBuildingFinder().find(slicedAtlas).forEach(System.out::println);
 
-        Assert.assertEquals(23, slicedAtlas.numberOfPoints());
+        Assert.assertEquals(25, slicedAtlas.numberOfPoints());
```
matthieun: Covert change :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
adahn6: fixed!

adahn6:
I updated this PR with a pretty substantial simplification-- it turns out that 6-digit precision is enough to guarantee determinism even without the per-OSM-shape cache of points. So the remaining logic is a lot simpler now. Reran sanity tests etc and everything still looks good.

PR #274, opened 2018-11-15
Fix missing ISO_COUNTRY_CODE tags for RawAtlasRelationSlicer edge case

lucaspcram:
### Description:
Previously, the `RawAtlasRelationSlicer` would fail to assign a country code to any relation which only had other relations as its members. This is because `RawAtlasRelationSlicer.updateAndSplitRelation` looks at the assigned country codes of the relation members in order to determine which code to assign the relation. In the case of a relation with only relations as members, those members would not have country codes in the `partiallySlicedRawAtlas`, since the codes were just assigned by the current `RawAtlasRelationSlicer` and so were only present in our changeset!

The fix is simple - if we hit the entity in the `partiallySlicedRawAtlas` but it does not have an `ISO_COUNTRY_CODE` tag, then we need to also check our current changeset for that entity's identifier, since the tag may be present there.

This bug has an interesting history. Before the raw flow, it was prevalent. Then, when @MikeGost implemented raw flow, it appeared to be fixed. However, we were bewildered when it suddenly returned, for seemingly no reason. That reason is now clear. Until #216 the `WaySectionProcessor` would drop any relation which had at least one relation as a member. So the raw flow actually did not fix the `N/A` bug - rather it introduced a new bug which masked the `N/A` relations in the final sliced and sectioned atlas!

Now we can sigh relief - between this fix and the awesome work from @adahn6 in #272, the raw flow is becoming extremely rock-solid.

### Potential Impact:

We should no longer be seeing relations with `N/A` country codes. Any downstream users who were filtering out `N/A` features previously will start seeing lots of new relations.

### Unit Test Approach:

A simple unit test demonstrating the issue has been added, `RelationsSlicingTest.testSlicingOnRelationWithOnlyRelationsAsMembers`.

### Test Results:
The unit test mentioned above fails without the fix and passes with the fix. In addition to this unit test, I have tested this fix locally with some .osm.pbf files, and confirmed that the `N/A` tag issue totally resolves itself.

------


adahn6:
Look great! thanks for getting this addressed so quickly 💯

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RelationSlicingTestRule.java
```diff
@@ -221,6 +221,19 @@
             })
     private Atlas multiPolygonWithHole;
 
+    @TestAtlas(points = { @Point(id = "1", coordinates = @Loc(value = LOCATION_13)),
+            @Point(id = "2", coordinates = @Loc(value = LOCATION_14))
+
+    }, relations = {
+            @Relation(id = "3", tags = { "bugs=no" }, members = {
```
MikeGost: bugs=no :) :)

matthieun:
YAY

PR #275, opened 2018-11-17
Tippecanoe Line Delimited GeoJSON Tweaks for atlas-checks

hallahan:
### Description:

Just a few tiny tweaks that improve the public access to code from #269.

### Potential Impact:

No impact, the actual functionality is identical.

### Unit Test Approach:

Small PR here, no need to change existing tests.

### Test Results:

Still passing.

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
@@ -54,29 +54,39 @@ public final class TippecanoeCommands
         }
         // When you look up the version, tippecanoe exits with 1, so getting here is normal.
         // We want to get into this catch.
+        // https://github.com/osmlab/atlas/pull/275#discussion_r235430727
         catch (final CoreException exception)
         {
-            final Optional<String> result = MONITOR.getResult();
-            if (result.isPresent())
+            logger.warn(
+                    "Note that tippecanoe exited with 1 when checking the version; this is known and does not cause a problem for us.",
+                    exception);
+            return checkVersion();
+        }
+        // Exception or not, we can check the version and proceed either way. If tippecanoe stops
+        // exiting with 1, then we will handle the same logic outside of the exception.
+        return checkVersion();
+    }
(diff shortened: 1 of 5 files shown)


hallahan:
This PR is to make the following atlas-checks PR possible:

https://github.com/osmlab/atlas-checks/pull/101

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
```diff
@@ -0,0 +1,146 @@
...
+     * @return Whether you have a valid tippecanoe ready to execute in your environment.
+     */
+    public static boolean hasValidTippecanoe()
+    {
+        final String[] commandArray = new String[] { "tippecanoe", "--version" };
+        try
+        {
+            RunScript.run(commandArray, Collections.singletonList(MONITOR));
+        }
+        // When you look up the version, tippecanoe exits with 1, so getting here is normal.
```
lucaspcram: That is very strange behavior on `tippecanoe`'s part - so I appreciate this explanation. I initially passed over it and was confused as to why you were doing all this in the `catch` block.

It might be worth opening on issue for it on `tippecanoe`. Just a thought.

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
```diff
@@ -0,0 +1,146 @@
...
+                    }
+                    else
+                    {
+                        logger.error(
+                                "Your version of tippecanoe is too old! The minimum version is {}",
+                                TippecanoeSettings.MIN_VERSION);
+                    }
+                }
+            }
+        }
```
lucaspcram: It might be worth adding some sort of "unexpected tippecanoe return value" or "failed to parse" log messages here and inside the `catch` block at the end.

That way if `tippecanoe` is ever patched to fix the `--version` exit code, or if the format of `--version` changes you won't just get a silent failure.

hallahan:
Ready for merge / release.

PR #276, opened 2018-11-20
SubAtlas cut type functionality

MikeGost:
### Description:

- The original intent was to add more granularity to the type of cutting performed when creating a subAtlas. Specifically, the intent was to keep the existing soft-cut option and add two new "hard-cut" and "hard-cut-relations-only" options. Both of these have been added for the `subAtlas` with predicate call and their examples and logic is documented in the code. 

- Updated the Polygon sub-atlas functionality to follow the same model as the Predicate sub-atlas call.

- Refactored all sub-atlas creation out of `BareAtlas` and into `SubAtlasCreator`, for cleaner behavior separation.

- Several new Atlas APIs checking for full feature containment within a given `Polygon`.

- Minor optimizations surrounding relation handling, leveraging `relationsLowerOrderFirst` method to remove the looping in the existing subAtlas code. 

- Found an bug in the `Rectangle.forCorners` method. The method takes in a lower left and an upper right location. However, if we pass in an upper left and lower right, the `Rectangle` creation will succeed, however, successive `fullyGeometricallyEncloses` method calls will fail because they will use member variable fields for what they think is the lower left and upper right and compare the latitude and longitude of those locations to some target location, leading to an invalid result. I've added constraints to verify that the lower left and upper right are actually that.

~~**DNM**: There is more refactor possible. Possible improvements include:~~
1. ~~More documentation~~
2. ~~Apply the same logic to the `subAtlas` with `Polygon` method~~
3. ~~Refactor to move all `subAtlas` code away from `BareAtlas` - that class has grown too large~~

### Potential Impact:

None - subAtlas logic did not change. There are now more Atlas APIs available and a finer granularity for all users to specify a cut type when creating sub atlases.

### Unit Test Approach:

All existing unit tests pass. I've created new unit tests to capture all entity within behavior, new subAtlas behavior and to verify the `Rectangle` `forCorners` fix.

### Test Results:

All integration tests pass.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/Atlas.java
```diff
@@ -764,31 +825,27 @@ default AtlasSize size()
...
      *            The matcher to consider
-     * @return An optional sub-atlas. The optional will be empty in case the matcher would return an
-     *         empty atlas, which is not allowed.
+     * @param cutType
+     *            The type of cut to perform
+     * @return An optional sub-atlas. The optional will be empty in case the matcher and cut-type
+     *         return an empty atlas, which is not allowed.
      */
-    Optional<Atlas> subAtlas(Predicate<AtlasEntity> matcher);
+    Optional<Atlas> subAtlas(Predicate<AtlasEntity> matcher, AtlasCutType cutType);
```
jklamer: Please leave the backwards compatible function in the interface.

jklamer: Any algorithms operating on the Atlas interface won't be able to use the old function without casting or converting to a specific implementation.

MikeGost: I'll leave it open for discussion, could go either why on this. Updating the API as a breaking change with the new signature makes sense to me, but if this causes too much churn, we can leave it as is, just not as clean, having two subAtlas methods here, with one doing the same thing as the other.

jklamer: I believe you can make a default implementation of the original signature that uses the newer signature. Then with you put a @Depricated on the old signature, and remove at the next major version increment. That would be my preference. Gives all the users a longer time to adjust for the same result.

MikeGost: Good idea, I'll give that a shot.

src/test/java/org/openstreetmap/atlas/geography/RectangleTest.java
```diff
@@ -90,6 +113,25 @@ public void testContract()
                 collapsedVertically.contract(Distance.TEN_MILES));
     }
 
+    @Test
+    public void testCoversAndCoversPartially()
+    {
+        Assert.assertTrue("Rectangle 3 fully contains rectangle 2",
+                this.rectangle3.fullyGeometricallyEncloses(this.rectangle2));
+        Assert.assertTrue("That means, it should also cover it partially",
+                this.rectangle3.overlaps(this.rectangle4));
```
adahn6: is this supposed to be 
```JAVA 
this.rectangle3.overlaps(this.rectangle2)
```
?

src/main/java/org/openstreetmap/atlas/geography/atlas/AbstractAtlas.java
```diff
@@ -164,6 +164,17 @@ public static Atlas forOsmPbf(final Resource resource)
         return Iterables.filterTranslate(areasIntersecting(polygon), item -> item, matcher);
     }
 
+    @Override
+    public Iterable<Area> areasWithin(final Polygon polygon)
+    {
+        final Iterable<Area> areas = this.getAreaSpatialIndex().get(polygon.bounds());
+        return Iterables.filter(areas, area ->
```
matthieun: Optional: This is the old way of filtering. How about bringing new changes with `Iterables.stream(...).filter(...)`?

MikeGost: Saw this too, wanted to stay consistent to the rest of the class. Let me go ahead and re-write the rest of the function to follow the same format.

src/test/java/org/openstreetmap/atlas/geography/RectangleTest.java
MikeGost: Good catch! This code was re-ordered, hadn't noticed this. Fixed!

PR #279, opened 2018-11-27
Tokenize string in RunScript

jwpgage:
### Description:

Tokenize the command string on whitespace. Previously this would pass the full command string to a single element of an array, which would cause RunScript to look for a program named the program name plus its arguments. This will correctly split the arguments from the program name.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
@@ -22,12 +22,12 @@ public final class RunScript
 
     public static void run(final String command)
     {
-        run(new String[] { command });
+        run(command.split("\\s+"));
     }
 
     public static void run(final String command, final List<RunScriptMonitor> monitors)
     {
-        run(new String[] { command }, monitors);
+        run(command.split("\\s+"), monitors);
     }
 
     public static void run(final String[] commandArray)


hallahan:
Perhaps we should identically replicate what's being done in Java with StringTokenizer? That way we won't have other bugs pop up in the future?

```java
public StringTokenizer(String str) {
        this(str, " \t\n\r\f", false);
    }
```

```java
public Process exec(String command, String[] envp, File dir)
        throws IOException {
        if (command.length() == 0)
            throw new IllegalArgumentException("Empty command");

        StringTokenizer st = new StringTokenizer(command);
        String[] cmdarray = new String[st.countTokens()];
        for (int i = 0; st.hasMoreTokens(); i++)
            cmdarray[i] = st.nextToken();
        return exec(cmdarray, envp, dir);
    }
```

src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
```diff
@@ -22,12 +22,12 @@
 
     public static void run(final String command)
     {
-        run(new String[] { command });
+        run(command.split("\\s+"));
```
lucaspcram: Just worth noting, the regex `\s+` will effectively match `[\r\n\t\f\v ]+`. The old `run` method split against `[ \t\n\r\f]` thru Java's `StringTokenizer` class. Considering `\v` is a vertical tab character, I doubt this will cause any breakages in client code (nobody uses vertical tabs).

lucaspcram: If you wanted to perfectly match what the old code did, this line should become:
```java
run(command.split("[ \t\n\r\f]"))
```

hallahan: Great!

hallahan:
Thanks for the clarification. Though not exact replica, more concise. LGTM!

src/main/java/org/openstreetmap/atlas/utilities/runtime/RunScript.java
MikeGost: I've learned a lot about regex last few days :)

PR #280, opened 2018-11-27
[Change] Finished implementing methods for ChangeAtlas

matthieun:
### Description:

Finished implementing methods for ChangeAtlas. Also added some tests.

On the side: Added a few methods to Iterables and PolyLine, for convenience in tests.

### Potential Impact:

N/A

### Unit Test Approach:

Test various use cases for changing features in a ChangeAtlas

### Test Results:

All pass except one that is ignored, as it will succeed when relations can self remove missing members.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
@@ -72,7 +72,7 @@ public class ChangeAtlas extends AbstractAtlas // NOSONAR
     }
 
     @Override
-    public Rectangle bounds()
+    public synchronized Rectangle bounds()
     {
         if (this.bounds == null)
         {
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
@@ -11,6 +11,7 @@ import org.openstreetmap.atlas.geography.Rectangle;
 import org.openstreetmap.atlas.geography.atlas.AbstractAtlas;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.AtlasMetaData;
+import org.openstreetmap.atlas.geography.atlas.builder.AtlasSize;
 import org.openstreetmap.atlas.geography.atlas.change.rule.ChangeType;
 import org.openstreetmap.atlas.geography.atlas.change.rule.FeatureChange;
 import org.openstreetmap.atlas.geography.atlas.change.validators.ChangeAtlasValidator;
@@ -38,6 +39,9 @@ public class ChangeAtlas extends AbstractAtlas // NOSONAR
     private final Change change;
     private final Atlas source;
 
+    private transient Rectangle bounds;
+    private transient AtlasMetaData metaData;
+
     public ChangeAtlas(final Atlas source, final Change change)
     {
         if (change == null)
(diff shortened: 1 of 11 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
```diff
@@ -56,19 +60,27 @@ public ChangeAtlas(final Atlas source, final Change change)
...
 
     @Override
-    public Rectangle bounds()
+    public synchronized Rectangle bounds()
     {
-        throw new UnsupportedOperationException();
+        if (this.bounds == null)
+        {
+            // Stream it to make sure the "Iterable" signature is used here (vs. Located, which
+            // would stack overflow).
```
lucaspcram: Hah, good catch.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
```diff
@@ -88,19 +100,26 @@ public Edge edge(final long identifier)
...
 
     @Override
-    public AtlasMetaData metaData()
+    public synchronized AtlasMetaData metaData()
     {
-        throw new UnsupportedOperationException();
+        if (this.metaData == null)
+        {
+            final AtlasSize size = new AtlasSize(this);
+            this.metaData = this.source.metaData().copyWithNewSize(size).copyWithNewOriginal(false);
```
lucaspcram: What happens here if the source `metadata` is also `null`?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
```diff
@@ -120,61 +139,65 @@ public Node node(final long identifier)
     @Override
     public long numberOfAreas()
     {
-        throw new UnsupportedOperationException();
+        return Iterables.size(areas());
```
lucaspcram: It might be worth caching these values once you've computed them with `Iterables.size()`? What do you think? Maybe it's just unnecessary code bloat?

lucaspcram:
```java
private Optional<Long> numberOfAreas = Optional.empty();
.
.
.
public long numberOfAreas()
{
    if (numberOfAreas.isPresent()) {
         return numberOfAreas.get();
    } else {
        numberOfAreas = Optional.of(Iterables.size(areas()));
        return numberOfAreas.get();
    }
}
```

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
```diff
@@ -96,9 +98,16 @@ private RelationMemberList membersFor(final RelationBean bean)
         final List<RelationMember> memberList = new ArrayList<>();
         for (final RelationBeanItem item : bean)
         {
-            memberList.add(new RelationMember(item.getRole(),
-                    getChangeAtlas().entity(item.getIdentifier(), item.getType()),
-                    getIdentifier()));
+            final AtlasEntity memberChangeEntity = getChangeAtlas().entity(item.getIdentifier(),
+                    item.getType());
+            if (memberChangeEntity == null)
```
lucaspcram: Once we update the member fetching for `ChangeRelation`, seeing this exception will mean something went seriously wrong - ie. it's a good sanity check that our code is working correctly.

src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlasTest.java
```diff
@@ -0,0 +1,243 @@
...
+import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.geography.atlas.items.Relation;
+import org.openstreetmap.atlas.geography.atlas.items.RelationMemberList;
+import org.openstreetmap.atlas.utilities.collections.Iterables;
+import org.openstreetmap.atlas.utilities.scalars.Distance;
+
+/**
+ * @author matthieun
+ */
+public class ChangeAtlasTest
```
lucaspcram: Good stuff.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
matthieun: NPE!!!

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
matthieun: Good idea.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
lucaspcram: Thinking about this some more, is it ever possible for `metaData` to actually be `null` in a valid `Atlas`? I don't think so. It looks like `PackedAtlasSerializer` will throw an exception if it cannot deserialize a field, so you'll never have a `null` metadata from a `PackedAtlas`, at least.

matthieun: I added a check anyway.

PR #282, opened 2018-11-27
Some new ChangeAtlas updates

lucaspcram:
### Description:

Nodes now filter out in/out edges that do not exist.

`forEntities` in `ChangeAtlas` now filters out entities that were removed.

`Bloated` features now properly calculate their bounds if their geometry is updated. The bounding box contains both the old geometry and the new.

Still need TODO: Edges need to return null if a connected node is missing.
Still need TODO: Relations needs to filter out non-existent members, and handle case where they become empty or shallow.


lucaspcram:
Shoot I refreshed and it messed up my target branch. Didn't even notice. Changing this now.

PR #284, opened 2018-11-27
[Change] More ChangeAtlas updates

lucaspcram:
### Description:

Nodes now filter out in/out edges that do not exist.

forEntities in ChangeAtlas now filters out entities that were removed.

Bloated features now properly calculate their bounds if their geometry is updated. The bounding box contains both the old geometry and the new.

Still need TODO: Edges need to return null if a connected node is missing.
Still need TODO: Relations needs to filter out non-existent members, and handle case where they become empty or shallow.

------


src/test/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeAtlasEdgeValidatorTest.java
```diff
@@ -45,12 +45,13 @@ public void testModifyEdgeAndNode()
...
 
         final Change change = changeBuilder.get();
-        Assert.assertEquals("[Edge: id=39001000001, startNode=38999000000, endNode=39002000000, "
-                + "polyLine=LINESTRING (-122.2457961 37.592796, -122.2450237 37.5926929, "
-                + "-122.2441049 37.5930666, -122.2429584 37.5926993), "
-                + "[Tags: [last_edit_user_name => myself], [last_edit_changeset => 1], "
-                + "[last_edit_time => 1513719782000], [last_edit_user_id => 1], [name => primary], "
-                + "[highway => primary], [last_edit_version => 1]]]",
+        Assert.assertEquals(
+                "[Edge: id=39001000001, startNode=38999000000, endNode=39002000000, "
```
matthieun: Does that pass spotless?

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
```diff
@@ -36,25 +48,31 @@ public static BloatedArea fromArea(final Area area)
...
-        return new BloatedArea(area.getIdentifier()).withBounds(area.asPolygon().bounds());
+        return new BloatedArea(area.getIdentifier(), area.asPolygon());
     }
 
     BloatedArea(final long identifier)
     {
         this(identifier, null, null, null);
     }
 
+    BloatedArea(final long identifier, final Polygon polygon)
```
matthieun: Shouldn't that take just the bounds and not a full polygon?

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
```diff
@@ -101,9 +119,8 @@ public int hashCode()
     @Override
     public Set<Relation> relations()
     {
-        return this.relationIdentifiers == null ? null
-                : this.relationIdentifiers.stream().map(BloatedRelation::new)
-                        .collect(Collectors.toSet());
+        return this.relationIdentifiers == null ? null : this.relationIdentifiers.stream()
```
matthieun: :( looks like we have a save action mismatch.

lucaspcram: I believe it's one of those cases where my save action does not agree with spotless.

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
lucaspcram: We could do that instead, and go back to your private `withOriginalBounds` method idea.

src/test/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeAtlasEdgeValidatorTest.java
lucaspcram: I think it does. I ran `spotlessApply` before committing.

PR #292, opened 2018-12-01
[Change] Fixed bug with missing parent relations

matthieun:
### Description:

- Validation checks that "claimed" parent relations, when they exist, also claim the feature as a member. (Reciprocity)
- Relation members are now filtered in case one of the declared members does not exist
- Relation bean merging is generalized, and is used:
  - When merging two `FeatureChange` items of the same relation
  - When `ChangeAtlas` has a source and an override relation member list to merge
- BloatedEntities now all declare their specific equals method, that checks all the members, null or not.

### Potential Impact:

N/A

### Unit Test Approach:

Unit tests have been amended

### Test Results:

All unit tests pass

------


matthieun:
Found one more bug. Reopening when fixed.

PR #293, opened 2018-12-03
[Change] Added an AtlasDiff class to produce Changes from a before and after Atlas

lucaspcram:
EDIT: this has been re-opened and is now ready for merge into `change`, barring any further review concerns or requests. Since closing, I have also refactored a couple things, see the `Refactored` section below.

### Description:

Added a new class `AtlasDiff`, which can take a before and after `atlas`, and produce a `Change` object that would transform `before` into `after`.  See the javadoc in `AtlasDiff` for more details. This class is intended to replace the `AtlasDelta` class.

~`AtlasDiff` configuration is flexible. It allows for quick diff computation using the entities builtin to the after `atlas`. But it also allows the user to use new `BloatedEntities` for the underlying feature changes, which makes the resulting `Change` fully serializable.~ *Use of `BloatedEntites` is now required.* One can optionally specify to `saveAllGeometry`, which allows for the changes to be easily visualized (e.g. a Node that had its tags changed will also save its `Location`, so that a GeoJSON viewer can highlight the changed Node). Currently, the `saveAllGeometry` is default enabled.

This class will become more valuable once the `Change` object can serialize itself to GeoJSON. This will allow for easy viewing of `atlas` differences using your favorite GeoJSON visualizer.

### Refactored

1) Refactored the `BloatedX` tag methods to be `withAddedTag`/`withRemovedTag` instead of `withTagExtra`/`withTagLess` - this was just a minor change for clarity when quickly reading code.

2) Refactored `BloatedRelation::withMembers` - there are now two interfaces: `BloatedRelation::withMembers` and `BloatedRelation::withMembersAndSource`. The reason for this: having two interfaces allows us to explicitly check for a `null` source. Before, passing a `null` source was indistinguishable from when the user called one of the overloaded, non-sourced versions of `withMembers` (since these versions just called the sourced version with a `null` source). This meant that explicitly passing a `null` source (possibly by mistake) to `withMembers` would silently cause unintended behaviour in downstream code. I have opted to remove this potential foot-gun now, at the cost of a slightly-more-verbose API.

3) Removed ability for `Change` to be empty, and for `FeatureChange` to have a non-bloated entity reference. This greatly simplified the diff code.

### Potential Impact:

N/A

### Unit Test Approach:

Unit tests contained in `AtlasDiffTest` and `AtlasDiffTestRule`.

### Test Results:

All unit tests pass, finally!

------


src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -235,6 +238,7 @@ else if (thisEntity instanceof Relation)
                     final RelationMemberList thatMembers = ((Relation) thatEntity).members();
                     if (!thisMembers.equals(thatMembers))
                     {
+                        logger.error("ERRORRRRR: {} vs {}", thisMembers, thatMembers);
```
adahn6: this is the only acceptable error format

lucaspcram: Heh. There's a lot of junk logging in here due to a confusing bug. That will not be in the final PR :)

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -75,6 +76,12 @@ public static FeatureChange remove(final AtlasEntity reference)
     }
 
     public FeatureChange(final ChangeType changeType, final AtlasEntity reference)
+    {
+        this(changeType, reference, null);
```
adahn6: IMO null should be avoided in favor of an empty String, but this is probably a nit. thoughts?

lucaspcram: You're probably right. But this is also going to be removed once the bug is solved. Just temporarily added this for logging purposes. Although I suppose we can make a case to @matthieun to add this functionality permanently. However, it would need to respect the merge semantics in a sensible way.

adahn6: fair enough!

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
```diff
@@ -0,0 +1,467 @@
...
+    }
+
+    /**
+     * Create a set of modify {@link FeatureChange}s ie. ADDs that change an existing feature. The
+     * set will contain an individual {@link FeatureChange} for certain types of differences. Eg. if
+     * a Point's location AND tags changed, then the set will contain two feature changes, one for
+     * the location and one for the tags. However, if a Node's location and connected edges change,
+     * this will come in as one feature change.<br>
+     *
+     * @param entity
```
matthieun: We usually expect comments on each variable here if the javadoc comment is present.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
```diff
@@ -0,0 +1,467 @@
...
+
+        /*
+         * Determine if the entities are actually different in any relevant way. If so, then we can
+         * decide how to create the feature change.
+         */
+
+        /*
+         * Detect changed tags.
+         */
+        AtlasDiffHelper.getTagChangeIfNecessary(beforeEntity, afterEntity, useBloatedEntities,
```
matthieun: Concise and readable!

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
```diff
@@ -0,0 +1,467 @@
...
+                    .getNodeChangeIfNecessary((Node) beforeEntity, (Node) afterEntity,
+                            useGeometryMatching, useBloatedEntities, saveAllGeometries)
+                    .ifPresent(featureChanges::add);
+        }
+
+        /*
+         * Detect if the entities were Edges and some Edge properties changed. We check for Edge
+         * polylines, as well as the start and end Node.
+         */
+        if (entity instanceof Edge)
```
matthieun: else if? So that it does not re-evaluate every time?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiffHelper.java
```diff
@@ -0,0 +1,771 @@
...
+            final List<PolyLine> candidates = Iterables.stream(otherEdges).map(Edge::asPolyLine)
+                    .collectToList();
+            final Optional<PolyLineRoute> match = source.costDistanceToOneWay(candidates)
+                    .match(Distance.ZERO);
+            if (match.isPresent() && match.get().getCost().isLessThanOrEqualTo(Distance.ZERO))
+            {
+                // The edge was probably split by way sectioning without changing itself.
+                logger.trace("Edge {} from {} has no equal member but found a match with no cost.",
+                        edge, edge.getAtlas().getName());
+                return true;
```
matthieun: If I remember correctly, this one was to ignore the ways that are way-sectioned differently but still have the same geometry overall. In the Change case, those would still need to have FeatureChange objects after all. Is that something that is covered after?

src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlasTest.java
```diff
@@ -50,8 +50,9 @@
     public void testBounds()
     {
         final Atlas atlas = this.rule.getAtlas();
-        Assert.assertEquals("POLYGON ((-122.2450237 37.5920679, -122.2450237 37.5938783, "
-                + "-122.2412753 37.5938783, -122.2412753 37.5920679, -122.2450237 37.5920679))",
+        Assert.assertEquals(
```
matthieun: Spotless?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
```diff
@@ -0,0 +1,467 @@
...
+                .forEach(this.addedEntities::add);
+
+        /*
+         * Check for entities that were potentially modified in the after atlas (here, a potentially
+         * modified entity is any entity what was present in both the before and after atlas). If we
+         * find any, add them to a set for later processing. We will use this set to create
+         * FeatureChanges.
+         */
+        Iterables.stream(this.before)
+                .filter(beforeEntity -> !isEntityMissingFromGivenAtlas(beforeEntity, this.after,
```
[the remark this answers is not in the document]
lucaspcram: Are you thinking some kind of helper predicate supplier? Like:
```java
private Predicate<AtlasEntity> getFilter(Atlas atlas, boolean geoMatching) {
    return entity -> isEntityMissingFromAtlas(entity, atlas, geoMatching)
}
// then
Iterables.stream(this.after).filter(getFilter(this.before, geoMatching)).forEach(...);

```

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
lucaspcram: Yeah good catch. I was copy-pasting from the catch-all `getTagChanges`/`getParentRelationChanges` part and forgot to add the `else` statements here.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiffHelper.java
lucaspcram: Let's definitely discuss this further, I am not sure I am 100 percent clear.

src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlasTest.java
lucaspcram: I ran `./gradlew spotlessApply` and this is the formatting it does. Maybe your `spotlessApply` is different from mine?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/Change.java
```diff
@@ -94,6 +99,11 @@ public String getName()
         }
     }
 
+    public boolean hasChanges()
+    {
+        return !this.featureChanges.isEmpty();
```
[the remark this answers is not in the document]
lucaspcram: You are right. Since `Change` is located this should be disallowed. I will update.

lucaspcram:
Closing until ready to stop annoying PR emails. Will re-open when ready.

lucaspcram:
@matthieun I removed all the `Sonar` warnings, yes.

PR #295, opened 2018-12-04
[Change] Keep track of explicitly removed relation members

matthieun:
### Description:

This fixes a bug where a relation member that was intentionally removed would be added back when merging relation beans.

### Potential Impact:

N/A

### Unit Test Approach:

Altered unit tests.

### Test Results:

Tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
@@ -293,6 +293,40 @@ public class BloatedRelation extends Relation implements BloatedEntity
         return withMembers(source, members.asBean(), members.bounds());
     }
 
+    /**
+     * Assign this {@link BloatedRelation} with members.
+     * <p>
+     * In case this {@link BloatedRelation} is created from an existing relation, and the new member
+     * list has had some existing members removed, use
+     * {@link #withMembers(Optional, RelationBean, Rectangle)}
+     *
+     * @param members
+     *            The members of the relation
+     * @param bounds
+     *            The bounds of all the members of the relation.
+     * @return This
+     */
+    public BloatedRelation withMembers(final RelationBean members, final Rectangle bounds)
+    {
+        return withMembers(Optional.empty(), members, bounds);
+    }
+
+    /**
+     * Assign this {@link BloatedRelation} with members.
+     * <p>
+     * In case this {@link BloatedRelation} is created from an existing relation, and the new member
+     * list has had some existing members removed, use
+     * {@link #withMembers(Optional, RelationMemberList)}
+     * 
+     * @param members
+     *            The full members of the Relation
+     * @return This
+     */
+    public BloatedRelation withMembers(final RelationMemberList members)
+    {
+        return withMembers(Optional.empty(), members);
+    }
+
     public BloatedRelation withOsmRelationIdentifier(final Long osmRelationIdentifier)
     {
         this.osmRelationIdentifier = osmRelationIdentifier;
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
@@ -4,6 +4,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
```diff
@@ -241,25 +242,55 @@ public BloatedRelation withIdentifier(final long identifier)
...
+     *            The relation that was used as a base to create that BloatedRelation, if any. Due
+     *            to the weak nature of relation membership across Atlas(es), this helps decide what
+     *            relation members are forcibly removed if any.
+     * @param members
+     *            The members of the relation
+     * @param bounds
+     *            The bounds of all the members of the relation.
+     * @return This.
+     */
+    public BloatedRelation withMembers(final Optional<Relation> source, final RelationBean members,
```
lucaspcram: Any particular reason we have `source` as an `Optional`? It's fine, I just don't see `Optionals` as method parameters very often.

Also, is there any advantage to having `source` be a parameter? Can we not just store `source` as a transient member at `shallowFrom()` call-time? That way, we avoid the headache of something like:
```java
BloatedRelation bloatedRelation = BloatedRelation.shallowFrom(someSourceRelation);
.
. // some code here
.
bloatedRelation.withMembers(Optional.of(accidentallyUsedADifferentSourceRelation), newMembers)
```

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
```diff
@@ -241,25 +242,55 @@ public BloatedRelation withIdentifier(final long identifier)
...
+            final Rectangle bounds)
     {
         this.members = members;
+        if (source.isPresent())
+        {
+            // This has been created from an existing relation, make sure to record the members that
+            // have been intentionally omitted, so as not to add them back in the future when either
+            // merging FeatureChanges or stitching MultiAtlases.
+            for (final RelationMember member : source.get().members())
```
lucaspcram: We probably want to do a `this.members.clear()` right above this `for` loop. That way, if someone does:
```java
bloatedRelation.withMembers(newMembers)
.
. // more code here
.
bloatedRelation.withMembers(someConditionalDictatesToUseTheseInstead)
```
you don't end up with a corrupted `members` set.

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
matthieun: So the point of having a shallow object it to not store the source relation if it does not have any need for it. At that point, we do not know if we will have a need for it, as removing a relation member is a sub-set of the potential use case... Also BloatedRelation is not in the change package as it was intended to be more independent, which is why I made the source optional. In case of creating a BloatedRelation for something other than a change, you do not have a source relation. Let me know if that makes sense, I am open to changing it if needed :)

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
lucaspcram: Ah you're right. I missed that line :)

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedRelation.java
lucaspcram: Ok I see. To be fair, you're only storing a relation reference. Now, it would prevent that relation from being garbage collected. But is there ever a case where you would create a `BloatedRelation` from some relation `R`, and then have `R` go out of scope (and get GC'd) while the `BloatedRelation` lives on? Otherwise, storing an extra reference to `R` in the `BloatedRelation` won't have much of an effect (it's just an 8 byte object reference).

But I do see what you're saying as far as keeping it general, and not too tied to our "change" use-case. That being said, what about overloading the API?


```java
// General version for other use-cases, no need to worry about passing a source
public BloatedRelation withMembers(final RelationBean members,
            final Rectangle bounds) {
    this.withMembers(null, members, bounds);
}

// New version for change use-cases
public BloatedRelation withMembers(Relation source, final RelationBean members,
            final Rectangle bounds)
    {
        this.members = members;
        if (source != null)
        {
            // This has been created from an existing relation, make sure to record the members that
            // have been intentionally omitted, so as not to add them back in the future when either
            // merging FeatureChanges or stitching MultiAtlases.
            for (final RelationMember member : source.members())
            {
                if (!members.getItemFor(member.getEntity().getIdentifier(),
                        member.getEntity().getType()).isPresent())
                {
                    this.members.addItemExplicitlyExcluded(member.getEntity().getIdentifier(),
                            member.getRole(), member.getEntity().getType());
                }
            }
        }
        updateBounds(bounds);
        return this;
    }
```

matthieun: That sounds good. I will make it clear in the comments that in the case of removing members, the other one needs to be used.

lucaspcram: Thanks!

matthieun: Argh I kept the optional... one moment.

matthieun: Argh it was merged!

PR #296, opened 2018-12-04
Tippecanoe Improvements for atlas-checks

hallahan:
### Description:

There are a few small refactorings that make the line-delimited GeoJSON & tippecanoe code work better with atlas-checks.

### Potential Impact:

This PR makes the following PR possible:

https://github.com/osmlab/atlas-checks/pull/101

### Unit Test Approach:

Just refactoring and making a few things public.

### Test Results:

N/A

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
@@ -10,6 +10,7 @@ import java.util.Optional;
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.geography.atlas.geojson.LineDelimitedGeoJsonConverter;
 import org.openstreetmap.atlas.utilities.runtime.RunScript;
 import org.openstreetmap.atlas.utilities.runtime.SingleLineMonitor;
 import org.openstreetmap.atlas.utilities.time.Time;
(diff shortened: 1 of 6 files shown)


hallahan:
cc/ @lucaspcram

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
```diff
@@ -39,6 +40,55 @@ private TippecanoeCommands()
...
+     *
+     * @param geojsonDirectory
+     *            The directory to decompress.
+     */
+    public static void decompress(final Path geojsonDirectory)
+    {
+        final Time time = Time.now();
+        final String directory = geojsonDirectory.toString();
+        final String cat = String.format(
+                "find '%s' -type f -name '*.geojson.gz' -exec gzip -dr \\{\\} +", directory);
```
lucaspcram: Woohoo!

src/main/java/org/openstreetmap/atlas/utilities/vectortiles/TippecanoeCommands.java
```diff
@@ -39,6 +40,55 @@ private TippecanoeCommands()
...
+     * @param geojsonDirectory
+     *            The directory of GeoJSON files to concatenate.
+     */
+    public static void concatenate(final Path geojsonDirectory)
+    {
+        final Time time = Time.now();
+        logger.info("Concatenating GeoJSON...");
+        final String directory = geojsonDirectory.toString();
+        final String cat = String.format(
+                "find '%s' -type f -name '*.geojson' \\! -path '%s'/%s -exec cat \\{\\} + > '%s/'%s",
```
lucaspcram: As long as the commands calling this function are never run with super-user privileges (or even as a different user than the one invoking it), this is safe.

I am curious what you have in mind in regards to how the `TippecanoeExporter` and the `LineDelimitedGeoJSONConverter` will generally be invoked by end users? I would recommend for whatever program invokes these two commands  to at the very least sanitize its input for characters like `;`.

hallahan: Note that the method is `public static void concatenate(final Path geojsonDirectory)`. At least we know it's a `Path` getting fed in there.

lucaspcram: True. But someone could still do something like `-geojsonPath="myPath; echo vulnerable"`. However - I took a closer look and as far as I can tell, every user-provided string has been single-quoted, so I think it's actually all good!

I'm just always extra cautious about these things.

PR #302, opened 2018-12-07
Add Level Tag

sayas01:
### Description:

This PR adds LevelTag to Atlas. LevelTag defines the floor number of an OSM feature. The tag information is useful when dealing with ways that are tied to a floor level of a multilevel building such as a multilevel parking lot.
For further details, see:
1) TagInfo - https://taginfo.openstreetmap.org/keys/level#overview
2) OSM Wiki page - https://wiki.openstreetmap.org/wiki/Key:level

### Potential Impact:

No potential impact except that the new tag can be used.

### Unit Test Approach:

NA

### Test Results:

NA

------


src/main/java/org/openstreetmap/atlas/tags/LevelTag.java
```diff
@@ -0,0 +1,48 @@
...
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+import org.openstreetmap.atlas.tags.annotations.extraction.LongExtractor;
+
+/**
+ * OSM level tag
+ *
+ * @author sayas01
+ */
+@Tag(value = Tag.Validation.LONG, taginfo = "http://taginfo.openstreetmap.org/keys/level#values", osm = "http://wiki.openstreetmap.org/wiki/Level")
+
```
MikeGost: Nit: extra space

src/main/java/org/openstreetmap/atlas/tags/LevelTag.java
```diff
@@ -0,0 +1,48 @@
...
+public interface LevelTag
```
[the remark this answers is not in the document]
MikeGost: Might be worth a Sonar ignore for all our tag classes, otherwise this will happen all over the place.

src/main/java/org/openstreetmap/atlas/tags/LevelTag.java
```diff
@@ -0,0 +1,48 @@
...
+import org.openstreetmap.atlas.tags.annotations.Tag;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+import org.openstreetmap.atlas.tags.annotations.extraction.LongExtractor;
+
+/**
+ * OSM level tag
+ *
+ * @author sayas01
+ */
+@Tag(value = Tag.Validation.LONG, taginfo = "http://taginfo.openstreetmap.org/keys/level#values", osm = "http://wiki.openstreetmap.org/wiki/Level")
```
MikeGost: I'm not sure that Long is the correct validator? From a brief look, around 5-6% of valid values are represented as strings. Thoughts on changing this to String? We could make it even more robust by accounting for split values and trying to convert to a Long right away, otherwise falling back to split on `;` for the correct value.

sayas01: The values for this tag are sometimes a combination of character and numeric, split by `,`,`;` or `:`.  Also there are values like `1-19` representing a range. There would be a couple of cases we will need to handle to get the correct value. For now, I have updated the type to String. Let me know your thoughts on this.

sayas01:
@MikeGost I have suppressed the rule for this class. We may need to do the same for other tag classes. I do not think I have the access to deactivate the rule in quality profile.

PR #306, opened 2018-12-13
Add validation for edge-node reciprocity

matthieun:
### Description:

- Fixed an issue with `BloatedRelation.withExtraMember` which was looking at the wrong member set for a tag source
- Added a bunch of convenience methods in Bloated entities
- Updated logging levels for validation
- Added Atlas validation for edge-node connectivity: all edges that reference a node must be present in the node's corresponding connected edge set.
- Fixed discrepancies in the `AtlasChangeGeneratorSplitRoundabout` test, surfaced by the new validation.

### Potential Impact:

Less ways to mis-construct a Change object.

### Unit Test Approach:

Updated existing unit test.

### Test Results:

Tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
@@ -137,6 +137,11 @@ public class BloatedArea extends Area implements BloatedEntity
                 + ", tags=" + this.tags + ", relationIdentifiers=" + this.relationIdentifiers + "]";
     }
 
+    public BloatedArea withAddedTag(final String key, final String value)
+    {
+        return withTags(BloatedEntity.addNewTag(getTags(), key, value));
+    }
+
     public BloatedArea withAggregateBoundsExtendedUsing(final Rectangle bounds)
     {
         if (this.aggregateBounds == null)
@@ -170,9 +175,11 @@ public class BloatedArea extends Area implements BloatedEntity
         return this;
     }
 
-    public BloatedArea withAddedTag(final String key, final String value)
+    public BloatedArea withRelations(final Set<Relation> relations)
     {
-        return withTags(BloatedEntity.addNewTag(getTags(), key, value));
+        this.relationIdentifiers = relations.stream().map(Relation::getIdentifier)
+                .collect(Collectors.toSet());
+        return this;
     }
 
     public BloatedArea withRemovedTag(final String key)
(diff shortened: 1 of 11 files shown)


matthieun:
If it does, it is not captured by any unit test as they all pass.

lucaspcram:
Just looked more carefully and I think it's OK. I got a little thrown by the diff, because it looked like you added some new methods when in reality they just got re-ordered.

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
```diff
@@ -137,6 +137,11 @@ public String toString()
                 + ", tags=" + this.tags + ", relationIdentifiers=" + this.relationIdentifiers + "]";
     }
 
+    public BloatedArea withAddedTag(final String key, final String value)
+    {
+        return withTags(BloatedEntity.addNewTag(getTags(), key, value));
```
MikeGost: In `BloatedEntity.addNewTag(..)` - do we care about overriding an existing tag? The naming implies adding a new one, but we don't differentiate between an addition or replacement. Should we log something if we override?

src/test/java/org/openstreetmap/atlas/geography/atlas/change/AtlasChangeGeneratorSplitRoundabout.java
```diff
@@ -70,8 +74,13 @@
                         Sets.treeSet(newEdgeIdentifier2), Sets.hashSet())));
 
                 // End node has a replaced start edge identifier
-                result.add(FeatureChange.add(BloatedNode.from(edge.end())
+                result.add(FeatureChange.add(BloatedNode.shallowFrom(edge.end())
+                        .withInEdges(edge.end().inEdges())
                         .withInEdgeIdentifierReplaced(oldEdgeIdentifier, newEdgeIdentifier2)));
+                // Start node has a replaced end edge identifier
+                result.add(FeatureChange.add(BloatedNode.shallowFrom(edge.start())
```
MikeGost: This makes more sense now :)

src/main/java/org/openstreetmap/atlas/geography/atlas/bloated/BloatedArea.java
lucaspcram: This is a good point. We also have a `withReplacedTag` method, so I can see how this might be confusing. I think we could either:

1) Remove `withReplacedTag` all together, and just use `withAddedTag` in all cases (having it log a warning if it replaces a tag).
2) Change `withAddedTag` to do nothing if the tag already exists (and log an error/warn)

Curious what you think of `1` or `2`.

MikeGost: I'm leaning towards 1. - using `withAddedTag` and `removeTag`, with each method logging a warning if there is a replacement or if there tag being removed doesn't exist. This is also consistent with the `ChangeType` model of representing modify operations as adds.

lucaspcram: Sounds good to me!

matthieun: Ok, will do that in a future PR.

PR #309, opened 2018-12-17
Length and Altitude Extraction

Bentleysb:
### Description:

This pr adds an improved system for parsing distances from tags. Previously the only handling for this was in the `HeightConverter` class. This improves on that class in several ways. It accounts for all distance units defined by OSM (https://wiki.openstreetmap.org/wiki/Map_Features/Units). Altitude is a separate extractor, so negative values can be extracted. `get` methods have been added to all the appropriate tag classes.
With this updated system, `HeightConverter` is marked as deprecated. 

### Potential Impact:

If the `HeightConverter` class is used in downstream libraries, those references should be switched to the new length or altitude extractor. 

### Unit Test Approach:

Unit test were added for `LengthValidator`, and similar tests have been used for the new extractor classes. Existing unit test for `BuildingHeightTag` and `ComplexBuilding` also confirm this is working as expected. 

### Test Results:

Passing unit tests.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/HeightConverter.java
```diff
@@ -8,8 +8,12 @@
 /**
  * Example use: Building Height
  *
+ * @deprecated use {@link org.openstreetmap.atlas.tags.HeightTag} {@code get()} or
+ *             {@link org.openstreetmap.atlas.tags.annotations.extraction.AltitudeExtractor}
+ *             instead.
  * @author matthieun
  */
+@Deprecated
```
matthieun: What would be the cost of just removing it in this PR?

src/test/java/org/openstreetmap/atlas/tags/annotations/extraction/AltitudeExtractorTest.java
```diff
@@ -0,0 +1,43 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.openstreetmap.atlas.geography.Altitude;
+
+/**
+ * Unit test for {@link AltitudeExtractor}.
+ *
+ * @author bbreithaupt
+ */
+public class AltitudeExtractorTest
```
matthieun: Nice and simple! Could you extend those to all the unit use cases? KM, MI and NMI? And also some upper case, lower case and mixed case?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/HeightConverter.java
Bentleysb: The only cost would be if there is something downstream that is using it. I am not aware of anything that is, but wanted to play it safe. If you think it is safe enough, I am happy to remove it.

src/test/java/org/openstreetmap/atlas/tags/annotations/extraction/AltitudeExtractorTest.java
Bentleysb: I have not added other units to the AltitudeExtractor tests, because the Altitude class only takes meters currently.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/buildings/HeightConverter.java
matthieun: Ok, let's keep it then. We can then do a sweep later for all the deprecated annotations and remove the ones unused.

src/test/java/org/openstreetmap/atlas/tags/annotations/extraction/AltitudeExtractorTest.java
matthieun: Oh yes. Thanks!

PR #318, opened 2019-01-08
Removing deprecated code and updating FastClasspathScanner to ClassGraph

MikeGost:
### Description:

There are three changes here:
1. Addressing issue #204 - porting over to the latest ClassGraph code. This involves multiple changes in the way we are doing the scanning. I followed this [wiki](https://github.com/classgraph/classgraph/wiki/Porting-FastClasspathScanner-code-to-ClassGraph) for the migration. There are also helpful examples for grabbing implementations of an interface or annotation based classes [here](https://github.com/classgraph/classgraph/wiki/Code-examples). Lastly, I chose the most frequently used version to depend on - 4.4.12. The most recent version is 4.6.12. I chose 4.4.12 because it has the most [dependencies](https://mvnrepository.com/artifact/io.github.classgraph/classgraph) out of all released versions (banking on stability) and none of the patch notes since it look too relevant. Happy to update to the most recent one if anyone has an issue with this. See all available releases and their notes [here](https://github.com/classgraph/classgraph/releases).
2. Removing deprecated `subAtlas` calls - this is a breaking change, so any downstream users will need to update their APIs when these changes are released.
3. Fixing various Sonar issues found along the way

### Potential Impact:

All downstream users relying on the `Atlas` `subAtlas` API call will be broken.

### Unit Test Approach:

All existing unit tests pass, no new ones are needed.

### Test Results:

All integration tests pass.

------

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -53,7 +53,7 @@ dependencies
     compile packages.commons.cli
     compile packages.commons.csv
     compile packages.commons.lang
-    compile packages.fastclasspath
+    compile packages.classgraph
     compile packages.guava
     compile packages.jsonassert
     compile packages.jackson.core
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -15,7 +15,7 @@ project.ext.versions = [
     cli: '1.3.1',
     csv: '1.2',
     lang: '3.4',
-    fastclasspath: '2.9.4',
+    classgraph: '4.4.12',
     guava: '18.0',
     jsonassert:'1.3.0',
     jackson_core:'2.8.8',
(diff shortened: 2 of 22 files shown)


lukehutch:
@MikeGost I do recommend upgrading to 4.6.x -- there was a big speed enhancement to handling jarfiles (especially nested jars) in 4.6.0, and some of the bugfixes between what you're depending upon and the most recent version may be important. Total number of dependent projects is not a good measure of stability (at all), it is more a measure of how long a dependency went without changes between releases. Most projects are very slow to upgrade their dependencies.

I am about to release 4.6.13 with a bugfix for locating resources in the classpath without duplication. I recommend upgrading to that version. It should be perfectly stable.

MikeGost:
@lukehutch - thanks for weighing in, I'll update to the latest and give it a shot!

MikeGost:
Updated to the latest released here: https://github.com/osmlab/atlas/pull/320

Issue #163, opened 2018-07-10
Complex Entity Support for Areas of Interest


MikeGost:
There's many Areas of Interest (AOIs) that are modeled as relations (parks, pools, etc.). It would be useful to have a `ComplexEntity` type and builder to build and return what type of area it is.

sayas01:
This feature has been implemented and merged. Issue can be closed.

MikeGost:
Closing as implemented.

PR #319, opened 2019-01-10
Complex Entity for AOI

sayas01:
### Description:

This PR adds new feature to Atlas to include Complex Entity support for Areas Of Interest (AOI). This PR addresses the OSM issue [#163](https://github.com/osmlab/atlas/issues/163). The primary source of AOI tags is the [Map Features](https://wiki.openstreetmap.org/wiki/Map_Features) page, excluding tags or values that are not in Atlas. Two new classes added are:

1. ComplexAOI : ComplexEntity implementation of Multipolygon AOI relations and areas.
2. ComplexAOIFinder : Finds all relations and areas that are candidates for ComplexAOI and converts them into ComplexAOI. OSM tags of relations and areas in the provided Atlas are checked against a default list of TaggableFilter that is provided as a json resource ['aoi-tag-filter.json'](https://github.com/sayas01/atlas/blob/828c0bc3effbd67b704c52741345470dc89448b2/src/main/resources/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/aoi-tag-filter.json). Overloaded method `find` also allows to pass a TaggableFilter in which case, the AOI tags will be checked against both the given TaggableFilter and the default filter.

### Potential Impact:

The new feature could be made use of in getting geometries of AOI.

### Unit Test Approach:

Added test cases for the ComplexAOIFinder class.

### Test Results:

All tests passed.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOIFinder.java
```diff
@@ -0,0 +1,64 @@
...
+     *            The {@link Atlas} to browse.
+     * @return {@link Iterables} of {@link ComplexAOI}.
+     */
+    @Override
+    public Iterable<ComplexAOI> find(final Atlas atlas)
+    {
+        final Iterable<ComplexAOI> iterableOfComplexAOIRelations = Iterables
+                .asIterable(StreamSupport.stream(atlas.relations().spliterator(), false)
+                        .map(ComplexAOI::getComplexAOI).filter(Optional::isPresent)
+                        .map(Optional::get).collect(Collectors.toList()));
```
matthieun: If you collect to List here, shouldn't it already be iterable? My guess is that it does not need the Iterables.asIterable wrapper?

src/main/resources/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/aoi-tag-filter.json
```diff
@@ -0,0 +1,20 @@
+{
+  "filters": [
+    "aerialway->STATION",
```
matthieun: Here each line usually represents an "AND" if you use `ConfiguredTaggableFilter`. I see that you chose to not use that class and make every line an "OR" (using stream.anyMatch). It would be nicer to use ConfiguredTaggableFilter here, but it would mean adding all those filters on one line, with OR statements (`|`), and the whole thing would be very unreadable. So I guess it makes sense as is :)

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java
```diff
@@ -0,0 +1,185 @@
...
+
+/**
+ * Complex Entity for AOI Relations and AOI Areas. AOI relations are those that are
+ * {@link MultiPolygon} and have AOI tags in it and AOI Areas are those {@link Area}s that have AOI
+ * tags in it. AOI tags are checked against the {@link TaggableFilter} passed as an argument or to
+ * the default {@link TaggableFilter} of AOI tags if the tags are not explicitly specified.
+ *
+ * @author sayas01
+ */
+public final class ComplexAOI extends ComplexEntity
```
MikeGost: Nit: For those not familiar with the abbreviation, is it better to spell this out at `ComplexAreaOfInterest`?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOIFinder.java
sayas01: You are right! I have removed the wrapper.

src/main/resources/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/aoi-tag-filter.json
sayas01: It is the same reasons why I did not use the ConfiguredTaggableFilter. As there is a good chance we may want to update the filters in the future, maintaining the filters in new line would make it readable and easy to maintain as opposed to adding all the filters in a single line with OR statements.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java
```diff
@@ -0,0 +1,185 @@
...
+
+    /**
+     * This method creates a {@link ComplexAOI} for the specified {@link AtlasEntity} if it meets
+     * the requirements for Complex AOI relation. The AOI tags are checked against the default tags.
+     *
+     * @param source
+     *            The {@link AtlasEntity} for which the ComplexEntity is created
+     * @return {@link ComplexAOI} if created, else return empty.
+     */
+    public static Optional<ComplexAOI> getComplexAOI(final AtlasEntity source)
```
MikeGost: Can this method call `getComplexAOI(source, allTrueFilter)` with a filter that returns true? That might remove some code duplication.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java
```diff
@@ -0,0 +1,185 @@
...
+    {
+        super(source);
+        try
+        {
+            this.multiPolygon = RELATION_OR_AREA_TO_MULTI_POLYGON_CONVERTER.convert(source);
+        }
+        catch (final Exception exception)
+        {
+            setInvalidReason("Unable to convert Relation to MultiPolygon", exception);
+            throw new CoreException("Unable to convert Relation to MultiPolygon", exception);
```
MikeGost: Would it help having the relation identifier here for debugging?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOIFinder.java
```diff
@@ -0,0 +1,62 @@
...
+     *
+     * @param atlas
+     *            The {@link Atlas} to browse.
+     * @return {@link Iterables} of {@link ComplexAOI}.
+     */
+    @Override
+    public Iterable<ComplexAOI> find(final Atlas atlas)
+    {
+        final Iterable<ComplexAOI> iterableOfComplexAOIRelations = StreamSupport
+                .stream(atlas.relations().spliterator(), false).map(ComplexAOI::getComplexAOI)
```
MikeGost: Just curious, what if you set the parallel flag to true here, does that have any impact on results or runtime?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java
sayas01: This Exception is caught in the method [getComplexAOI](https://github.com/osmlab/atlas/blob/5d67cea598acb2f6dc6fd0ed8237e321ac46f204/src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java#L63) and the Relation is logged there. Would you prefer to have Relation identifier here as well?

MikeGost: Good point. As long as it's logged somewhere in the stack, that works for me :)

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOIFinder.java
sayas01: I tested the run time with sequential and parallel stream for a Test Atlas and found that the run time for parallel stream to be  smaller (by a couple of ms) than the sequential. It would vary with machine and Atlas. For now, with the result I got, I have updated the stream to be parallel.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/aoi/ComplexAOI.java
sayas01: I have updated this to call `getComplexAOI(source,  customAoiFilter -> false)` which would remove the duplicated code from this method. Please let me know if this is what you meant by your comment.

MikeGost: That's perfect!

MikeGost:
Needs a spotless update.

PR #323, opened 2019-01-15
Relation geojson geometry actually represents

jklamer:
### Description:

All relations were getting their geometry sent to 0,0 where a more accurate geojson location would be the center of their bounds. 

### Potential Impact:

More accurate relation geometry location approximations. 

### Unit Test Approach:

Ran all unit tests. 

### Test Results:

Tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
@@ -18,7 +18,6 @@ import java.util.stream.Stream;
 import org.openstreetmap.atlas.exception.CoreException;
 import org.openstreetmap.atlas.geography.GeometricSurface;
 import org.openstreetmap.atlas.geography.Located;
-import org.openstreetmap.atlas.geography.Location;
 import org.openstreetmap.atlas.geography.MultiPolygon;
 import org.openstreetmap.atlas.geography.Polygon;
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -308,7 +307,7 @@ public LocationIterableProperties toGeoJsonBuildingBlock()
         final Optional<String> shardName = getAtlas().metaData().getShardName();
         shardName.ifPresent(shard -> tags.put("shard", shard));
 
-        return new GeoJsonBuilder.LocationIterableProperties(Location.CENTER, tags);
+        return new GeoJsonBuilder.LocationIterableProperties(bounds().center(), tags);
```
Bentleysb: This is a great improvement! I was just thinking that it might be even better if it matched how `Relation.asGeoJsonGeometry()` handles relation geometry.

jklamer: I agree. The only issue is that ```LocationIterableProperties``` takes a ```Iterable<Location>``` which cannot represent our MultiPolygon object, which the other Geojson representation makes if possible. This is part of the reason  ```LocationIterableProperties``` is deprecated and we're working it out, but while we are it should work as intended.

PR #331, opened 2019-01-26
Tame too zealous validation on reverse edge

matthieun:
### Description:

Atlas allows tags to be different on forward and backward edge, this PR makes sure validation allows for that. It now checks for geometry only, while it previously failed for any non-geometry difference between forward and backward edge.

### Potential Impact:

The `atlas atlas-diff` command should behave properly now.

### Unit Test Approach:

Modified one unit test to remove exception expected, and added two new use cases.

### Test Results:

No other tests than unit-tests.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeValidator.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeValidator.java
@@ -1,6 +1,6 @@
 package org.openstreetmap.atlas.geography.atlas.change.validators;
 
-import java.util.Map;
+import java.util.Optional;
 import java.util.function.BiPredicate;
 
 import org.openstreetmap.atlas.exception.CoreException;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeValidator.java
```diff
@@ -59,11 +59,26 @@ protected void validateReverseEdgesHaveForwardMatchingCounterpart()
...
+                            .changeFor(ItemType.EDGE, forwardEdgeIdentifier);
+                    if (backwardGeometryChange && !forwardFeatureChangeOption.isPresent())
+                    {
+                        throw new CoreException(
+                                "Backward edge {} is {} but does not have a forward edge change reference present.",
+                                backwardEdgeIdentifier, backwardFeatureChange.getChangeType());
+                    }
+                    else if (!backwardGeometryChange && !forwardFeatureChangeOption.isPresent())
+                    {
+                        // This case is a non-geometry change on the backward edge only, which can
```
lucaspcram: Woo! Good catch.

PR #335, opened 2019-02-01
Located fully within

yazad3:
### Description:

Added a new ```Containable``` interface which supports ```within(GeometricSurface)``` in line with ```Relation```. This is now implemented by basic geometric shapes like,

- Location
- PolyLine

And by entities like,

- LocationItem (super class of Node and Point)
- LineItem (super class of Edge and Line)
- Relation (no change, already has within)
- Area

### Potential Impact:
None, new method added in existing classes.

### Unit Test Approach:
Unit tested the within method on above geometries and entities.

### Test Results:
The ```within``` method now offers an intutive means of checking if the said entity is within a certain ```GeometricSurface```.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/Containable.java
```diff
@@ -0,0 +1,21 @@
+package org.openstreetmap.atlas.geography.atlas.items;
+
+import org.openstreetmap.atlas.geography.GeometricSurface;
+
+/**
+ * Represents entities that can be contained inside a {@link GeometricSurface}.
+ *
+ * @author Yazad Khambata
+ */
+public interface Containable
```
yiqingj: this is just my personal design opinion but all located features should be containable. So I don't see enough value to define an interface for it.

yazad3: The original plan was to implement this change in Located interface. However, the implications of doing so was wide reaching with about ~50 classes directly/indirectly needed an implementation for within. In some cases the implementation was not obvious, so discussed with @lucaspcram and added a new interface, which was a safe option that met the goal.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/LocationItem.java
```diff
@@ -19,7 +19,7 @@
  *
  * @author matthieun
  */
-public abstract class LocationItem extends AtlasItem
+public abstract class LocationItem extends AtlasItem implements Containable
```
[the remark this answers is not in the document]
yazad3: Good suggestion, done!

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Containable.java
lucaspcram: I agree that this is not an ideal decision from a design standpoint - but in the interest of getting this functionality off the ground it makes some sense. I think it may be worth creating a TODO item to eventually change this to `Located`.

yazad3: @lucaspcram and @yiqingj - how about we add `within` to `Located` with a default implementation that checks based on `Located#bounds()`. If a `Located` implementation wants to improve the within check it was be overwrite this change there.

lucaspcram: Oh that's actually a great idea. Then you can only implement the more granular functionality where you need it. Just make sure this is clear in the doc comments, so future users can implement a more granular check should they need it.

yiqingj: @lucaspcram @yazad3 the bounds implementation could be a temporary solution. Just from an end user point of view, if I were a user to use this method, I'd rather see it unimplemented than having unexpected behavior.  `Within` to me has a very clear definition and using bound will cause false negative result, imagining a user not paying close attention to the doc comments and has to debug why the result is not correct and eventually found out it's caused by this, I would definitely be angry if I were that person. :)
that being said having a `withinBound` method doesn't sound like a bad idea.

yazad3: @yiqingj and @matthieun - Thanks I am adding a default implementation that throws `UnsupportedOperationException`.

PR #339, opened 2019-02-07
Deprecated a constructor in InputStreamResource

lucaspcram:
### Description:

The newly deprecated constructor often caused gotchas when reading atlases and other resources, since it created a read-once `InputStreamResource`

### Potential Impact:

N/A, other than the fact that some downstream code may now get notices they are using a deprecated API.

### Unit Test Approach:

N/A

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/streaming/resource/InputStreamResource.java
+++ b/src/main/java/org/openstreetmap/atlas/streaming/resource/InputStreamResource.java
@@ -21,7 +21,8 @@ public class InputStreamResource extends AbstractResource
      * to many gotchas, since the resource becomes read-once. For e.g. loading an atlas from an
      * InputStreamResource will fail, since atlas uses an under-the-hood read to determine the
      * serialization format.
-     *
+     * 
+     * @deprecated
      * @param input
      *            the input stream
      */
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/streaming/resource/InputStreamResource.java
```diff
@@ -16,11 +16,28 @@
 {
     private final Supplier<InputStream> inputStreamSupplier;
 
+    /**
+     * This constructor is deprecated. Creating an InputStreamResource using a raw InputStream leads
+     * to many gotchas, since the resource becomes read-once. For e.g. loading an atlas from an
+     * InputStreamResource will fail, since atlas uses an under-the-hood read to determine the
+     * serialization format.
+     *
```
matthieun: This needs to have `@deprecated` in the javadoc too. [See sonar](
https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWjI-m_sChQW8w82rtlG&open=AWjI-m_sChQW8w82rtlG&pullRequest=339).

lucaspcram: Ah ok. The Eclipse plugin did not catch this.

PR #341, opened 2019-02-08
Dynamic slicing for relations

adahn6:
### Description:

This PR contains one big feature addition, as well as several smaller changes in support of that feature. First, the big picture: this PR adds in dynamic expansion during the relation slicing step so that all coastline relations are completely loaded and country sliced appropriately. Note that this operates only on multipolygon, coastline tagged relations. The end result is that all multipolygon water bodies defined by coastlines (such as inland lakes and seas) are now appropriately sliced by country, ensuring each country is appropriately assigned the portion of the water body that overlap with that country's borders. 

In order to support this feature, a number of important changes were made, listed in no particular order below:

- Country slicing is now entirely divided into two independent steps, point-and-line slicing and relation slicing. This is because when dynamically expanding, the DynamicAtlas needs to be able to load in point-and-line sliced data from other shards; now, all shards can be point-and-line sliced before relation slicing begins so that data can be loaded during relation slicing.

- All sliced line identifier logic has changed. Previously, line identifiers generated during slicing were deterministically generated, but not deterministically assigned. The end result is that while each shard would have line segments numbering the same (e.g. 1 2 and 3), they were non-deterministically assigned based on the shard being processed; this means that two shards could contain different lines with the same identifier. To fix this, when slicing lines, all line identifiers are generated ahead of time, then organized into queues for each country code based on alphabetical order. So if country A has two line slices and country B has one, then a queue would be generated for each country with country A's queue containing ids [1,2] and country B's containing [3]. This guarantees that line slicing will always assign the same line slice the same identifier. This was critical to ensure proper MultiAtlas support when dynamically loading coastline relations from other country's shards.

- Similarly sliced point identifier logic has changed. Now, all SyntheticBoundaryNodes generated during line slicing will use the *last three digits* of the identifier, based on the sliced id as defined above. Previous logic used the parent line id, not the sliced line identifier, and resulted in multiple points being assigned the same identifier, which again caused issues for MultiAtlas. 

- ComplexWaterFinder previously did not have support for the SEA tag, so that has been added in along with relevant unit tests.

- There are now no filters on what relations are brought in, and certain boundaries relating to water and AOIs are now brought in-- this has changed the expected values in a number of tests.

- A new cut type has been introduced, the SilkCut. This cut type is the same as a SoftCut, but preserves any point definitions on lines being included in the soft cut. This change is important because WaySectioning relies on Point definitions in the underlying Atlas, and was breaking when dealing with the new data on Atlases that had been dynamically sliced.

### Potential Impact:

Identifiers have changed, so anything relying on consistent IDs may need to be updated. Additionally, filters have been expanded to allow more entities in, so anything relying on entity counts may need to be updated.

### Unit Test Approach:

Unit tests have been updated for entity counts with the expanded filters, and SEA tagged entities have been included in the basic ComplexWater tests. Additionally, water relation tests have been added to the RelationSlicingTest to include both water relations that should be dynamically expanded on, and non-water relations that should be ignored by the filter.

### Test Results:

------


gradle.properties
```diff
@@ -1,6 +1,11 @@
 group=org.openstreetmap.atlas
 version=5.3.6-SNAPSHOT
 
+# a trick to make our distribution package follow the same naming convention.
+releaseBuild=false
+artifactory_repoKey=libs-snapshot-local
+>>>>>>> d60705a... Added in dynamic exploration of Relations during slicing
```
jwpgage: leftover from merge conflict?

adahn6: whoops! good catch

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/waters/handler/SeaHandler.java
```diff
@@ -0,0 +1,58 @@
...
+    private static final Logger logger = LoggerFactory.getLogger(SeaHandler.class);
+
+    public static boolean isSea(final AtlasEntity atlasEntity)
+    {
+        /*
+         * (1) natural == water && water == "sea"
+         */
+        return Validators.isOfType(atlasEntity, NaturalTag.class, NaturalTag.WATER)
+                && Validators.isOfType(atlasEntity, WaterTag.class, WaterTag.SEA);
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+                    NaturalTag.COASTLINE);
+
+    // Dynamic expansion filter will be a combination of the water relations and points
+    private static final Predicate<AtlasEntity> dynamicAtlasExpansionFilter = entity -> pointPredicate
+            .test(entity) || relationPredicate.test(entity);
+
     /**
-     * Slices against the given country and boundary map. Note: Will assume the entire world to be
-     * the target MultiPolygon to use.
+     * The default constructor for the old, pre-water relation pipeline-- this method will slice
```
MikeGost: Would remove "pipeline" here..

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
             final CountryBoundaryMap countryBoundaryMap)
     {
         this.countries.addAll(countries);
         this.countryBoundaryMap = countryBoundaryMap;
+        this.sharding = null;
+        this.atlasFetcher = null;
+    }
+
+    /**
+     * Updated constructor for the water relation pipeline-- this method will use the Atlas fetcher
```
MikeGost: Same pipeline mention here.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+     * function and sharding tree to dynamically expand on all water relations so they can be sliced
+     * appropriately.
+     *
+     * @param country
+     *            The list of countries to slice. Should really just be one country?
```
MikeGost: Should really just be one country should be removed?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+     *            before returning.
+     */
+    public RawAtlasCountrySlicer(final Set<String> country,
+            final CountryBoundaryMap countryBoundaryMap, final Sharding sharding,
+            final Function<Shard, Optional<Atlas>> atlasFetcher)
+    {
+        if (sharding == null || atlasFetcher == null)
+        {
+            throw new IllegalArgumentException(
+                    "Must supply a valid sharding and fetcher function for sectioning!");
```
MikeGost: Sectioning --> Slicing

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+     *            the fetcher should MultiAtlas these line sliced country Atlas files together
+     *            before returning.
+     */
+    public RawAtlasCountrySlicer(final String country, final CountryBoundaryMap countryBoundaryMap,
+            final Sharding sharding, final Function<Shard, Optional<Atlas>> atlasFetcher)
+    {
+        if (sharding == null || atlasFetcher == null)
+        {
+            throw new IllegalArgumentException(
+                    "Must supply a valid sharding and fetcher function for sectioning!");
```
MikeGost: Sectioning --> Slicing

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+     *            The country boundary map
+     * @param sharding
+     *            The sharding tree
+     * @param atlasFetcher
+     *            A function return an atlas for a given shard. NOTE: The function expects line
+     *            sliced Atlases. And since one shard can have multiple country sliced Atlas files,
+     *            the fetcher should MultiAtlas these line sliced country Atlas files together
+     *            before returning.
+     */
+    public RawAtlasCountrySlicer(final Set<String> country,
```
MikeGost: Can this constructor be combined with the one on line 155? They are doing identical work, one is for one country and another for a set of countries.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -252,17 +254,46 @@ else if (slicesBelongToSingleCountry(slices)
...
             final CountrySlicingIdentifierFactory lineIdentifierFactory = new CountrySlicingIdentifierFactory(
                     line.getIdentifier());
-            final CountrySlicingIdentifierFactory pointIdentifierFactory = new CountrySlicingIdentifierFactory(
-                    line.getIdentifier());
 
-            final List<TemporaryLine> createdLines = new ArrayList<>();
+            final Map<String, Queue<Long>> lineIdsByCountry = new HashMap<>();
+            final List<String> countryCodesForSlices = new ArrayList<>();
+
+            // Before processing slices, sort them and record country codes. Then, alphabetize the
```
MikeGost: Awesome commenting!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -252,17 +254,46 @@ else if (slicesBelongToSingleCountry(slices)
...
+            // country codes and pre-generate identifiers assigned to the country codes. This way,
+            // all lines are deterministically assigned the same identifier regardless of what order
+            // the slices are returned in or which country is slicing the line.
+            for (final Geometry slice : slices)
+            {
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -252,17 +254,46 @@ else if (slicesBelongToSingleCountry(slices)
...
+                    lineIdsByCountry.put(countryCode, idsForCountry);
+                }
+            });
 
+            final List<TemporaryLine> createdLines = new ArrayList<>();
             try
             {
                 for (final Geometry slice : slices)
                 {
+
```
MikeGost: Nit: extra

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -328,8 +367,62 @@ private long createNewPointIdentifier(
...
+                    return false;
+
+                }, AtlasCutType.SOFT_CUT)
+                        .orElseThrow(() -> new CoreException(
+                                "Cannot have an empty atlas after way sectioning {}",
+                                this.initialShard.getName()));
+
+                return countrySubAtlas.subAtlas(this.initialShard.bounds(), AtlasCutType.SILK_CUT)
+                        .orElseThrow(() -> new CoreException(
+                                "Cannot have an empty atlas after way sectioning {}",
```
MikeGost: After relation slicing**

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -328,8 +367,62 @@ private long createNewPointIdentifier(
...
+                            {
+                                return true;
+                            }
+                        }
+                    }
+                    return false;
+
+                }, AtlasCutType.SOFT_CUT)
+                        .orElseThrow(() -> new CoreException(
+                                "Cannot have an empty atlas after way sectioning {}",
```
MikeGost: After relation slicing**

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -576,6 +669,28 @@ private String getShardOrAtlasName()
...
     }
 
+    /**
+     * Check if the {@link Geometry} should be filtered out based on the provided bound.
+     *
+     * @param geometry
+     *            The {@link Geometry} to check.
+     * @return {@code true} if the given geometry should be filtered out.
+     */
+    private boolean isOutsideWorkingBound(final Geometry geometry)
```
MikeGost: This looks identical to the method in `RawAtlasPointAndLineSlicer` - consider pulling it out from both into the `RawAtlasSlicer`.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -907,6 +1022,8 @@ private void preProcessMultiPolygonRelation(final Relation relation)
         mergeOverlappingClosedMembers(relation, outers, inners);
     }
 
+    // TODO come back and verify we're keeping track of all required statistics
```
MikeGost: Heh, we should at some point come back to this.. :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -153,10 +153,10 @@ protected static boolean fromSameCountry(final Line one, final Line two)
             firstCountries.retainAll(new HashSet<>(secondCountries));
             return !firstCountries.isEmpty();
         }
-
-        throw new CoreException(
```
MikeGost: Quick question - why was this taken out?

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/AtlasCutType.java
```diff
@@ -32,6 +32,8 @@
  * its members will be included in the member list, even if not matched by the given matcher.
  * </ul>
  * </li>
+ * <li>Silk-cut: Same as a soft-cut, but always soft-cuts relations and keeps all points on line
```
MikeGost: Cool concept!

src/main/resources/org/openstreetmap/atlas/geography/atlas/pbf/osm-pbf-relation.json
```diff
@@ -1,6 +1,4 @@
 {
     "filters": [
-    	"natural->!coastline",
```
MikeGost: Nice to this gone!!

src/test/java/org/openstreetmap/atlas/tags/filters/ConfiguredTaggableFilterTest.java
```diff
@@ -58,7 +58,7 @@ public void testDefaultOsmosisRelationConfiguration()
     {
         final Resource relationResource = new InputStreamResource(
```
MikeGost: Do you want rename the resource to match too?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
adahn6: ah, was a clunky attempt to reference the entrypoints for relation slicing. will reword to be more specific.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
adahn6: hmm, not sure, this wasn't a very clear way of expressing this. with the decoupling of point-and-line slicing and relation slicing, it seems like the most well-supported code paths are to country slice each shard one country at a time. that's not to say multiple country slicing doesn't work, but rather to say that we should probably think about deprecating it?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: good question-- I found it was causing some issues during the loading of the point-and-line sliced shard data from other countries. but things have changed a lot since then, so I'll test and see if I need this removed still.

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/AtlasCutType.java
adahn6: it was between SilkCut and SofterCut! 😂

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
adahn6: good idea!

src/test/java/org/openstreetmap/atlas/tags/filters/ConfiguredTaggableFilterTest.java
adahn6: oh, good catch

adahn6: on closer examination, just going to remove this test entirely. the default filter no longer filters anything out, so I don't think there's a real need to test it, especially since filtering behavior is already being tested by the way filter configuration test below.

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/SubAtlasCreator.java
```diff
@@ -401,6 +401,166 @@ public SubAtlasCreator(final Atlas atlas)
         return Optional.ofNullable(result);
     }
 
+    @SuppressWarnings("squid:S3776")
+    @Override
+    public Optional<Atlas> silkCut(final GeometricSurface boundary)
```
lucaspcram: I know this has been discussed but I do wonder if it is worth diving into the `SubAtlasCreator` class and factoring out common behaviour. These methods are huge and a lot of the stuff seems similar.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -328,8 +367,62 @@ private long createNewPointIdentifier(
...
+     *            The {@link Atlas} file we need to trim
+     * @return the {@link Atlas} for the bounds of the input shard
+     */
+    private Atlas cutSubAtlasForOriginalShard(final Atlas atlas)
+    {
+        try
+        {
+            if (this.initialShard != null)
+            {
+
```
matthieun: Whitespace

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -328,8 +367,62 @@ private long createNewPointIdentifier(
...
+                final Atlas countrySubAtlas = atlas.subAtlas(entity ->
```
matthieun: Maybe put that lambda in a variable a line above for readability? I was initially surprised by the SOFT_CUT in the middle line 408 :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
matthieun: Sonar will do it for us!

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/SubAtlasCreator.java
```diff
@@ -401,6 +401,166 @@ public SubAtlasCreator(final Atlas atlas)
         return Optional.ofNullable(result);
     }
 
+    @SuppressWarnings("squid:S3776")
```
matthieun: Hmmmm? What is that? How about we keep the sonar errors, so at least we can see where they are later? :D

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
     private final Set<String> countries = new HashSet<>();
 
+    private final List<Shard> loadedShards = new ArrayList<>();
+
+    // Bring in all points -- this is important for proper sectioning later downstream, otherwise we
+    // will bring in geometry definitions with no underlying points
+    private static final Predicate<AtlasEntity> pointPredicate = entity -> entity instanceof Point;
+
+    // Bring in all relations that are tagged as Water or Coastline
+    private static final Predicate<AtlasEntity> relationPredicate = entity -> entity.getType()
```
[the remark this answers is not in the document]
adahn6: that is a great idea!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+     * @param initialShard
+     *            The initial shard being sliced
+     * @return A fully sliced (lines and relations) Atlas with water relations expanded upon
+     */
+    public Atlas sliceRelations(final Shard initialShard)
+    {
+        final Time time = Time.now();
+        final Atlas partiallySlicedExpandedAtlas = buildExpandedAtlas(initialShard);
+        final String shardName = getShardOrAtlasName(partiallySlicedExpandedAtlas);
+        logger.info("Started relation slicing for shard {}", shardName);
```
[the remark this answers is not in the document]
adahn6: yeah thinking about that now, it makes more sense to predefine. I'll refactor that now

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -1,102 +1,311 @@
...
+                initialShard.getName());
+
+        // Keep track of all loaded shards. This will be used to cut the sub-atlas for the shard
+        // we're processing after all slicing is completed. Initial shard will always be first!
+        this.loadedShards.add(initialShard);
+
+        final DynamicAtlasPolicy policy = new DynamicAtlasPolicy(this.atlasFetcher, this.sharding,
+                initialShard, Rectangle.MAXIMUM).withDeferredLoading(true)
+                        .withExtendIndefinitely(false).withAggressivelyExploreRelations(true)
+                        .withAtlasEntitiesToConsiderForExpansion(dynamicAtlasExpansionFilter::test);
```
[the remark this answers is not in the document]
adahn6: probably not!

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/SubAtlasCreator.java
adahn6: ok, sure!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -237,6 +260,11 @@ private void createNewLineMemberForRelation(final LineString lineString,
         final List<Long> newLineShapePoints = new ArrayList<>(lineString.getNumPoints());
         final Coordinate[] lineCoordinates = PRECISION_REDUCER.edit(lineString.getCoordinates(),
                 lineString);
+
+        if (isOutsideWorkingBound(lineString))
```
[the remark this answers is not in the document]
adahn6: I'll update comments to make this clearer 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
adahn6: good idea. I also improved the comments here explaining why there are two cuts, and the purpose of SILK_CUT vs SOFT_CUT

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/AtlasCutType.java
lucaspcram: I am actually having some trouble understand exactly the difference between `silk-cut` and `soft-cut`. Would you mind giving me a bit more explanation, and maybe fleshing out the description here a bit more? Thanks.

adahn6: absolutely! I'll update this now

adahn6: to be clear here as well as in the comments-- a soft cut preserves line geometry that crosses shard boundaries, so long as it starts in the shard. the problem is, the points for the coordinates of that line don't follow that same logic; if the point is inside the boundary, it's preserved, otherwise it's removed. this is a problem later in way sectioning, because that logic looks at the line geometry, then often tries to find the underlying Point for line coordinates. if it can't find a Point, it breaks. 

Before this, we weren't using SubAtlas until after way sectioning, so it wasn't a problem. however, now, we're loading in lines that cross shard boundaries, then soft cutting to keep line geometry but removing the underlying points. SILK_CUT is a new type that will search each line being kept and flag any Points along that line as worth keeping, even if they're outside the shard boundary. Does that make more sense?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -375,7 +381,15 @@ private long createNewPointIdentifier(
...
-     * Cuts out data outside the shard boundary, since we don't need any data beyond that.
+     * Used to cut out unnecessary data loaded in during DynamicAtlas exploration from the final
+     * shard. First, a subatlas is called to remove lines loaded in from countries not being
+     * sliced-- this is important because when dynamically exploring water relations, line sliced
+     * data from other countries is loaded so the whole relation can be loaded. That data is not
+     * necessary once the relation has been sliced, so it is removed. Then, a separate subatlas call
+     * is made to trim out geometry outside the shard bounds. Note that SILK_CUT is used so that any
+     * lines that expand past the borders of the shard have their underlying point geometries kept.
+     * This inclusion is critical for way-sectioning, which needs all lines to have their points for
+     * purposes of edge detection.
```
adahn6: @lucaspcram added extra comments for SILK_CUT here!

src/main/java/org/openstreetmap/atlas/geography/atlas/sub/AtlasCutType.java
```diff
@@ -32,8 +32,13 @@
  * its members will be included in the member list, even if not matched by the given matcher.
  * </ul>
  * </li>
- * <li>Silk-cut: Same as a soft-cut, but always soft-cuts relations and keeps all points on line
- * geometries kept by the slice.
+ * <li>Silk-cut: Largely the same as a Soft-cut, however there is one important difference. During
```
adahn6: @lucaspcram added extra comments for SILK_CUT here!

PR #346, opened 2019-02-14
Support YAML formatted config files

gitclonefun:
### Description:

- Add support for YAML formatted config files for both Standard and Merged configurations.

- Add a path to load the config with the specified config type or unknown which will attempt to load all supported formats

### Potential Impact:

- Config files can be JSON or YAML now.  Note that we can support any combination of JSON or YAML files.

- If the unknown format type is defined, the order attempted to parse is JSON then YAML.  So if you use YAML files we will have to try and fail to parse as JSON before parsing as YAML.

### Unit Test Approach:

- [x] created yaml version of all json test configs
- [x] copied all tests that used json configs and used yaml configs
- [x] added tests that mix json and yaml configs
- [x] added tests to make sure the specific format and try-all-formats works

### Test Results:

- [x] All Ok

### Example Config
In JSON
```
{
  "feature": {
    "ABC": {
      "enable": true,
      "list": [
        "nodes",
        "edges"
      ],
      "range": {
        "max": 20.0,
        "min": 10.0
      }
    }
  }
}
```

In  YAML
```
feature:
  ABC:
    enable: true
    list:
      - nodes
      - edges
    range:
      max: 20.0
      min: 10.0
```
------


src/test/resources/org/openstreetmap/atlas/utilities/configuration/developmentOverriding.yml
```diff
@@ -0,0 +1,3 @@
+feature.range:
+  max: 35.0
+  override.ABC.max: 70.0
```
lucaspcram: tiny nit: ending newline
Same nit for 3 other checked-in files.

src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
```diff
@@ -89,21 +105,39 @@ private StandardConfigurable(final String key, final Raw defaultValue,
...
+                        .orElseThrow(() -> new CoreException("Unable to load JSON configuration."));
+                return;
+            case YAML:
+                this.configurationData = this.readConfigMapFromYAML(configBytes)
+                        .orElseThrow(() -> new CoreException("Unable to load YAML configuration."));
+                return;
+            case UNKNOWN:
+            default:
+                // If the config format is unknown, attempt to load the config with each format
+                // until one finds some data
```
lucaspcram: Very cool.

src/test/resources/org/openstreetmap/atlas/utilities/configuration/keywordOverridingApplication.yml
```diff
@@ -0,0 +1,14 @@
...
+    - edges
+  range:
+    max: 20.0
+    min: 10.0
+  override:
+    XYZ:
+      range:
+        max: 30.0
+    ABC.list:
+      - nodes
```
lucaspcram: ending new line

src/test/resources/org/openstreetmap/atlas/utilities/configuration/yaml_dot_compressed.yml
```diff
@@ -0,0 +1,2 @@
+a.b.c: 0
+a.b.d: 1
```
lucaspcram: ending newline

src/test/resources/org/openstreetmap/atlas/utilities/configuration/yaml_dot_expanded.yml
```diff
@@ -0,0 +1,4 @@
+a:
+  b:
+    c: 0
+    d: 1
```
lucaspcram: ending newline

gitclonefun:
Thanks for taking a look @lucaspcram 

Updated all the test .yml files to have ending newlines.

dependencies.gradle
```diff
@@ -20,6 +20,7 @@ project.ext.versions = [
     jsonassert:'1.3.0',
     jackson_core:'2.8.8',
     jackson_databind:'2.8.8.1',
+    jackson_dataformat_yaml:'2.8.8',
```
matthieun: Any transitive dependencies that are brought in that could conflict with existing ones?

src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
```diff
@@ -222,4 +256,54 @@ private Object resolve(final String key, final Map<String, Object> currentContex
         }
         return null;
     }
+
+    @SuppressWarnings("unchecked")
+    private Optional<Map<String, Object>> readConfigMapFromJSON(final byte[] readBytes)
```
matthieun: `Config` -> `Configuration` to avoid abbreviations.

src/test/java/org/openstreetmap/atlas/utilities/configuration/MergedConfigurationTest.java
```diff
@@ -18,33 +18,61 @@
...
-    private static final String KEYWORD_OVERRIDDEN_DEV_CONFIGURATION = "org/openstreetmap/atlas/utilities/configuration/developmentOverriding.json";
-    private static final String OVERRIDE_CONFIGURATION = "org/openstreetmap/atlas/utilities/configuration/development.json";
-    private static final String PARTIAL_CONFIGURATION = "org/openstreetmap/atlas/utilities/configuration/feature.json";
+    private static final String KEYWORD_OVERRIDDEN_CONFIGURATION_JSON = "org/openstreetmap/atlas/utilities/configuration/keywordOverridingApplication.json";
+    private static final String KEYWORD_OVERRIDDEN_CONFIGURATION_YAML = "org/openstreetmap/atlas/utilities/configuration/keywordOverridingApplication.yml";
+
+    private static final String KEYWORD_OVERRIDDEN_DEV_CONFIGURATION_JSON = "org/openstreetmap/atlas/utilities/configuration/developmentOverriding.json";
+    private static final String KEYWORD_OVERRIDDEN_DEV_CONFIGURATION_YAML = "org/openstreetmap/atlas/utilities/configuration/developmentOverriding.yml";
+
+    private static final String OVERRIDE_CONFIGURATION_JSON = "org/openstreetmap/atlas/utilities/configuration/development.json";
```
matthieun: Would it be nicer to use `MergedConfigurationTest.class.loadResourceAsStream` instead below, so the package paths do not have to be hardcoded here? It would help in any future refactor.

src/test/java/org/openstreetmap/atlas/utilities/configuration/MergedConfigurationTest.java
```diff
@@ -60,101 +88,209 @@ public void testDotFormat() throws IOException
...
+                InputStream override = loader.getResourceAsStream(OVERRIDE_CONFIGURATION_JSON))
+        {
+            testLayered(base, override);
+        }
+    }
+
+    private void testLayered(final InputStream base, final InputStream override)
+    {
+
+        final Configuration configuration = new MergedConfiguration(new InputStreamResource(base),
```
matthieun: If that base input stream will be read more than once, it might cause issues here. It might be better to pass around a supplier that re-creates a new stream at every call, and then pass that supplier to the InputStreamResource constructor.

src/main/java/org/openstreetmap/atlas/utilities/configuration/StandardConfiguration.java
gitclonefun: 👍

dependencies.gradle
gitclonefun: Good question.  I think this is safe though.

As far as I can tell this depends on Jackson core (which we're already using) and SnakeYAML which we're not using anywhere else (so no version conflict) and has no additional dependencies.

gitclonefun:
@matthieun all comments addressed

- [x] moved to using `X.class.loadResourceAsStream` in Standard and Merged config tests and updated the various resource constants to just be the resource name without the path.

- [x] updated MergedConfigurationTest to pass around an InputStream supplier rather than creating the InputStreams directly.  I see now that the InputStreamResource has that approach deprecated so this fixes that warning.  One question about this approach though - is previously creating the InputStream was wrapped in a try block so we implicitly closed the streams - with the Supplier approach does that mean the caller of the Supplier.get is responsible for closing the stream (which I suppose makes more sense than being responsible for it at a different level than where you read it).

dependencies.gradle
matthieun: We will have to watch the version of Jackson core vs. the one that comes in Spark, for atlas-generator.... Thanks.

Issue #314, opened 2018-12-27
TaggableFilter: ! operator short circuits multi-value tests


Bentleysb:
The documentation for [TaggableFilters](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/filters/README.md) states that `water->!pond,!lake` should return true for "Anything without water=pond or water=lake". In attempting to implement a similar filter, `highway->!service,!constructions`, it did not appear that this hold true. In my case, the filter would only return false if highway was equal to service, instead of service or construction. I believe this is caused by line 101 in [Taggable](https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/Taggable.java), as it always returns a boolean for a value proceeded by a `!` regardless of there being any other values to check. 
As a note, it is possible to work around this by using the filter `highway->!service&highway->!construction`.

matthieun:
Hi @Bentleysb just to be sure. The filter you pasted says construction"s". Was that a typo while writing that issue?

Bentleysb:
@matthieun, yes that is just a typo in the issue description.

matthieun:
So the filter was misbehaving, but the description was wrong too. `highway->!service&highway->!construction` is the real way to go to achieve what you want. The attached PR should make that clear. Thanks!

PR #355, opened 2019-02-19
TaggableFilter with double negative

matthieun:
### Description:

Fix for issue #314.

The scheme `A->X,Y` means the tag `A` has value `X` OR value `Y`. In case of double negative (`A->!X,!Y`), the result is always true as it is equivalent "Tag A has (NOT value X OR NOT value Y)" thus "Tag A has NOT (value X AND value Y)". Because "Tag A has (value X AND value Y)" is impossible and always false, the former is always true.

### Potential Impact:

This might impact everything using `TaggableFilter` that might have benefited from that silent error. (`AtlasStatistics`, `ConfiguredTaggableFilter`)

### Unit Test Approach:

Added one extra unit test.

### Test Results:

Ran integration tests and they pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/tags/filters/README.md
+++ b/src/main/java/org/openstreetmap/atlas/tags/filters/README.md
@@ -58,6 +58,12 @@ water->!pond,!lake
 Anything without water=pond or without water=lake. That one is not very useful as it is always true because one of the two will always be true (The water tag can have only one value).
 
 ```
+water->!pond&water->!lake
+```
+
+Anything without water=pond and without water=lake.
+
+```
 water->!
 ```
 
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/tags/filters/README.md
```diff
@@ -55,7 +55,7 @@ Anything without water=pond
 water->!pond,!lake
 ```
 
-Anything without water=pond or water=lake
+Anything without water=pond or without water=lake. That one is not very useful as it is always true because one of the two will always be true (The water tag can have only one value).
```
Bentleysb: It might be nice to have an example that shows how to properly do this with `water->!pond&water->!lake`.

matthieun: Let me do that!

PR #363, opened 2019-02-22
Remove bloated from non-test code

yazad3:
### Description:

Noticed some of the `toString` and error messages were printing _Bloated_ instead of _Complete_. So cleaned it up. In the process refactored some of the method named in `AtlasDiff`. Several tests still use the word bloat which is not ceaned up.

### Potential Impact:

None.

### Unit Test Approach:

None, just renamed code and corrected String literals. Existing test cases to catch issues if any.

### Test Results:

Existing test succeeded.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
@@ -157,7 +157,7 @@ public class AtlasDiff
     }
 
     /**
-     * Saving all geometries means that we bloat features with their geometry even if they do not
+     * Saving all geometries means that we enrich features with their geometry even if they do not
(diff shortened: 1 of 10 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/diff/AtlasDiff.java
```diff
@@ -157,7 +157,7 @@ public Atlas getBeforeAtlas()
     }
 
     /**
-     * Saving all geometries means that we bloat features with their geometry even if they do not
+     * Saving all geometries means that we enrich features with their geometry even if they do not
```
matthieun: Nice :)

yazad3: Thanks

PR #365, opened 2019-02-25
bug fix in Polygon isClockwise method

snb3300:
### Description:
```isClockwise``` method in polygon checks the locations in order to infer the directionality. It uses the **sum over edges** [http://stackoverflow.com/questions/1165647](url) method to get the orientation. But this method expects the points on a cartesian plane and we are using earth coordinates directly which causes some of the cases to fail.

Example Polygon which fails:   
**POLYGON ((-70.0020146 12.5265405, -70.0019978 12.5265112, -70.0019564 12.526534, -70.0019733 12.5265633, -70.0020146 12.5265405))**

To fix this we need to convert lat/long into cartesian coordinates. Please refer [https://www.grasshopper3d.com/forum/topics/best-way-to-translate-latitude-longitude-data-into-xyz-points?commentId=2985220%3AComment%3A1804229](url)

The conversion is:
x = R * cos(lat) * cos(lon)  y = R * cos(lat) * sin(lon)

### Potential Impact:

This should not impact any downstream processes as we are not changing the geometry of polygons

### Unit Test Approach:
Wrote a unit test which was failing for the previous code.

The polygon in the unit test has a counter clockwise orientation but isClockwise was returning true previously.

### Test Results:

All the unit tests passed.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Polygon.java
@@ -20,6 +20,7 @@ import org.openstreetmap.atlas.utilities.collections.Iterables;
 import org.openstreetmap.atlas.utilities.collections.MultiIterable;
 import org.openstreetmap.atlas.utilities.scalars.Angle;
 import org.openstreetmap.atlas.utilities.scalars.Surface;
+import org.openstreetmap.atlas.utilities.tuples.Tuple;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -383,19 +384,22 @@ public class Polygon extends PolyLine implements GeometricSurface
      */
     public boolean isClockwise()
     {
-        long sum = 0;
-        long lastLatitude = Long.MIN_VALUE;
-        long lastLongitude = Long.MIN_VALUE;
-        for (final Location point : this)
+        double sum = 0;
+        Tuple<Double, Double> previousCartesianPoint = null;
+
+        for (final Location point : this.closedLoop())
         {
-            if (lastLongitude != Long.MIN_VALUE)
+            final Tuple<Double, Double> newCartesianPoint = getCartesianCoordinates(
+                    point.getLatitude(), point.getLongitude());
+
+            if (previousCartesianPoint != null)
             {
-                sum += (point.getLongitude().asDm7() - lastLongitude)
-                        * (point.getLatitude().asDm7() + lastLatitude);
+                sum += (newCartesianPoint.getFirst() - previousCartesianPoint.getFirst())
+                        * (newCartesianPoint.getSecond() + previousCartesianPoint.getSecond());
             }
-            lastLongitude = point.getLongitude().asDm7();
-            lastLatitude = point.getLatitude().asDm7();
+            previousCartesianPoint = newCartesianPoint;
         }
+
         return sum >= 0;
     }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/main/java/org/openstreetmap/atlas/geography/Polygon.java [hunk 1]
```diff
@@ -714,4 +718,29 @@ private void verifyVertexIndex(final int index)
             throw new CoreException("Invalid Vertex Index {}.", index);
         }
     }
+
+    /**
+     * This method will return cartesian equivqlent coordinates for earth coordinates
```
adahn6: nit: "equivalent"

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -714,4 +718,29 @@ private void verifyVertexIndex(final int index)
...
+     *      "https://www.grasshopper3d.com/forum/topics/best-way-to-translate-latitude-longitude-data-into-xyz-points?commentId=2985220%3AComment%3A1804229"></a>
+     * @see <a href=
+     *      "https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates"></a>
+     * @see <a href="https://en.wikipedia.org/wiki/Earth_radius"></a>
+     */
+    private Tuple<Double, Double> getCartesianCoordinates(final Latitude latitude,
+            final Longitude longitude)
+    {
+        // mean earth radius
+        final double earthRadius = 6371;
```
MikeGost: See `Distance. AVERAGE_EARTH_RADIUS`

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -714,4 +718,29 @@ private void verifyVertexIndex(final int index)
...
+
+        final double pointX = earthRadius * Math.cos(latitude.asRadians())
```
MikeGost: How about xCoordinate and yCoordinate?

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -714,4 +718,29 @@ private void verifyVertexIndex(final int index)
...
+     * This method will return cartesian equivalent coordinates for earth coordinates
+     * 
+     * @return Tuple which represents X and Y on catesian plane
+     * @see <a href=
+     *      "https://www.grasshopper3d.com/forum/topics/best-way-to-translate-latitude-longitude-data-into-xyz-points?commentId=2985220%3AComment%3A1804229"></a>
+     * @see <a href=
+     *      "https://en.wikipedia.org/wiki/Geographic_coordinate_conversion#From_geodetic_to_ECEF_coordinates"></a>
+     * @see <a href="https://en.wikipedia.org/wiki/Earth_radius"></a>
+     */
+    private Tuple<Double, Double> getCartesianCoordinates(final Latitude latitude,
```
MikeGost: Instead of Doubles, consider using `EarthCenteredEarthFixedCoordinate`.

MikeGost: You could add this as a static method to that class also.

snb3300: Actually, I don't even need a separate method here, directly used ```EarthCenterEarthFixedCoordinate``` object in the ```isClockwise``` method

PR #367, opened 2019-02-27
added a new pbf2raw command

adahn6:
### Description:

Adds in an atlas command for creating raw Atlas files from PBF files directly. 

### Potential Impact:

None.

### Unit Test Approach:

N/A

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
@@ -1,0 +1,177 @@
+package org.openstreetmap.atlas.utilities.command.subcommands;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.ArrayList;
+import java.util.List;
+
+import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas;
+import org.openstreetmap.atlas.geography.atlas.packed.PackedAtlas.AtlasSerializationFormat;
+import org.openstreetmap.atlas.geography.atlas.raw.creation.RawAtlasGenerator;
+import org.openstreetmap.atlas.geography.sharding.CountryShard;
+import org.openstreetmap.atlas.streaming.resource.File;
+import org.openstreetmap.atlas.streaming.resource.FileSuffix;
+import org.openstreetmap.atlas.utilities.command.AtlasShellToolsException;
+import org.openstreetmap.atlas.utilities.command.abstractcommand.CommandOutputDelegate;
+import org.openstreetmap.atlas.utilities.command.abstractcommand.OptionAndArgumentDelegate;
+import org.openstreetmap.atlas.utilities.command.parsing.ArgumentArity;
+import org.openstreetmap.atlas.utilities.command.parsing.ArgumentOptionality;
+import org.openstreetmap.atlas.utilities.command.parsing.OptionOptionality;
+import org.openstreetmap.atlas.utilities.command.subcommands.templates.MultipleOutputCommand;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author samgass
+ */
+public class PbfToRawAtlasCommand extends MultipleOutputCommand
+{
+    private static final Logger logger = LoggerFactory.getLogger(PbfToRawAtlasCommand.class);
+
+    // The filetype for PBFs-- will strip this from the filename to make the Atlas filename
+    private static final String PBF_FILETYPE = ".pbf";
+
+    // The hint for the input path for the PBF file(s) to convert
+    private static final String PBF_PATH_HINT = "pbf";
+
+    // The country name for the country shards Atlas file(s) to output
+    private static final String COUNTRY_NAME = "countryName";
+
+    private static final String COUNTRY_NAME_DESCRIPTION = "The country for the shard to build";
+
+    private final OptionAndArgumentDelegate optionAndArgumentDelegate;
+    private final CommandOutputDelegate outputDelegate;
+
+    private List<File> pbfs;
+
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * @author samgass
+ */
+public class PbfToRawAtlasCommand extends MultipleOutputCommand
+{
+    private static final Logger logger = LoggerFactory.getLogger(PbfToRawAtlasCommand.class);
+
+    // The filetype for PBFs-- will strip this from the filename to make the Atlas filename
+    private static final String PBF_FILETYPE = ".pbf";
```
lucaspcram: Check out the `FileSuffix` enum for this. `FileSuffix.PBF.toString()`

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
```diff
@@ -0,0 +1,177 @@
...
+import org.openstreetmap.atlas.utilities.command.parsing.ArgumentOptionality;
+import org.openstreetmap.atlas.utilities.command.parsing.OptionOptionality;
+import org.openstreetmap.atlas.utilities.command.subcommands.templates.MultipleOutputCommand;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * @author samgass
+ */
+public class PbfToRawAtlasCommand extends MultipleOutputCommand
```
lucaspcram: :D

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
```diff
@@ -0,0 +1,177 @@
...
+    @Override
+    public void registerOptionsAndArguments()
+    {
+        final Integer[] contexts = this.optionAndArgumentDelegate.getFilteredRegisteredContexts()
+                .toArray(new Integer[0]);
+        this.registerArgument(PBF_PATH_HINT, ArgumentArity.VARIADIC, ArgumentOptionality.REQUIRED,
+                contexts);
+        this.registerOptionWithRequiredArgument(COUNTRY_NAME, COUNTRY_NAME_DESCRIPTION,
+                OptionOptionality.REQUIRED, COUNTRY_NAME);
+        super.registerOptionsAndArguments();
```
lucaspcram: Awesome you got this right. The design currently requires you call `super` last, which is kinda annoying. I wish it was more intuitive.

src/main/resources/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommandExamplesSection.txt
```diff
@@ -0,0 +1,6 @@
+Convert a PBF to a raw Atlas:
```
lucaspcram: 👍
I love documentation.

src/main/resources/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasDescriptionSection.txt
```diff
@@ -0,0 +1,5 @@
+Convert PBF file(s) into raw (i.e. unsliced, unsectioned) Atlas file(s). The default behavior
```
lucaspcram: 👍👍
I *really* love documentation.

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
```diff
@@ -0,0 +1,177 @@
...
+        super();
+        this.optionAndArgumentDelegate = this.getOptionAndArgumentDelegate();
+        this.outputDelegate = this.getCommandOutputDelegate();
+    }
+
+    @Override
+    public int execute()
+    {
+        // set up the output path from the parent class
+        final int code = super.execute();
```
lucaspcram: Another little wart in my design....

What I really want instead of having abstract super classes is to have some way of defining command templates. Since you can only inherit from one superclass, having the command templates (like `MultipleOutputCommand`) be abstract classes is limiting. And it leads to design warts like the one here.

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
adahn6: oh good this seemed really hacky but I didn't know about FileSuffix

lucaspcram: It's a @MikeGost classic. I evangelize it at every opportunity.

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/PbfToRawAtlasCommand.java
adahn6: I'm on board with this actually. realistically, the parent class here legitimately is going to be executing code beforehand, combined with the extra behavior implemented by the subclass-- that seems to fit the abstract class model nicely.

lucaspcram: I suppose. I just worry that down the line we may desire more flexibility - what if we want a `MultipleOutputCommand` that is also a `DoohickyCommand`. But I suppose for now this design keeps things simple.

PR #369, opened 2019-02-28
Atlas GeoJson Interfaces

jklamer:
### Description:

While mostly cosmetic changes, this PR established Interfaces and utilities to formalize and facilitate the conversion and flexibility of converting Atlas objects into GeoJSON objects.  This is an interface builder so that we can formalize our contracts within the GeoJSON Spec( referenced heavily https://tools.ietf.org/html/rfc7946). This should allow us to be much clearer with our representations and be more flexible with reads and writes. 

Things that are new:
- Multiple GeoJSON Interfaces under the class path "org.openstreetmap.atlas.geography.geojson" that formalize the parts an object must implement to be represented as GeoJson
- GeoJsonTypes enum fitting with the spec of not extensible 
- Building on the work of @hallahan , adding functions to GeoJsonUtils to allow for the easy construction of GeoJson Objects. 
- Representing Atlas as a Geojson Feature collection
- a couple missing GeoJson representable classes

Things that are changed:
- [BREAKING CHANGE] asGeojson() method on some classes that is scarcely implemented now returns JsonObject
- [BREAKING CHANGE] asGeoJsonGeometry() now should only return the geometry object, which can be the whole GeoJson, but for features this method no longer returns the properties. This is to keep in clear alignment with the spec.
- AtlasEntities no longer have the shard information in their properties - the shard information is included in the Atlas properties. Reduces the duplicated information for every object in a shard
- Relations can be represented as FeatureCollections as well as Features.

Things that are yet to be done:
- Deduplicate work between GeojsonBuilder, GeojsonUtilies, GeojsonReader, Geojson coordinate converters, and other assorted places.
- Utilize interfaces for reading and writing simplifications. (Buffered writer for Atlas toGeoJson for example)
- various other cleaning tasks 

### Potential Impact:

This will break some code relying on GeoJsonObect (Json wrapper) methods to access 

### Unit Test Approach:

Added tests for new methods and classes. Will be relying on Sonar to help illuminate coverage

### Test Results:

Passes new tests, and old tests where it is meant to. Representing a Polyline as a Feature no longer makes sense, and some of the new utilities are used to duplicate what the test is expecting the representation to be. I am open to improving these but would like guidance. 

------


src/main/java/org/openstreetmap/atlas/geography/atlas/command/AtlasGeoJSONSubCommand.java
```diff
@@ -61,7 +61,7 @@ protected int finish(final CommandMap command)
     @Override
     protected void handle(final Atlas atlas, final CommandMap command)
     {
-        getWriter(command).write(atlas.asGeoJson(x -> true).jsonObject());
+        getWriter(command).write(atlas.asGeoJson(x -> true));
```
yazad3: Suggestion - provide an overloaded asGeoJson() that that in turn calls asGeoJson(x -> true)

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
```diff
@@ -213,6 +212,18 @@ public AtlasEntity entity(final long identifier, final ItemType type)
         }
     }
 
+    @Override
+    public Iterable<AtlasEntity> getGeoJsonObjects()
+    {
+        throw new UnsupportedOperationException();
```
yazad3: Please consider returning an empty list here.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
```diff
@@ -93,13 +92,13 @@ public Area area(final long identifier)
...
+    public JsonObject asGeoJson()
     {
         throw new UnsupportedOperationException();
     }
 
     @Override
-    public GeoJsonObject asGeoJson(final Predicate<AtlasEntity> matcher)
+    public JsonObject asGeoJson(final Predicate<AtlasEntity> matcher)
     {
         throw new UnsupportedOperationException();
```
yazad3: Please consider returning an Empty equivalent of the json object.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/Relation.java
```diff
@@ -110,8 +117,7 @@ public JsonObject asGeoJsonGeometry()
             {
                 final String message = String.format("%s - %s",
                         exception.getClass().getSimpleName(), exception.getMessage());
-                properties.addProperty("exception", message);
-                logger.warn("Unable to recreate multipolygon for relation {}.", getIdentifier(),
+                logger.error("Unable to recreate multipolygon for relation {}.", getIdentifier(),
```
yazad3: Sonar has also pointed this out, but you are missing a pair of squiggly braces in the message template.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
jklamer: I am open to it, but I am following the pattern of the other methods in the EmptyAtlas.java and am not sure about breaking from that convention.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
jklamer: I am open to it, but I am following the pattern of the other methods in the EmptyAtlas.java and am not sure about breaking from that convention.

src/main/java/org/openstreetmap/atlas/geography/atlas/BareAtlas.java
```diff
@@ -85,16 +85,30 @@ protected BareAtlas()
...
+        return GeoJsonUtils.featureCollection(this);
     }
 
     @Override
-    public GeoJsonObject asGeoJson(final Predicate<AtlasEntity> matcher)
+    public JsonObject asGeoJson(final Predicate<AtlasEntity> matcher)
     {
-        return new GeoJsonBuilder().create(Iterables.filterTranslate(entities(),
-                atlasEntity -> atlasEntity.toGeoJsonBuildingBlock(), matcher));
+        return GeoJsonUtils.featureCollection(new GeoJsonFeatureCollection<AtlasEntity>()
```
matthieun: Nice!

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
matthieun: EmptyAtlas serves just as a reference to Complete features, so by default, it does not do anything.

src/test/java/org/openstreetmap/atlas/geography/geojson/GeoJsonBuilderTest.java
```diff
@@ -82,11 +81,12 @@ public void testCreateFeatureCollectionFromPropertiesLocated()
                     featureCollection.jsonObject().get("features").getAsJsonArray().get(i)
                             .getAsJsonObject().get("geometry").getAsJsonObject().get("type")
                             .getAsString());
-            Assert.assertEquals(
-                    PolyLine.TEST_POLYLINE.asGeoJson().jsonObject().get("features").getAsJsonArray()
-                            .get(0).getAsJsonObject().get("geometry"),
-                    featureCollection.jsonObject().get("features").getAsJsonArray().get(i)
-                            .getAsJsonObject().get("geometry").getAsJsonObject());
+            // Polyline asGeoJson no longer results in feature collection
```
matthieun: Can we remove the code altogether?

PR #372, opened 2019-03-03
Followup fix for the isClockwise method

snb3300:
### Description:

I had made a fix in https://github.com/osmlab/atlas/pull/365 to convert earth coordinates in cartesian coordinates to calculate the sum over edges for polygon orientation. But sum over edges fails for a lot of cases because the polygons are not on a plane but on a sphere.

Few examples which would fail the sum over edges cases:

1. POLYGON((70.5233941 -49.2953329,70.5234584 -49.2955148,70.5235872 -49.2963964,70.5230722 -49.2963964,70.5229649 -49.2962145,70.5227718 -49.2961585,70.5225143 -49.2960186,70.5224285 -49.2958507,70.5226216 -49.2955568,70.5223426 -49.2955288,70.5225358 -49.295095,70.5233941 -49.2953329))

2. POLYGON((69.5530556 -49.2504684,69.5530878 -49.2504019,69.5530985 -49.2503494,69.5531736 -49.2504089,69.5533614 -49.2504054,69.553415 -49.2504439,69.553356 -49.250584,69.5531629 -49.2505455,69.5530556 -49.2504684))

3. POLYGON((-58.7546961 -51.5537665,-58.754588 -51.5537003,-58.754411 -51.553737,-58.7543466 -51.5538204,-58.7543573 -51.5539338,-58.7544324 -51.5539838,-58.7545995 -51.5539833,-58.754706 -51.5539371,-58.7547543 -51.5538638,-58.7547597 -51.5538004,-58.7553229 -51.5538738,-58.7546961 -51.5537665))

This PR calculates the signed area of polygon on a sphere to get the orientation. If the area is negative that means orientation is clockwise else it is counter clockwise.

There are standard ways to calculate the area of polygon on a sphere. Please refer [Spherical Trigonometry](http://www.gutenberg.org/files/19770/19770-pdf.pdf?session_id=374cbf5aca81b1a742aac0879dbea5eb35f914ea) to understand the properties of a triangle on a sphere. It also derives formulae to calculate the area of triangle/polygon on a sphere.

I also referred http://mathforum.org/library/drmath/view/51879.html to understand the Trigonometry behind the calculations.

### Potential Impact:

This should not impact any downstream processes as we are not changing the geometry of polygons

### Unit Test Approach:

Added few more unit tests to make sure the new approach is passing. There were few failures near the poles so the unit tests mostly adds those polygons to existing ones.

### Test Results:

All the unit tests passed.


src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -378,31 +378,51 @@ public boolean isApproximatelyNSided(final int expectedNumberOfSides, final Angl
...
+                // You can derive this from the formula on Page 74, point 102 of the book
+                sphericalExcess += 2 * Math.atan2(alpha * Math.sin(deltaLongitude),
+                        1 + alpha * Math.cos(deltaLongitude));
             }
-            previousCartesianPoint = newCartesianPoint;
+            previousLocation = currentLocation;
         }
-        return sum >= 0;
+
+        return sphericalExcess * Distance.AVERAGE_EARTH_RADIUS.asKilometers()
```
matthieun: Is multiplying by the earth radius twice here necessary? The result will always have the same sign as `sphericalExcess` anyway.

snb3300: Yes, I agree that it is not necessary but I kept it so that anyone reading this should know that we are calculating the area of the polygon

matthieun: Yes, I think you should remove it as it does not bring any more information. To let people know that we are calculating the area of a polygon, a comment should do it!

snb3300: okay, sounds good 👍

PR #373, opened 2019-03-05
Segment intersection bug fix

jklamer:
### Description:

I ran into a bug where two segments that end at the same location were returning false from a ```Segment::intersects(Segment)``` call when positioned in a certain way but not in other ways. 

I copied their geometries and into Antartica to prove the issue was reproducible. The two segments (in WKT) are
```
LINESTRING (112.9699474 -84.7999528, 112.9699948 -84.7999669)
LINESTRING (112.9650809 -84.7999622, 112.9699948 -84.7999669)
```

**Background:** 
To figure out what was supposed to be happening inside the method I had to dive into [this section](https://dl.acm.org/citation.cfm?id=130777) of Franklin Antonio's Graphics Gems III, 1992, which describes the algorithm that is implemented within the Atlas function. 

The basic premise is that you can represent any point along the geometric line (of infinite length) that the segment is cut from using a ratio * the vector from segment start to segment end + the start vector. Such that if a point lies along the segment its ratio representation is within the range [0,1]. The algorithm uses this knowledge and solves for the point of intersection of the two infinite geometric lines represented by the segments in terms of its ratios with respect to the two segments, which we arbitrarily call α(alpha) for segment 1 and β(beta)  for segment 2. 

For reference the resulting formula for the calculation of α and β is below. Where A is the vector from the start of segment 1 to the end of segment 1, B is the vector from the end of segment 2 to the start of segment 2, and C is the vector from the start of segment 2 to the start of segment 1. 
screen shot 2019-03-04 at 4 26 14 pm

There is special case handling for parallel lines and you can optimize away having to do the division and actually solving for the ratio if all you want to know if intersection occurs or not. So we end up with local variables that are the numerator of the alpha equation, the numerator of the beta equation, and the denominator that both equations share.

**The Bug:**
When segments intersect at their end points then both ratios should end up being 1 (the inclusive end of the range). However in the calculation of our local variables there is a potential problem that arises if we do the math as doubles. For the lines above it results in the following debugging information:
image
In our situation we should see alphaNumerator==commonDenominator==betaNumerator. But due to our double representation the betaNumberator is 1e-23 off. This becomes a clearer issue when do the actual division to see the ratio these numbers represent. 
image
This ratio is outside the required [0,1].

**The Fix:**
Our underlying location information is represented in dm7 (```long```) precision, so if we use that we don't have to worry about any precision loss from one representation to the other. And because the algorithm optimizes away all division operations, there is no potential precision loss from mathematical operations using ```long```s. 

This creates the potential but rare case for ```long``` overflow. In that case we use the old double based algorithm (renamed to reflect its approximate nature) to look for intersection existence. 

Below is a mathematical proof of why both failure cases can't happen at the same time within the Atlas geographic system.

**Proof:**
Thanks to @ranek for talking with me about some of this math.

_Assumption:_ This double precision bug will only happen when the segments intersect at their start or end points.

Reasoning for this Assumption:
- By looking at the precision loss we got by the end of 3 mathematical operations (at evaluation time) on doubles with 7 decimal points was 1e-23. While that small amount is able to throw off evaluation at the ratio range limit, the smallest location change we are capable of is a 1 dm7 change. This is  1e-7 in double space which is many orders of magnitude bigger than 1e-23. The resulting ratio (while potentially not perfect) will not get pushed outside the acceptable range unless it it mathematically supposed to be at the inclusive end of the range. Basically all of the closest locations are vastly different enough when look at our margin for error that we shouldn't have to worry about overlapping error spaces. Even if it's possible, I don't believe that it matters within the context of the next part of the proof. 

_Theorem_: It is impossible for the function to overflow on its ```long```  operations when the segments start or end at the same place within the Atlas geography system. 

Looking at functions for the ratios above you see that for each numerator and the common denominator you are multiplying some x component of a vector with the y component of another. Never x component by another x component. The max x component (MaxX) in ```long``` of any vector between two locations within the Atlas dm7 system is ~3.6e9 (Anti-meridian to Anti-meridian). The max y component (MaxY) of any vector between two locations within the Atlas dm7 system is 1.8e9 (South Pole to North Pole). 

So the max result of any multiplication within the ratio equations is MaxX * MaxY. We can prove that this will not overflow by calculating the number of bits needed to represent the number with log(MaxX * MaxY)/log(2). If if you're worried about about overflow in that calc use log(MaxX)/log(2) +  log(MaxY)/log(2). This yields ceil(62.4906995211)=63 which is little enough bits for a 64 bit signed value. So all the multiplications are overflow safe.

Then there are the subtraction operations, which if you have two large values that both work in the same direction( large value needing 63 bits - (-large value needing 63 bits)), potentially creates the case for overflow. However _we only need to worry about this overflow if we can't rely on the double version of the method to accurately detect as a back-up. We can't rely on the double method only when the segments starts or ends are the same. So is it possible to create a situation where the subtraction functions overflow and the segments start or end in the same location_? If not then there are no potential precision issues for. 

Every subtraction is in the form V1x * V2y - V1y * V2x. And to get the max values out of it let's say they both expand the whole MaxX. We also want the vectors to meet after traveling as much total distance in the y direction, so they either start or end at Lat(90) or Lat(-90). To fit all these requirements these vectors must either start or end at the north or south end of one Anti-meridian and end or start meeting somewhere in the middle on the far Anti-meridian (Lon(-180) to Lon(180) or visa versa). For the sake of simplicity I will walk through the rest of the example with two vectors where one starts at the "top left" and other at the "bottom left" of the map and meet somewhere on the "right side". They meet at a height of _delta_ * MaxY. Substituting all these values into the equation you get: 

MaxX* _delta_ * MaxY - (_delta_ - 1) * MaxY * MaxX . 

screen shot 2019-03-04 at 6 30 38 pm
 
You find that the _delta_ falls away and you are left with the result of the subtraction always being MaxX * MaxY no matter where they meet on the opposite Anti-meridian, which we proved earlier to be small enough to avoid overflow. You can try different situations and find that the max value you can get for the subtraction is always MaxX * MaxY provided the segments meet at one end.  

So after creating the largest possible vectors in the Atlas geography system that meet at an end, the subtraction operations will not overflow. This means that we do not need to worry about the double precision issue when there is overflow, making the double based function a suitable backup to the long based function. 

**TLDR:** I ran into a bug that arose due to double math being approximate by nature. To solve this I use ```long```s in the function instead because there is no division. Using ```long```s creates a rare scenario where overflow happens, but in those cases I am able to fall back on the ```double``` based method because it is mathematically impossible within our Geographic/Location system to create a case that fails both. 

### Potential Impact:

Very unique situations that should have been returning true for intersections and were returning false will now return true. This has the potential to massively change the results of some algorithms, but it shouldn't break the logical flow of those algorithms, would would be getting the right result most of the time.

### Unit Test Approach:

I created a test that failed the double based function due to precision, and I created a unit test that failed the long based function due to overflow. I then made the new default endpoint be the long based function and used Math.subtractExact to catch any overflow, at which point I use the double based function. 

### Test Results:

Passes all existing tests and new ones.

------


jklamer:
@matthieun I refactored some of the redundant if blocks without compromising the flow that corresponds to the original algorithm. I don't think SonarCude will be happy till I have a couple of decently massive if statements that will be hard to read and refer back to the original. I can keep going but I want to make sure there is a line.

matthieun:
So there are some easy ones, but not all are worth it I agree. How about those 4:

- https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWlPh0BWe1QvbR9zyEVi&open=AWlPh0BWe1QvbR9zyEVi&pullRequest=373
- https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWlPh0BWe1QvbR9zyEVj&open=AWlPh0BWe1QvbR9zyEVj&pullRequest=373
- https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWlPh0BWe1QvbR9zyEVk&open=AWlPh0BWe1QvbR9zyEVk&pullRequest=373
- https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AWlPh0BWe1QvbR9zyEVl&open=AWlPh0BWe1QvbR9zyEVl&pullRequest=373

If you think even those are not worth it, then we can approve and merge.

jklamer:
@matthieun I'll fix those but I'll also add a comment to make it clear what part of the algorithm it is corresponding to.

jklamer:
ping, on if any outstanding requests

PR #378, opened 2019-03-13
Speed Validator/Extractor: Handle Implicit and None values

Bentleysb:
### Description:

This adds handling for 2 non-numeric speed value types to both the SpeedValidator and SpeedExtractor classes. These values are defined [here](https://wiki.openstreetmap.org/wiki/Key:maxspeed#Values).
The first value is `none`, and simply means there is no legal speed for the road. The validator will return that this is valid, but the extractor will return an empty optional, as there is no corresponding numeric value. 
The second value type uses descriptors, instead of numeric values (e.g. `RO:urban`), and are called implicit values. These values are more often used in the source:maxspeed tag, but are widely used in the maxspeed tag in some countries. A list of these values can be found [here](https://wiki.openstreetmap.org/wiki/Speed_limits#Country_code.2Fcategory_conversion_table). This PR includes a JSON resource that maps these values to there corresponding speed values. If an implicit value is in the key set of the JSON resource the validator will say it is valid, and the extractor will attempt to parse the corresponding speed value. 

### Potential Impact:

Adds more parse-able speed values.
Downstream libraries may need to adjust to using the SpeedExtractor to handle implicit values.
The `none` value is valid but has no numeric equivalent. Anything that uses the SpeedValidator should take this into account. This is a breaking change as the SpeedValidator can now return true while the SpeedExtractor will return an empty optional for the same value. 

### Unit Test Approach:

Unit test were added for the two types of values in the SpeedValidator's unit tests.
All new unit test were written for the SpeedExtractor that cover all cases, as none existed before.

### Test Results:

None

------


src/main/java/org/openstreetmap/atlas/tags/annotations/extraction/SpeedExtractor.java
```diff
@@ -18,33 +23,40 @@
     private static final SpeedValidator VALIDATOR = new SpeedValidator();
     private static final String SINGLE_SPACE = " ";
     private static final Logger logger = LoggerFactory.getLogger(SpeedExtractor.class);
+    private static final JsonObject IMPLICIT_SPEED_MAP = new Gson().fromJson(
+            new JsonReader(
+                    new InputStreamReader(SpeedExtractor.class.getClassLoader().getResourceAsStream(
+                            "org/openstreetmap/atlas/tags/annotations/implicit-speed-values.json"))),
```
matthieun: To avoid issues during potential future refactorings, we should not list package names like this. It would be better to do:
```
... new InputStreamReader(Tag.class.getResourceAsStream("implicit-speed-values.json")) ...
```
Here `Tag` is in the same package as the resource you are trying to fetch.

src/main/java/org/openstreetmap/atlas/tags/annotations/validation/SpeedValidator.java
```diff
@@ -15,6 +21,11 @@
 public class SpeedValidator implements TagValidator
 {
     private static final DoubleValidator DOUBLE_VALIDATOR;
+    private static final JsonObject IMPLICIT_SPEED_MAP = new Gson().fromJson(
+            new JsonReader(
+                    new InputStreamReader(SpeedValidator.class.getClassLoader().getResourceAsStream(
+                            "org/openstreetmap/atlas/tags/annotations/implicit-speed-values.json"))),
```
matthieun: Same here!

PR #379, opened 2019-03-13
Some draft ideas for FeatureChange refactoring

lucaspcram:
In the interest of supporting a more robust `FeatureChange`, one that contains a before view of the feature in question.

I am opening a draft because this is really messy, and I want to have some discussion before moving forward with the brunt of the work.

I am going to self-comment on parts of the code I think are important.

**EDIT**
~Basic tasks are complete. Still remaining: figure out how to remove the explicitlyExcluded field from `RelationBean` and `RelationMemberList`.~
I will do the above in a later PR.


src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -47,61 +47,82 @@
...
         this.validateUsefulFeatureChange();
     }
 
+    public FeatureChange(final ChangeType changeType, final AtlasEntity updatedView)
+    {
+        this(changeType, updatedView, null);
+    }
+
     @Override
     public Rectangle bounds()
```
lucaspcram: This method will need to be updated. Before/after bounding box aggregation logic previously assigned to `CompleteEntities` will need to be performed here. This means will now need to strictly enforce `Located`ness for `CompleteEntities`. More on this below.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -246,131 +278,242 @@ public String toJson()
...
+     * CompleteEntity, and will contain only those fields which have been updated in the afterView.
+     * E.g. Suppose the afterView is a CompleteNode with ID 1, and the only updated field is the tag
+     * Map. Then the beforeView will be a CompleteNode with ID 1, where the only populated field is
+     * the tag Map. However, the afterView's tag Map will be fetched from the Atlas context. Having
+     * a beforeView for each {@link FeatureChange} allows for more robust merging strategies.
+     *
+     * @param atlas
+     *            the atlas context
+     */
+    private void computeBeforeViewUsingAtlasContext(final Atlas atlas) // NOSONAR
```
lucaspcram: Here is how we compute the beforeView of a `FeatureChange` based on a given atlas context. This beforeView will be a `CompleteEntity`, and only contain the fields that have changed.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -246,131 +278,242 @@ public String toJson()
...
+                if (updatedPointView.getLocation() != null)
+                {
+                    ((CompletePoint) beforeViewUpdatesOnly)
+                            .withLocation(beforePointView.getLocation());
+                }
+                break;
+            /*
+             * Relation specific updates. The only Relation-specific field is the member list.
+             */
+            /*
```
lucaspcram: What to do about this allKnownOsmMembers case? I guess I just don't understand this aspect of relations well at all. (What is it, and why do we need it?)

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -379,88 +522,145 @@ else if (other.bounds() != null)
...
+                                            .collect(Collectors.toCollection(TreeSet::new)),
+                            FeatureChangeMergers.directReferenceMergerSorted,
+                            FeatureChangeMergers.diffBasedIntegerSortedSetMerger);
+            final CompleteNode result = new CompleteNode(getIdentifier(),
+                    mergedLocationBean.getMergedAfterMember(), mergedTags,
+                    mergedInEdgeIdentifiersBean.getMergedAfterMember(),
+                    mergedOutEdgeIdentifiersBean.getMergedAfterMember(), mergedParentRelations);
+            if (result.bounds() == null)
+            {
+                throw new CoreException("TODO CompleteNode result bounds null, investigate");
```
lucaspcram: What to do here? Before, we had:

```
if (bounds() != null)
{		
    result = result.withAggregateBoundsExtendedUsing(bounds());		
}		
else if (other.bounds() != null)		
{		
    result = result.withAggregateBoundsExtendedUsing(other.bounds());		
}
```

Now, it seems we do not need to handle this.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergers.java
```diff
@@ -0,0 +1,321 @@
...
+
+    static final BinaryOperator<SortedSet<Long>> directReferenceMergerLooseSorted = (left,
+            right) -> Sets.withSortedSets(false, left, right);
+
+    static final BinaryOperator<Set<Long>> directReferenceMergerLoose = (left, right) -> Sets
+            .withSets(false, left, right);
+
+    static final BinaryOperator<RelationBean> relationBeanMerger = RelationBean::merge;
+
+    static final TernaryOperator<SortedSet<Long>> diffBasedIntegerSortedSetMerger = (before,
```
lucaspcram: An example of how a diff based merger can work. This one merges the in/out edge identifier sets. This logic may be generalizable.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergers.java
```diff
@@ -0,0 +1,321 @@
...
+     * @param afterEntityLeft
+     * @param beforeEntityRight
+     * @param afterEntityRight
+     * @param memberExtractor
+     * @param simpleMergeStrategy
+     * @param diffBasedMergeStrategy
+     * @return a {@link MergedMemberBean} containing the merged beforeMember view and the merged
+     *         afterMember view.
+     */
+    static <M> MergedMemberBean<M> mergeMember(final String memberName,
```
lucaspcram: The newly proposed member merging function.

lucaspcram:
Also, the diffs here are much easier to read if using the split view instead of the unified view.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
lucaspcram: Right now I am throwing an exception to see if this case ever arises as I write more test cases.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
```diff
@@ -193,23 +192,7 @@ public Edge edge(final long identifier)
...
-            case LINE:
-                return new CompleteLine(identifier);
-            case POINT:
-                return new CompletePoint(identifier);
-            case RELATION:
-                return new CompleteRelation(identifier);
-            default:
-                throw new CoreException("Unknown ItemType {}", type);
-        }
+        throw new UnsupportedOperationException();
```
matthieun: Whyyy?

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryOperator.java
```diff
@@ -0,0 +1,14 @@
...
+ * Represents an operation upon three operands of the same type, producing a result of the same type
+ * as the operands. This is a specialization of {@link TernaryFunction} for the case where the
+ * operands and the result are all of the same type.
+ *
+ * @author lcram
+ */
+@FunctionalInterface
+public interface TernaryOperator<T> extends TernaryFunction<T, T, T, T>
+{
+
```
matthieun: whitespace

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
lucaspcram: Ok. I have modified that constructor now, so it is no longer possible for a `CompleteEntity` to ever be unbounded.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergers.java
lucaspcram: I am leaving them because we fall back on them in cases where no beforeView is provided.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -246,131 +283,241 @@ public String toJson()
...
         }
-        return FeatureChange.add(result);
+        return this;
     }
 
-    private <M> M mergedMember(final String memberName, final AtlasEntity left,
-            final AtlasEntity right, final Function<AtlasEntity, M> memberExtractor,
-            final Optional<BinaryOperator<M>> memberMergerOption)
+    /**
+     * Compute the beforeView using a given afterView and Atlas context. The beforeView is a
```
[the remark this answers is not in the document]
lucaspcram: Yeah this is a good idea.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -246,131 +283,241 @@ public String toJson()
...
+     * CompleteEntity, and will contain only those fields which have been updated in the afterView.
+     * E.g. Suppose the afterView is a CompleteNode with ID 1, and the only updated field is the tag
+     * Map. Then the beforeView will be a CompleteNode with ID 1, where the only populated field is
+     * the tag Map. However, the afterView's tag Map will be fetched from the Atlas context. Having
+     * a beforeView for each {@link FeatureChange} allows for more robust merging strategies.
+     *
+     * @param atlas
+     *            the atlas context
+     */
+    private void computeBeforeViewUsingAtlasContext(final Atlas atlas) // NOSONAR
```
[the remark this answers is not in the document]
lucaspcram: I might refactor it - but I just don't think it's cognitively complex. It's basically a big switch statement.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -246,131 +283,241 @@ public String toJson()
...
+             * Area specific updates. The only Area-specific field is the polygon.
+             */
+            case AREA:
+                final Area updatedAreaView = (Area) this.updatedView;
+                final Area beforeAreaView = (Area) beforeViewFromAtlas;
+                beforeViewUpdatesOnly = CompleteArea.shallowFrom(beforeAreaView);
+                if (updatedAreaView.asPolygon() != null)
                 {
-                    result = memberMergerOption.get().apply(leftMember, rightMember);
+                    ((CompleteArea) beforeViewUpdatesOnly).withPolygon(beforeAreaView.asPolygon());
```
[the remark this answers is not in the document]
lucaspcram: Yes, because `beforeViewUpdatesOnly` is of type `AtlasEntity`. So I need to safely downcast in the switch statement to access the right specialized subclass methods (`withPolygon` in this case).

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergers.java
```diff
@@ -0,0 +1,328 @@
...
+        final Set<Long> addedToLeftView = com.google.common.collect.Sets.difference(afterLeft,
+                before);
+        final Set<Long> addedToRightView = com.google.common.collect.Sets.difference(afterRight,
+                before);
+
+        final Set<Long> removedMerged = Sets.withSets(false, removedFromLeftView,
+                removedFromRightView);
+        final Set<Long> addedMerged = Sets.withSets(false, addedToLeftView, addedToRightView);
+
+        final Set<Long> collision = com.google.common.collect.Sets.intersection(removedMerged,
```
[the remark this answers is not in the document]
lucaspcram: Once I write all of these I am going to look for common themes to generalize.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
lucaspcram: We could put it back, but then we need to arbitrarily locate the returned `CompleteEntities`. Maybe null island? This is part of that idea that we should never allow a `Located` to return `null` from a `bounds` call. I don't think it is reasonable for clients and users of `Located` objects to be forced to `null` check the result of `bounds` at every turn. I am open to having my mind changed on this. But as it stands, enforcing this will make the FeatureChange merging code easier to write I think.

PR #380, opened 2019-03-15
Allowed additional imports for predicates

lucaspcram:
### Description:

Added some more imports for predicates. You can now use `Tag` classes and `Validator` classes.

Example:
```
$ atlas subatlas --predicate 'Validators.from(FootwayTag.class, e).isPresent()' myatlas.atlas
```
@jklamer if there are any more packages you think may be useful, please call them out so we can include them in this PR.

### Potential Impact:

N/A

### Unit Test Approach:

N/A

### Test Results:

N/A

------


lucaspcram:
@jklamer good idea. Added.

PR #381, opened 2019-03-18
Improved dynamic slicing performance

adahn6:
### Description:

This PR makes a two smaller changes to improve performance around dynamic relation slicing. First, the filter for DynamicAtlas policy now exclusively uses Relations to determine whether to expand to a shard. The previous policy mistakenly checked for any points, which triggered false-positive expansions to unnecessary shards. Second, the RawAtlasCountrySlicer will now sub atlas out all relations from the DynamicAtlas that weren't a part of the original shard before slicing relations. This prevents the unnecessary computation of slicing relations picked up by the expanded DyanmicAtlas, as these relation will all have been cut out of the final fully sliced atlas anyway. Note that this works because the relations are removed with the SubAtlas SILK_CUT policy, which will preserve any relations that are members of relations that will be kept-- this means if the DynamicAtlas loads a relation that wasn't part of the original shard but *is* a member of a relation that was a part of the original shard, it will be kept. 

Additionally, a small fix was included for certain corner cases of SilkCut.

### Potential Impact:

None, as this should not change the final output Atlas, only improve the performance of the slicing operation!

### Unit Test Approach:

None, as this doesn't change output data at all, only performance.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
@@ -19,7 +19,7 @@ import org.openstreetmap.atlas.geography.atlas.dynamic.DynamicAtlas;
 import org.openstreetmap.atlas.geography.atlas.dynamic.policy.DynamicAtlasPolicy;
 import org.openstreetmap.atlas.geography.atlas.items.AtlasEntity;
 import org.openstreetmap.atlas.geography.atlas.items.ItemType;
-import org.openstreetmap.atlas.geography.atlas.items.Point;
+import org.openstreetmap.atlas.geography.atlas.sub.AtlasCutType;
 import org.openstreetmap.atlas.geography.boundary.CountryBoundaryMap;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasCountrySlicer.java
```diff
@@ -265,14 +260,50 @@ public Atlas sliceRelations(final Shard initialShard)
...
+            final RawAtlasSlicer relationSlicer = new RawAtlasRelationSlicer(
+                    partiallySlicedExpandedAtlasWithExtraRelationsRemoved, initialShard,
+                    this.countries, this.countryBoundaryMap);
+            logger.info(COMPLETED_TASK_MESSAGE, RELATION_SLICING_TASK, shardName,
+                    time.elapsedSince());
+            return relationSlicer.slice();
+        }
+        else
+        {
+            throw new CoreException("Error while preparing subatlas for relation slicing shard {}!",
```
matthieun: Here I would have a message that says that the sub-ed partiallySlicedExpandedAtlas returned nothing...

PR #382, opened 2019-03-18
Retrieve DAG process groups

matthieun:
### Description:

Return a function that gets all the groups within the same priority level in a DAG. As tested in this class:
```
                  Copper
                /
        Coinage
      /         \
Metal              Gold
      \         /
        Precious
                \
                   Platinum
```

Returns:
```
1. Metal
2. Coinage, Precious
3. Copper, Gold, Platinum
```

### Potential Impact:

None, new method

### Unit Test Approach:

Added one extra unit test

### Test Results:

Passing

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/graphs/DirectedAcyclicGraph.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/graphs/DirectedAcyclicGraph.java
@@ -150,19 +150,19 @@ public class DirectedAcyclicGraph<V> implements Serializable
         final Set<V> sourcesNotAdded = new HashSet<>(getSources());
         while (!sourcesNotAdded.isEmpty())
         {
-            final Set<V> allCandidates = new HashSet<>();
-            final Set<V> candidates = new HashSet<>();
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/utilities/graphs/DirectedAcyclicGraph.java
```diff
@@ -136,6 +138,50 @@ public boolean isSource(final V vertex)
...
+    public List<Set<V>> processGroups()
+    {
+        final Stack<Set<V>> stack = new Stack<>();
+        stack.push(new HashSet<>(getSinks()));
+        final Set<V> added = new HashSet<>(getSinks());
+        final Set<V> sourcesNotAdded = new HashSet<>(getSources());
+        while (!sourcesNotAdded.isEmpty())
+        {
+            final Set<V> allCandidates = new HashSet<>();
+            final Set<V> candidates = new HashSet<>();
```
lucaspcram: This is a minor nit: but it may be helpful for future readers if these two variables had more distinguishing names. `allCandidates` and `candidates` may not be immediately clear.

PR #389, opened 2019-03-25
A few slicing changes

adahn6:
### Description:

This PR makes a number of changes around country slicing for lines. First, an issue has been fixed with respect to closed Line slicing. Previously, all closed Lines were being cast to Polygon geometry, then sliced. While this makes sense initially, as the closed line is a valid Polygon geometry, there is an Edge case (ha!) that must be consider-- some closed Lines will later become Edges during way-sectioning. If these Lines (future Edges) are sliced as Polygons, then the slicer will attempt to preserve closed geometry by using the lines in the boundary map. For example, in the unit test LineAndPointSlicingTest.testClosedLineSpanningTwoCountries(), the following closed Line spans two countries:

closedLineSpanningTwoCountriesUnitTestRaw

After slicing, it's bisected by the country boundary line as two closed shapes:

closedLineSpanningTwoCountriesUnitTestSliced1
closedLineSpanningTwoCountriesUnitTestSliced2

For many closed Lines, this makes sense-- this is treating them like a future Area object. However, for Lines that are future Edges, this is problematic; Edges are intended for navigable Lines, and this logic will unintentionally create a navigable Edge along the boundary. In order to fix this, all Lines are now checked against the Line to Edge configuration filter; should a closed Line qualify as an Edge, it is simply sliced as a Line. The result can be seen in the unit test LineAndPointSlicingTest.testClosedEdgeSpanningTwoCountries(), which uses the same test data as above but with tagging that ensures the Line qualifies as an Edge. The resulting sliced Line looks like this:

closedEdgeSpanningTwoCountriesUnitTestSliced1
closedEdgeSpanningTwoCountriesUnitTestSliced2

The second major fix included in this PR changes the output of Line slicing to be more consistent. Previously, all returned slices from the JTS geometry calls were treated as independent valid Line slices. However, this approach is naive; there are cases in which slices are both within the same country and also intersect. Given that the original geometry is a single line, it makes sense to attempt to preserve this by condensing any intersecting Line slices. For example, in the unit test above for the closed Edge spanning two countries, the original slicing behavior returned the following output:

closedEdgeSpanningTwoCountriesUnitTestSlicedUnmerged1
closedEdgeSpanningTwoCountriesUnitTestSlicedUnmerged2
closedEdgeSpanningTwoCountriesUnitTestSlicedUnmerged3

This output isn't ideal-- the two Line slices for the right hand side are really just an artifact of the way the Line is divided during the slicing operation, not due to any necessity. The fix for this simply uses JTS intersect calls on all slices return for the same country; any intersecting geometry will be merged, and any non-intersecting geometry will be left as-is. 

While this could be a concerning change given that it will pass all sliced geometry through this merge logic, testing shows almost entirely unchanged behavior, other than the above listed unit test. Only two other test data changed, the first of which was the following Line from RawAtlasIntegrationTest. Previously, slicing on Line 313508037 resulted in two sliced Lines with intersecting coordinates:

`[Line: id=313508037004000, polyLine=LINESTRING (-8.204767 7.549113, -8.204767 7.549113), [Tags: [last_edit_user_name => Daniel Specht], [last_edit_changeset => 26878374], [last_edit_time => 1416361534000], [last_edit_user_id => 2393795], [iso_country_code => GIN], [highway => unclassified], [last_edit_version => 1]]]`
And
`[Line: id=313508037003000, polyLine=LINESTRING (-8.2051141 7.5487909, -8.204767 7.549113), [Tags: [last_edit_user_name => Daniel Specht], [last_edit_changeset => 26878374], [last_edit_time => 1416361534000], [last_edit_user_id => 2393795], [iso_country_code => GIN], [highway => unclassified], [last_edit_version => 1]]]`

After the merge code, the result is the much more reasonable single joined line 
`[Line: id=313508037003000, polyLine=LINESTRING (-8.2051141 7.5487909, -8.204767 7.549113), [Tags: [last_edit_user_name => Daniel Specht], [last_edit_changeset => 26878374], [last_edit_time => 1416361534000], [last_edit_user_id => 2393795], [iso_country_code => GIN], [highway => unclassified], [last_edit_version => 1]]]`

The second test change from this was for InvalidMultipolygonSlicingTest.testSelfIntersectingOuterRelationAcrossBoundary(). This change is because while the Line being sliced is closed, it's an invalid Polygon geometry due to being a self-intersecting Line. When a closed Line can't be cast to Polygon because of this, it is sliced like a standard Polyline; this is the same behavior that is now default for Edge slicing. The merge code will now close the slices from the beginning and end of the Line, resulting in a cleaner look. Here's the before:

invalidPolygonSlice1Before
invalidPolygonSlice2Before

And here's after, joined since they intersect:

invalidPolygonSlice1After

Lastly, the number of allowable Points per Line has been upped to 100000, and the PointAndLineSlicer code has been somewhat refactored to increase readability. 

### Potential Impact:
Hopefully none, as data should be largely the same. 

### Unit Test Approach:

Unit tests detailed above.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
@@ -26,7 +26,6 @@ import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.items.Line;
 import org.openstreetmap.atlas.geography.atlas.items.Point;
 import org.openstreetmap.atlas.geography.atlas.pbf.AtlasLoadingOption;
-import org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier.AbstractIdentifierFactory;
 import org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier.CountrySlicingIdentifierFactory;
 import org.openstreetmap.atlas.geography.atlas.pbf.slicing.identifier.PointIdentifierFactory;
 import org.openstreetmap.atlas.geography.atlas.raw.slicing.changeset.ChangeSetHandler;
@@ -54,9 +53,6 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
 {
     private static final Logger logger = LoggerFactory.getLogger(RawAtlasPointAndLineSlicer.class);
 
-    // The raw Atlas to slice
-    private final Atlas rawAtlas;
-
     // Keeps track of all changes made during slicing
     private final SimpleChangeSet slicedPointAndLineChanges;
 
@@ -65,8 +61,7 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
 
     public RawAtlasPointAndLineSlicer(final Atlas atlas, final AtlasLoadingOption loadingOption)
     {
-        super(loadingOption, new CoordinateToNewPointMapping());
-        this.rawAtlas = atlas;
+        super(loadingOption, new CoordinateToNewPointMapping(), atlas);
         this.slicedPointAndLineChanges = new SimpleChangeSet();
     }
 
@@ -79,18 +74,18 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
     public Atlas slice()
     {
         final Time time = Time.now();
-        logger.info("Starting Point and Line Slicing for {}", getShardOrAtlasName());
+        logger.info("Starting Point and Line Slicing for Atlas {}", getShardOrAtlasName());
 
         // Slice lines and points
         sliceLines();
         slicePoints();
 
         // Apply changes and rebuild the atlas with the changes before slicing relations
-        final ChangeSetHandler simpleChangeBuilder = new SimpleChangeSetHandler(this.rawAtlas,
+        final ChangeSetHandler simpleChangeBuilder = new SimpleChangeSetHandler(getStartingAtlas(),
                 this.slicedPointAndLineChanges);
         final Atlas atlasWithSlicedWaysAndPoints = simpleChangeBuilder.applyChanges();
 
-        logger.info("Finished Point and Line Slicing for {} in {}", getShardOrAtlasName(),
+        logger.info("Finished Point and Line Slicing for Atlas {} in {}", getShardOrAtlasName(),
                 time.elapsedSince());
         getStatistics().summary();
         return atlasWithSlicedWaysAndPoints;
@@ -164,7 +159,8 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
 
         if (result == null || result.isEmpty())
         {
-            logger.error("Invalid Geometry for line {}", lineIdentifier);
+            logger.error("Invalid Geometry for line {} for Atlas {}", lineIdentifier,
+                    getShardOrAtlasName());
         }
 
         return result;
@@ -176,19 +172,14 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
         while (pointIdentifierFactory.hasMore())
         {
             final long identifier = pointIdentifierFactory.nextIdentifier();
-            if (this.rawAtlas.point(identifier) == null)
+            if (getStartingAtlas().point(identifier) == null)
             {
                 return identifier;
             }
         }
         throw new CoreException(
-                "Slicing line {} exceeded maximum number {} of supported new points",
-                AbstractIdentifierFactory.IDENTIFIER_SCALE_DEFAULT, lineIdentifier);
-    }
-
-    private String getShardOrAtlasName()
-    {
-        return this.rawAtlas.metaData().getShardName().orElse(this.rawAtlas.getName());
+                "Slicing Line {} exceeded maximum number {} of supported new Points for Atlas {}",
+                pointIdentifierFactory.getIdentifierScale(), lineIdentifier, getShardOrAtlasName());
     }
 
     private boolean isOutsideWorkingBound(final Map<String, String> tags)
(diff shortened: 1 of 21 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -164,6 +170,22 @@ private void addRawAtlasPointsToLine(final Location location,
...
+    {
+        while (pointIdentifierFactory.hasMore())
+        {
+            final long identifier = pointIdentifierFactory.nextIdentifier();
+            if (this.rawAtlas.point(identifier) == null)
+            {
+                return identifier;
+            }
+        }
+        throw new CoreException(
```
matthieun: Is there a way to print the Atlas' name here too?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -205,195 +263,132 @@ private boolean isOutsideWorkingBound(final Map<String, String> tags)
...
-            this.slicedPointAndLineChanges.updateLineTags(line.getIdentifier(), tags);
-            updateLineShapePoints(line);
-        }
-        else if (slicesBelongToSingleCountry(slices)
-                && !getCountryBoundaryMap().shouldForceSlicing(line))
+        // Used to generate identifiers for new points and lines
+        final CountrySlicingIdentifierFactory lineIdentifierFactory = new CountrySlicingIdentifierFactory(
+                line.getIdentifier());
+        final PointIdentifierFactory pointIdentifierFactory = new PointIdentifierFactory(
+                line.getIdentifier());
```
matthieun: Is it expected they both start from the line ID?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -205,195 +263,132 @@ private boolean isOutsideWorkingBound(final Map<String, String> tags)
...
-            // This line belongs to a single country, check to make sure it's the right one
-            if (isOutsideWorkingBound(slices.get(0)))
+            if (slice instanceof org.locationtech.jts.geom.Polygon)
             {
-                this.slicedPointAndLineChanges.createDeletedToCreatedMapping(line.getIdentifier(),
-                        Collections.emptySet());
+                final org.locationtech.jts.geom.Polygon polygonForSlice = (org.locationtech.jts.geom.Polygon) slice;
+                if (polygonForSlice.getNumInteriorRing() > 0)
+                {
+                    logger.error("Warning: Line {} had a multipolygon result for slicing!",
```
matthieun: Same, would be nice to have the Atlas' name here.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -205,195 +263,132 @@ private boolean isOutsideWorkingBound(final Map<String, String> tags)
...
-                this.slicedPointAndLineChanges.updateLineTags(line.getIdentifier(),
-                        createLineTags(slices.get(0), line.getTags()));
-                updateLineShapePoints(line);
+                // Check if the slice is within the working bound and mark all points for this
+                // slice for removal if so
+                if (isOutsideWorkingBound(slice))
+                {
+                    // increment the identifier factory, but don't bother slicing the line
+                    // this guarantees deterministic id assignment regardless of which countries
+                    // are being sliced
```
matthieun: 🚀

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/InvalidMultipolygonSlicingTest.java
```diff
@@ -224,14 +217,15 @@ public void testSelfIntersectingOuterRelationAcrossBoundary()
...
+        final Atlas slicedAtlas = rawAtlasSlicer.slice(rawAtlas);
+        System.out.println(slicedAtlas);
 
         // Assert that we CAN build a valid building with this relation
         new ComplexBuildingFinder().find(slicedAtlas)
                 .forEach(building -> Assert.assertFalse(building.getError().isPresent()));
 
         Assert.assertEquals(29, slicedAtlas.numberOfPoints());
-        Assert.assertEquals(8, slicedAtlas.numberOfLines());
+        Assert.assertEquals(7, slicedAtlas.numberOfLines());
```
matthieun: Is that one of the Edge (case) lines?

adahn6: ahh, whoops, should have mentioned this; I'll update the PR with a description of what this unit test looks like with the new behavior

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
adahn6: yup, I'll update for that

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
adahn6: I made this change to simplify some logic. previously, we used the slice's ID, and limited to 999 points generated. Now, we use the original Line ID, but expand to an allowable 999,999 points. while this is technically smaller in overall point generation space, it's actually functionally larger as the previous logic would fail out if any slice ran past 999, which sometimes happens. Now, one large slice can run along a country border and use several thousand points without issue.

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/InvalidMultipolygonSlicingTest.java
adahn6: one thing that's important to note here is that this *does* join the beginning and the end of the Line. I think this makes sense, because we're already splitting the Line into pieces, but *if* it's important to preserve that, we can easily disable the merge code and/or disallow merge for start and end pieces.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -134,7 +136,11 @@ private void addRawAtlasPointsToLine(final Location location,
         // Create the JTS Geometry from Line
         Geometry geometry;
 
-        if (line.isClosed())
+        if (isAtlasEdge(line))
```
lucaspcram: Good catch here.

PR #393, opened 2019-03-26
FeatureChange now supports a beforeView for more robust merging

lucaspcram:
### Description:

`FeatureChange` now supports a beforeView computed from an atlas context. The beforeView allows for more complex `FeatureChange#merge` logic.

Here is an example of how the new logic would apply to a tag merge. Imagine we have the following parameters:

beforeView: [water=lake, gazebo=yes]
afterViewLeft: [water=**pond**, gazebo=yes]
afterViewRight: [water=lake, ~gazebo=yes~]

Under the old system, which ignored the beforeView, we get:

**result: FAIL**
This would generate a key collision on the *water=pond* vs *water=lake* afterViews.

Under the new system, taking the beforeView into account, we get:

**result: [water=pond]**
The merging code sees that *afterViewLeft* modified the *water* tag, and didn't touch the *gazebo* tag. It sees that *afterViewRight* removed the *gazebo* tag, and didn't touch the *water* tag. Since there are no direct diff conflicts here, we can go ahead and merge these changes together.

Additionally, these changes will allow for a richer diff serialization.

### Potential Impact:

Downstream code that uses `FeatureChange` may see some API breakage. Additionally, new merge logic may lead to unexpected results.

### Unit Test Approach:

Added tons of additional unit tests, including two new unit test classes. See diff for details.

### Test Results:

All tests passing.

------


src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java
```diff
@@ -0,0 +1,55 @@
...
+     *
+     * @param s
+     *            the first function argument
+     * @param t
+     *            the second function argument
+     * @param u
+     *            the third function argument
+     * @return the function result
+     */
+    R apply(S s, T t, U u);
```
yazad3: Suggestion: Consider adding throws in the signature for apply so that implementers don't have to handle exceptions explicitly. This is the unfortunate downside of Java lambdas, where none of the abstract methods in the functional package `throws` anything which forces you to write `try`...`catch` re-throw - this foils the developers ability to use the lambda shortcuts like `x -> whatever()` or `this::whatever`

src/test/java/org/openstreetmap/atlas/geography/atlas/change/MemberMergeStrategiesTest.java
```diff
@@ -0,0 +1,333 @@
...
+         * while the other afterView removes it entirely.
+         */
+        this.expectedException.expect(CoreException.class);
+        this.expectedException.expectMessage(
+                "diffBasedRelationBeanMerger failed due to ADD/REMOVE conflict(s) on key(s): [LINE, 1, lineRole1]");
+        MemberMergeStrategies.diffBasedRelationBeanMerger.apply(beforeBean, afterBean1, afterBean2);
+    }
+
+    @Test
+    public void testDiffBasedRelationBeanMergeREMOVEREMOVEConflict()
```
yazad3: Is REMOVEREMOVE intentional?

lucaspcram: Ah yes. It's a `REMOVE/REMOVE` conflict, but you can't put a `/` in the name.

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java
lucaspcram: What should it throw? Just `Exception`?

yazad3: Yes, that is what I would suggest.

lucaspcram: If I add the `throws Exception`, is there a way to prevent compilation errors in cases like:

```java
static final TernaryOperator<Long> diffBasedLongMerger = 
(before, afterLeft, afterRight) ->
    {
     .
     .
     .
    }
diffBasedLongMerger.apply(before, after1, after2));
```

I get a compiler error on the `diffBasedLongMerger.apply(before, after1, after2));` since it does not catch the exception.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/MemberMergeStrategies.java
```diff
@@ -0,0 +1,453 @@
...
+import org.openstreetmap.atlas.geography.Polygon;
+import org.openstreetmap.atlas.geography.atlas.builder.RelationBean;
+import org.openstreetmap.atlas.geography.atlas.builder.RelationBean.RelationBeanItem;
+import org.openstreetmap.atlas.utilities.collections.Maps;
+import org.openstreetmap.atlas.utilities.collections.Sets;
+import org.openstreetmap.atlas.utilities.function.TernaryOperator;
+
+/**
+ * A utility class to store the various merge strategies utilized by the {@link FeatureChange} merge
+ * code. These are low-level strategies used to merge the underlying data structures.
```
matthieun: This is a really nice example of code re-use.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/EmptyAtlas.java
```diff
@@ -193,6 +193,10 @@ public Edge edge(final long identifier)
     @Override
     public AtlasEntity entity(final long identifier, final ItemType type)
     {
+        /*
+         * Note that the AtlasEntities returned by this method will technically break the Located
```
matthieun: Would that be worth adding to the method's javadoc?

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java
```diff
@@ -0,0 +1,55 @@
...
+ *            the type of the first argument to the function
+ * @param <T>
+ *            the type of the second argument to the function
+ * @param <U>
+ *            the type of the third argument to the function
+ * @param <R>
+ *            the type of the result of the function
+ * @author lcram
+ */
+@FunctionalInterface
```
matthieun: This is really nice

config/checkstyle/suppressions.xml
```diff
@@ -10,6 +10,8 @@
     <suppress checks="MultipleStringLiterals" files="src/integrationTest/*" />
     <suppress checks="InnerAssignment" files="src/test/*" />
     <suppress checks="InnerAssignment" files="src/integrationTest/*" />
-    <suppress checks="AvoidEscapedUnicodeCharacters" files="src/main/java/org/openstreetmap/atlas/utilities/command/terminal/TTYAttribute.java"/>
+    <suppress checks="AvoidEscapedUnicodeCharacters" files="src/main/java/org/openstreetmap/atlas/utilities/command/terminal/TTYAttribute.java" />
+    <suppress checks="ParameterName" files="src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java" />
+    <suppress checks="ParameterName" files="src/main/java/org/openstreetmap/atlas/utilities/function/TernaryOperator.java" />
```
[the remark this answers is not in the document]
lucaspcram: It's complaining about this line in `TernaryFunction`:
```
R apply(S s, T t, U u);
```

It says that `s`, `t`, and `u` parameter names do not match the required regex `^([a-z][a-zA-Z0-9]{2,}$|[e])$`.

Normally I think this is a reasonable warning. But in this case I don't see a reason to mangle the parameters names to appease the style checker. So I opted to add an ignore rule instead.

Also, note that I was following the precedent set in `java.util.Function`, which declares its `apply` method like:

```
R apply(T t);
```

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -55,6 +55,7 @@
             Location.forString("37.39018121506223,-122.03110367059708"),
             Location.forString("37.39021104996917,-122.0310714840889"),
             Location.forString("37.390234491673446,-122.03111171722412"));
+    public static final Polygon CENTER = new Polygon(Location.forString("0,0"));
```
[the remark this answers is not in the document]
lucaspcram: Good call. I would note that sonar complains about this and the other example `Polygon` declarations, since they expose a static mutable object. Since we only use them for testing purposes I am tempted to `// NOSONAR` all the warnings. Perhaps instead we can ignore them in the sonar UI?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -51,102 +38,215 @@
...
-    private static final BinaryOperator<RelationBean> relationBeanMerger = RelationBean::merge;
 
     private final ChangeType changeType;
-    private final AtlasEntity reference;
+    private AtlasEntity beforeView;
+    private final AtlasEntity afterView;
+
+    /**
+     * Create a new {@link ChangeType#ADD} {@link FeatureChange} with a given afterView. The
+     * afterView should be a {@link CompleteEntity} that specifies how the newly added or modified
```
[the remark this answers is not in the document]
lucaspcram: It could, technically. However, we disallow this in the constructor of `FeatureChange`.

```java
if (!(reference instanceof CompleteEntity))
{
    throw new CoreException(
        "FeatureChange requires CompleteEntity, found reference of type {}",
         reference.getClass().getName());
}
```

In fact, this was already the case before my updates - I simply notated this fact in the doc comment. I am in favor of enforcing `CompleteEntity`, since using a different entity type will almost always be due to a user error/misunderstanding/misuse.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
```diff
@@ -0,0 +1,735 @@
...
+        final CompleteRelation mergedAfterRelation = new CompleteRelation(left.getIdentifier(),
+                mergedTagsBean.getMergedAfterMember(), mergedBounds,
+                mergedMembersBean.getMergedAfterMember(),
+                mergedAllRelationsWithSameOsmIdentifierBean.getMergedAfterMember() != null
+                        ? mergedAllRelationsWithSameOsmIdentifierBean.getMergedAfterMember()
+                                .stream().collect(Collectors.toList())
+                        : null,
+                mergedAllKnownMembersBean.getMergedAfterMember(),
+                mergedOsmRelationIdentifier.getMergedAfterMember(),
+                mergedParentRelationsBean.getMergedAfterMember());
```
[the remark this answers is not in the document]
lucaspcram: Yeah the `Relation` case really got out of hand. There are so many changeable fields.

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java
yazad3: Just wrote a quick test case to demo the implications on usability and readability of FunctionalInterfaces that chose to throw CheckedExceptions as a part of their contract or otherwise.

Compare the easierToRead test cases with "messy" test cases.

```java
import org.junit.Assert;
import org.junit.Test;

import java.io.IOException;

/**
 * @author Yazad Khambata
 */
public class ExceptionalFunctionalTest {

    public static final String STR = "Hello";

    @Test
    public void testEasierToRead1() {
        final String s = SomeClass.higerOrderFunction1(this::veryComplexWork, STR);
        verify(s);
    }

    @Test
    public void testEasierToRead2() {
        final String s = SomeClass.higerOrderFunction1(str -> this.veryComplexWork(str), STR);
        verify(s);
    }

    @Test
    public void testEasierToRead3() {
        final ExceptionalFunctional<String, String> ef = this::veryComplexWork;
        final String s = ef.applyQuietly(STR);
        verify(s);
    }

    @Test
    public void testMessy1() {
        final String s = SomeClass.higerOrderFunction2(x -> {
            try {
                return this.veryComplexWork(x);
            } catch (Exception e) {
                throw new RuntimeException(e);
            }
        }, STR);
        verify(s);
    }

    @Test
    public void testMessy2() {
        ConventionalFunctional<String, String> ef = x -> {
            try {
                return this.veryComplexWork(x);
            } catch (final Exception e) {
                throw new RuntimeException(e);
            }
        };
        final String s = ef.apply(STR);
        verify(s);
    }

    private String veryComplexWork(final String str) throws Exception {
        if (false) {
            throw new IOException("ha!");
        }

        return str.toUpperCase();
    }

    private void verify(final String s) {
        Assert.assertEquals("HELLO", s);
    }

    static class SomeClass {
        public static <X, Y>  X higerOrderFunction1(final ExceptionalFunctional<X, Y> exceptionalFunctional, final Y y) {
            return exceptionalFunctional.applyQuietly(y);
        }

        public static <X, Y>  X higerOrderFunction2(final ConventionalFunctional<X, Y> exceptionalFunctional, final Y y) {
            return exceptionalFunctional.apply(y);
        }
    }

    @FunctionalInterface
    interface ExceptionalFunctional<X, Y> {
        X apply(Y y) throws Exception;

        default X applyQuietly(final Y y) {
            try {
                return this.apply(y);
            } catch (final Exception e) {
                throw new IllegalStateException(e);
            }
        }
    }

    @FunctionalInterface
    interface ConventionalFunctional<X, Y> {
        X apply(Y y);
    }
}
```

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
```diff
@@ -0,0 +1,735 @@
...
+     *            a function that can extract the member from its entity (e.g. for the polyline of
+     *            an Edge, this would be Edge::asPolyLine)
+     * @param simpleMergeStrategy
+     *            a simple merge strategy to use if the before views are missing
+     * @param diffBasedMergeStrategy
+     *            a merge strategy that relies on the before views to perform a more complex merge
+     * @return a {@link MergedMemberBean} containing the merged beforeMember view and the merged
+     *         afterMember view.
+     */
+    static <M> MergedMemberBean<M> mergeMember(final String memberName,
```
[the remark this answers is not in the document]
lucaspcram: I agree that it is. But my thinking is the following:

While the logic is complex, ultimately it is self contained. I feel like splitting it into a bunch of separate methods would make the logic less clear. I think there's a clear sequence the logic is following, which I attempt to elucidate in the comments.

That being said I am open to hearing suggestions on which pieces could be broken out without interrupting the flow of the function.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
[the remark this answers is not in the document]
lucaspcram: The other solution, which would appease sonar, would could introduce a breaking change like:
```
- public static final Polygon CENTER = new Polygon(Location.CENTER);
+ public static Polygon getCenter()
+ {
+     return new Polygon(Location.CENTER)
+ }
```
This would require changing unit/int tests. Not sure how widespread usage of `CENTER` would be in downstream repositories. I would guess relatively minimal impact.

Perhaps this is worth doing in a separate sonar cleanup PR?

PR #394, opened 2019-03-26
Fix source-sink bug with DAG processGroups

matthieun:
### Description:

DAG ProcessGroups would be stuck in the event of a Vertex being a source and a sink at the same time.

Also:
- Fixed some sonar issues.

### Potential Impact:

One less bug

### Unit Test Approach:

Added one unit test for the use case.

### Test Results:

Tests pass

------


matthieun:
Bug was introduced by #382

PR #402, opened 2019-04-02
Add new destination tags

snb3300:
### Description:
Added more destination tags for osm ways. Example: https://www.openstreetmap.org/way/673553199

### Potential Impact:
This PR is not adding any logic so no potential downstream impact. We will be supporting some new OSM tags

### Unit Test Approach:

Added unit tests

### Test Results:

Tests passed

------


lucaspcram:
Nice additions, thanks.

PR #407, opened 2019-04-05
ChangeAtlas performance improvements

matthieun:
### Description:

> Below, "forgiveness" means the action `ChangeAtlas` takes to infer potentially missing `FeatureChange` object. For example, if one point is removed, but the relation it belongs to is not updated, `ChangeAtlas` will automatically update the relation member list to get rid of the point.

- `ChangeAtlas` caches its relations, which are very expensive to compute (due to forgiveness).
- `ChangeRelation` caches its member list, which is very expensive to compute (due to forgiveness).
- `ChangeEdge` caches its start/end nodes, which are very expensive to compute (due to forgiveness).
- `ChangeNode` caches its in/out edges, which are very expensive to compute (due to forgiveness).
- All ChangeItems cache their parent relations, which are very expensive to compute (due to forgiveness).
- `AtlasChangeGenerator`'s `expandNodeBounds` retooled with an index to avoid the nested for loops.
  - Added a `// NOSONAR` here to avoid breaking that method in a lot of other static methods within the interface.

### Potential Impact:

A nice speedup in ChangeAtlas validation (quadratic to linear)

### Unit Test Approach:

Updated existing tests.

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
@@ -1,7 +1,9 @@
 package org.openstreetmap.atlas.geography.atlas.change;
 
+import java.util.Map;
 import java.util.Objects;
 import java.util.Optional;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.BiFunction;
 import java.util.function.LongFunction;
 import java.util.function.Supplier;
@@ -38,6 +40,7 @@ import org.openstreetmap.atlas.utilities.collections.MultiIterable;
 public class ChangeAtlas extends AbstractAtlas // NOSONAR
 {
     private static final long serialVersionUID = -5741815439928958165L;
+    private static final ChangeRelation NULL_PLACEHOLDER = new ChangeRelation(null, null, null);
 
     private final Change change;
     private final Atlas source;
@@ -51,6 +54,10 @@ public class ChangeAtlas extends AbstractAtlas // NOSONAR
     private transient Long numberOfLines;
     private transient Long numberOfPoints;
     private transient Long numberOfRelations;
+
+    // Computing relations in ChangeAtlas is very expensive, so we cache them here.
+    private transient Map<Long, ChangeRelation> relationsCache;
+    private transient Object relationsCacheLock = new Object();
 
     private static void checkChanges(final Change... changes)
     {
@@ -321,7 +328,25 @@ public class ChangeAtlas extends AbstractAtlas // NOSONAR
     @Override
     public Relation relation(final long identifier)
     {
-        final Relation relation = entityFor(identifier, ItemType.RELATION,
+        final Long longIdentifier = identifier;
+        Map<Long, ChangeRelation> localRelationCache = this.relationsCache;
+        if (localRelationCache == null)
+        {
+            synchronized (this.relationsCacheLock)
+            {
+                localRelationCache = this.relationsCache;
+                if (localRelationCache == null)
+                {
+                    this.relationsCache = new ConcurrentHashMap<>();
+                }
+            }
+        }
+        if (this.relationsCache.containsKey(longIdentifier))
+        {
+            final ChangeRelation cached = this.relationsCache.get(longIdentifier);
+            return cached == NULL_PLACEHOLDER ? null : cached;
+        }
+        final ChangeRelation relation = entityFor(identifier, ItemType.RELATION,
                 () -> this.source.relation(identifier),
                 (sourceEntity, overrideEntity) -> new ChangeRelation(this, (Relation) sourceEntity,
                         (Relation) overrideEntity));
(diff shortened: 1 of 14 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
```diff
@@ -31,6 +31,10 @@
     private final Relation source;
     private final Relation override;
 
+    // Computing ChangeRelation members is very expensive, so we cache it here.
+    private transient RelationMemberList membersCache;
+    private transient Object membersCacheLock = new Object();
```
adahn6: never seen the transient keyword before! that's cool

matthieun: Yep, it means it will not be serialized, and when deserializing set to `null`.

matthieun:
Re-working some of this. Closing until then..

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
lucaspcram: It has been the source of many bugs in the past :P

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
```diff
@@ -52,6 +64,30 @@
     private transient Long numberOfPoints;
     private transient Long numberOfRelations;
 
+    // Computing relations in ChangeAtlas is very expensive, so we cache them here.
+    private transient Map<Long, ChangeRelation> relationsCache;
```
lucaspcram: This is very clever. Leveraging the `atlas` immutability is a great idea. We really shouldn't have been recomputing relations on the fly every time one was requested from the `ChangeAtlas`.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
```diff
@@ -321,26 +368,23 @@ public Point point(final long identifier)
...
-        return relation;
+        final Predicate<ChangeRelation> nullableRelation = relationCandidate -> relationCandidate
+                .members().isEmpty();
+
+        final Supplier<ChangeRelation> creator = () -> entityFor(identifier, ItemType.RELATION,
+                () -> this.source.relation(identifier),
+                (sourceEntity, overrideEntity) -> new ChangeRelation(this, (Relation) sourceEntity,
+                        (Relation) overrideEntity));
+
+        return getFromCacheOrCreate(this.relationsCache, cache -> this.relationsCache = cache,
```
lucaspcram: I assume that returning a cached `ChangeRelation` on re-requests significantly speeds up the validation step. Is there any significant memory impact? Since validation touches all relations, you are going to have a full cache after validation I would imagine. Unless I am missing something.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeEdge.java
```diff
@@ -28,6 +29,18 @@
     private final Edge source;
     private final Edge override;
 
+    // Computing Parent Relations is very expensive, so we cache it here.
+    private transient Set<Relation> relationsCache;
+    private transient Object relationsCacheLock = new Object();
+
+    // Computing Start Node is very expensive, so we cache it here.
```
lucaspcram: Why was computing the start node expensive? Wouldn't we just check the parent `ChangeAtlas` for the presence of a source/override, and return the appropriate one? There must be an extra step there I am glossing over.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
matthieun: Yep, but this is also a risk with the memory pressure. This will duplicate memory needed to make this work.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java
matthieun: Yes, full cache after validation, which means each feature is duplicated every time it is accessed. Here they are all duplicated at the end of validation.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeEdge.java
matthieun: Yes, as if a start node is not present, we apply the forgiveness principle and we remove that edge, which in turn removes it from the other connected node, and from the parent relation etc.

lucaspcram: Ah right forgot about that implication. In that case this makes sense.

PR #411, opened 2019-04-10
use configuration to handle water types

snb3300:
### Description:
This PR modifies the way we create ```ComplexWaterEntity``` using ```ComplexWaterEntityFinder```. Currently ```ComplexWaterEntityFinder``` uses different handler classes to handle different water feature types. For example we have classes for ```LakeHandler```, ```RiverHandler``` etc. And in these classes we have strict rules for determining if a water feature can be a lake or river. We also specify what kind of Atlas entities can be turned into specific water feature types in the handlers. If we want to change the filter for any of these or want to add more types then we need to update these handler classes or create new ones. Instead of this approach, this PR removes the class based water feature handlers and accepts a configuration file to create these. This way it is easier to configure and the filters can change dynamically.  

### Potential Impact:

Downstream won't be impacted if default ```water-handler.json``` is used. In case of updated configuration downstream processes need to handle it.

### Unit Test Approach:

Made sure that default water handler configuration behaves exactly similar to the class based approach

### Test Results:
All the unit tests passed with the default config file

------


src/main/java/org/openstreetmap/atlas/utilities/configuration/ConfiguredFilter.java
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    private Predicate<AtlasEntity> getFilter()
+    {
+        if (this.filter == null)
+        {
+            Predicate<AtlasEntity> localTemporaryPredicate = atlasEntity -> true;
+            if (!this.predicate.isEmpty())
+            {
+                final StringToPredicateConverter<AtlasEntity> predicateReader = new StringToPredicateConverter<>();
```
lucaspcram: Nice use of this.

PR #412, opened 2019-04-10
Change Event Handling

yazad3:
### Description:

`CompleteEntity`-ies now have a mechanism and framework to "inform" a set of "listeners" about the changes that it is undergoing. Currently I have added a generic Listener and listeners for `tag` changes. 

In the near future we can have similar events be triggered for `relation`-ship and `geometry` changes.

In order to reduce redundency I have also gone ahead and used the magic of generics to remove repetative code in several `tag` methods in specific `CompleteEntity` implementations.

### Potential Impact:

None.

### Unit Test Approach:

Wrote a unit test that proves that the correct `event` is triggered with the correct `event` data, also making sure no junk events are triggered (for example when replacing a tag we don't trigger add and delete events but a replace event).

### Test Results:

The `event`s trigger as expected

------


src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -199,6 +203,11 @@ public static FeatureChange remove(final AtlasEntity reference, final Atlas atla
         }
     }
 
+    public String getFeatureChangeId()
```
matthieun: Let's do "Identifier", we generally avoid abbreviations.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
```diff
@@ -31,6 +35,32 @@
     private SortedSet<Long> outEdgeIdentifiers;
     private Set<Long> relationIdentifiers;
 
+    CompleteNode(final long identifier)
+    {
+        this(identifier, null, null, null, null, null);
+    }
+
+    public CompleteNode(final Long identifier, final Location location,
```
matthieun: Is there any chance to reverse that re-ordering (and the other ones)? We order static methods at the top of the class, and here some static methods are being pushed down.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteRelation.java
```diff
@@ -36,6 +39,9 @@
     private Long osmRelationIdentifier;
     private Set<Long> relationIdentifiers;
 
+    @Delegate
+    private final TagChangeDelegate tagChangeDelegate = TagChangeDelegate.newTagChangeDelegate();
```
matthieun: Could that have any impact on Serialization? Otherwise I would make it transient here!

src/test/java/org/openstreetmap/atlas/geography/atlas/change/eventhandling/listener/TestTagChangeListenerImpl.java
```diff
@@ -0,0 +1,31 @@
+package org.openstreetmap.atlas.geography.atlas.change.eventhandling.listener;
+
+import org.openstreetmap.atlas.geography.atlas.change.eventhandling.event.TagChangeEvent;
+
+/**
+ * @author Yazad Khambata
+ */
+public class TestTagChangeListenerImpl implements TagChangeListener
```
matthieun: `TestTagChangeListenerImplementation`

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
yazad3: Ok.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
yazad3: Ok, I will move this

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteRelation.java
yazad3: The classes (and interfaces) in question have been marked Serialziable.

src/test/java/org/openstreetmap/atlas/geography/atlas/change/eventhandling/listener/TestTagChangeListenerImpl.java
yazad3: Will do

yazad3:
@matthieun - thank you for the review, I have addressed your comments.

PR #417, opened 2019-04-15
Count point barriers only

matthieun:
### Description:

Make sure only points are counted for barriers.

### Potential Impact:

Barrier counts might go slightly down. Most barriers are points.

### Unit Test Approach:

Use existing unit tests

### Test Results:

N/A

------


lucaspcram:
What's the motivation here? Are non-`Point` barriers considered invalid?

matthieun:
No, we just do not want to count them :)

lucaspcram:
Ah this is for stats! I didn't look carefully enough :)

PR #418, opened 2019-04-16
Conflicting beforeView forgiveness for Relations

lucaspcram:
### Description:

Since `Relation` views may differ across shards, we need some way to forgive inconsistent `RelationBean` beforeViews when merging relevant `FeatureChange`s. This PR provides that functionality, as well as additional tests.

Still to come: we must implement similar logic for `Node` `inEdge`/`outEdge` sets. This will be added in a future PR.

Additionally, we have implemented beforeView merging for `REMOVE`, since now there are cases (`Relation` and `Node`) where we occasionally must rectify inconsistent beforeViews.

~**DNM Notes:** We may need to rectify the fact that `FeatureChange` `REMOVE`s are no longer a simple merge case, since before views may conflict. For instance, suppose `Relation` 12 is removed by two different `FeatureChanges` which were generated on two different shards. If those shards have different views of `Relation` 12, then we need to merge those beforeViews properly in the `REMOVE` case. Let's not merge until this is rectified.~

### Potential Impact:

Code that previously failed with merge conflicts may now pass.

### Unit Test Approach:

See additional unit tests.

### Test Results:

Tests pass.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
```diff
@@ -272,6 +109,92 @@ else if (afterEntityLeft instanceof Relation)
...
+            final RelationBean rightOsmMembers = beforeRelationRight.allKnownOsmMembers().asBean();
+
+            if (!leftMembers.equalsIncludingExplicitlyExcluded(rightMembers))
+            {
+                mergedBeforeRelation.withMembers(RelationBean.mergeBeans(leftMembers, rightMembers),
+                        Rectangle.forLocated(beforeRelationLeft, beforeRelationRight));
+            }
+            if (!leftOsmMembers.equalsIncludingExplicitlyExcluded(rightOsmMembers))
+            {
+                mergedBeforeRelation.withAllKnownOsmMembers(
```
lucaspcram: Do we actually need to merge `allKnownOsmMembers`? I'm not sure if this field is ever inconsistent across shards. It sounds like it isn't.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
```diff
@@ -654,35 +605,55 @@ private static FeatureChange mergeRelations(final FeatureChange left, final Feat
...
+        final MergedMemberBean<RelationBean> mergedAllKnownMembersBean = new MemberMerger.Builder<RelationBean>()
+                .withMemberName("allKnownOsmMembers").withBeforeEntityLeft(beforeEntityLeft)
+                .withAfterEntityLeft(afterEntityLeft).withBeforeEntityRight(beforeEntityRight)
+                .withAfterEntityRight(afterEntityRight)
+                .withMemberExtractor(entity -> ((Relation) entity).allKnownOsmMembers() == null
+                        ? null : ((Relation) entity).allKnownOsmMembers().asBean())
+                .withAfterViewNoBeforeMerger(MemberMergeStrategies.simpleRelationBeanMerger)
+                .withAfterViewConsistentBeforeMerger(
+                        MemberMergeStrategies.diffBasedRelationBeanMerger)
+                .withAfterViewConflictingBeforeMerger(
```
lucaspcram: Again, do we need to resolve conflicting beforeViews here? If `allKnownOsmMembers` is consistent across shards, then I think not.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
matthieun: It looks like it is not.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChangeMergingHelpers.java
matthieun: Same I think...

src/main/java/org/openstreetmap/atlas/geography/atlas/change/MemberMerger.java
```diff
@@ -0,0 +1,438 @@
...
+
+        try
+        {
+            afterMemberResult = this.afterViewConflictingBeforeMerger.apply(beforeMemberLeft,
+                    afterMemberLeft, beforeMemberRight, afterMemberRight);
+        }
+        catch (final Exception exception)
+        {
+            throw new CoreException(
+                    "Attempted afterViewConflictingBeforeMerge merge failed for {} with beforeView: {} vs {}; afterView: {} vs {}",
```
matthieun: Is there a way to make this message more descriptive? It seems afterViewConflictingBeforeMerge is very implementation-specific, and when a user sees that in a stack trace, they wont understand

lucaspcram: I updated the message, let me know what you think. We can always change these messages in the future should we think of something better.

PR #408, opened 2019-04-08
Changed MultiNode to deterministically pick master node

adahn6:
### Description:

This PR updates the logic inside the MultiAtlasOverlappingNodesFixer to deterministically choose the master node based on comparing the underlying identifiers and choose the highest value one. There is no difference in the resulting data other than the identifiers.

### Potential Impact:
None

### Unit Test Approach:

None

### Test Results:

Describe other (non-unit) test results here.

------

In doubt: [Contributing Guidelines](CONTRIBUTING.md)

PR #421, opened 2019-04-19
Fixed MultiAtlas load with filter, and added tests

adahn6:
### Description:

This PR includes a fix for MultiAtlas.loadFromPackedAtlas() for loading over an Iterable of Resources with a Predicate filter. In cases in which the Predicate filters out all entities from a valid Atlas Resource in the Iterable, the result is an empty Optional; the previous code simply invokes .get() without checking, causing an error. This fix filters out all empty Optionals from the filtered Atlas Resources before loading, removing this error. 

Additionally, a few new tests have been added around this behavior and other [recent MultiAtlas behavior changes](https://github.com/osmlab/atlas/pull/408). As part of that change, lots of tests that were depending on PackedAtlas.getAtlas() and other Atlases built directly through PackedAtlasBuilder in test methods have been updated to have their own Rules with TestAtlases defined. 

Lastly, TestAtlas.Relation was augmented to allow for an OSM identifier to be specified-- this allows a synthetically sliced relation to be easily defined in TestAtlas annotations. Without being specified, it will default to the same identifier as the Relation.

### Potential Impact:
None, as this is just one bug fix and some tests were changed.

### Unit Test Approach:
Two unit tests capturing behavior previously not tested for, additionally reworked and streamlined test definitions.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
@@ -197,7 +197,7 @@ public class MultiAtlas extends AbstractAtlas
                         .stream(Iterables.translate(resources,
                                 resource -> PackedAtlas.load(resource).subAtlas(filter,
                                         AtlasCutType.SOFT_CUT)))
-                        .filter(optional -> optional.isPresent()), optional -> optional.get());
+                        .filter(Optional::isPresent), Optional::get);
         return new MultiAtlas(validAtlases, lotsOfOverlap);
     }
 
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
@@ -192,12 +192,13 @@ public class MultiAtlas extends AbstractAtlas
         {
             throw new CoreException("Can't create an atlas from zero resources");
         }
-        return new MultiAtlas(
-                Iterables
-                        .translate(resources,
-                                resource -> PackedAtlas.load(resource)
-                                        .subAtlas(filter, AtlasCutType.SOFT_CUT).get()),
-                lotsOfOverlap);
+        final Iterable<Atlas> validAtlases = Iterables
+                .translate(Iterables
+                        .stream(Iterables.translate(resources,
+                                resource -> PackedAtlas.load(resource).subAtlas(filter,
+                                        AtlasCutType.SOFT_CUT)))
+                        .filter(optional -> optional.isPresent()), optional -> optional.get());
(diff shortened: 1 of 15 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasTestRule.java
```diff
@@ -0,0 +1,105 @@
...
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Loc;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Node;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Point;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Relation;
+import org.openstreetmap.atlas.utilities.testing.TestAtlas.Relation.Member;
+
+/**
+ * @author samg
+ */
+public class PackedAtlasTestRule extends CoreTestRule
```
lucaspcram: Much appreciated updates. This makes the tests easier to understand.

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlas.java
```diff
@@ -192,12 +192,13 @@ public static MultiAtlas loadFromPackedAtlas(final Iterable<? extends Resource>
...
-                        .translate(resources,
-                                resource -> PackedAtlas.load(resource)
-                                        .subAtlas(filter, AtlasCutType.SOFT_CUT).get()),
-                lotsOfOverlap);
+        final Iterable<Atlas> validAtlases = Iterables
+                .translate(Iterables
+                        .stream(Iterables.translate(resources,
+                                resource -> PackedAtlas.load(resource).subAtlas(filter,
+                                        AtlasCutType.SOFT_CUT)))
+                        .filter(optional -> optional.isPresent()), optional -> optional.get());
```
lucaspcram: Small nit:
`Optional::isPresent` and `Optional::get`

adahn6: good call!

src/test/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlasTest.java
```diff
@@ -56,132 +51,13 @@
...
-        builder.addPoint(6, Location.TEST_1, pointTags);
-        builder.addPoint(7, Location.TEST_5, pointTags);
-
-        // Add relations
-        builder.addRelation(1, 1, structure1, relationTags);
-        builder.addRelation(2, 2, structure2, relationTags);
-
-        // Final call
-        this.atlas = (PackedAtlas) builder.get();
+        this.atlas = this.setup.getAtlas().cloneToPackedAtlas();
```
matthieun: Nice!

PR #423, opened 2019-04-23
Improve zip reading performance

matthieun:
### Description:

A method was missing in `ZipResource`. The iterator over each resource was only implementing the `int read()` method, forcing any buffered reader to fall back to that slow method. This PR adds `int read(final byte[] b, final int off, final int len)` to allow for faster buffer reading.

- Also squashed one sonar issue related to the `NoSuchElementException`

### Potential Impact:

Dramatic speed improvement in reading zip resources (including proto-serialized Atlas files): ~15x (tested with a profiler).

### Unit Test Approach:

Use existing unit tests.

### Test Results:

Tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipResource.java
+++ b/src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipResource.java
@@ -104,10 +104,10 @@ public class ZipResource
                             }
 
                             @Override
-                            public int read(final byte[] b, final int off, final int len)
+                            public int read(final byte[] bYte, final int offset, final int length)
                                     throws IOException
                             {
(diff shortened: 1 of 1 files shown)


adahn6:
wow!

src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipResource.java
```diff
@@ -101,6 +102,13 @@ public int read() throws IOException
                             {
                                 return ZipIterator.this.input.read();
                             }
+
+                            @Override
+                            public int read(final byte[] bYte, final int offset, final int length)
```
lucaspcram: `bYte` -> `buffer`?
I think `buffer` would be the idiomatic name here.

PR #424, opened 2019-04-24
Fixed a copy-reference bug with CompleteEntity tag updates

lucaspcram:
### Description:

The `CompleteEntity` methods `withAddedTag`, `withReplacedTag`, and `withRemovedTag` were broken when multiple `ChangeAtlas`es were chained together. This was due to a pointer copy in the `addNewTag` and `removeTag` methods, which caused disparate `CompleteEntities` to share the same underlying tag map. We now make copies of the maps, so updating one entity will not inadvertently update an unrelated `CompleteEntity` that happened to initialize itself with the same underlying map.

### Potential Impact:

May impact code that relied on the previous buggy behavior.

### Unit Test Approach:

Added `testStackedChangeAtlasesWithTagChanges` to `ChangeAtlasTest` which demonstrates the buggy behavior.

### Test Results:

All tests now pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
@@ -220,12 +220,6 @@ public class CompleteNode extends Node implements CompleteLocationItem<CompleteN
     @Override
     public void setTags(final Map<String, String> tags)
     {
-        this.tags = tags;
-    }
-
-    @Override
-    public void setTags(final Map<String, String> tags)
-    {
         this.tags = tags != null ? new HashMap<>(tags) : null;
     }
 
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
@@ -1,5 +1,6 @@
 package org.openstreetmap.atlas.geography.atlas.complete;
 
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Objects;
 import java.util.Set;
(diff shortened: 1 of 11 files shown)


yazad3:
Great job @lucaspcram! Thank you for the fix.

lucaspcram:
> Please add a test case that does the same set of change but without the context. Ideally with and without `context` being passed to the `CompleteEntity`.

@yazad3 Check the PR I added a test for this. It looks to be OK.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
```diff
@@ -189,6 +190,19 @@ public boolean isShallow()
...
+    @Override
+    public void setTags(final Map<String, String> tags)
+    {
+        if (tags != null)
+        {
+            this.tags = new HashMap<>(tags);
+        }
+        else
+        {
+            this.tags = new HashMap<>();
```
lucaspcram: @matthieun In the case that the parameter `tags` is `null`, should we be setting `this.tags` to an empty map? Or should we be doing `this.tags = null;`. Curious to hear your thoughts.

yazad3:
@lucaspcram - approved the changes. Thank you.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
matthieun: I believe we should keep it null in that case, as shallow version contain null members.

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
```diff
@@ -219,7 +220,7 @@ public void setExplicitlyExcludedOutEdgeIdentifiers(final Set<Long> edges)
     @Override
     public void setTags(final Map<String, String> tags)
     {
-        this.tags = tags;
+        this.tags = tags != null ? new HashMap<>(tags) : null;
```
matthieun: 👍

PR #433, opened 2019-05-15
CompleteEntity now supports generic geometry updates

lucaspcram:
### Description:

This will help prevent tedious code duplication when trying to update the geometry for a `CompleteEntity` with an unspecified compile-time type.

### Potential Impact:

N/A

### Unit Test Approach:

N/A

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
@@ -170,9 +170,8 @@ public class CompleteNode extends Node implements CompleteLocationItem<CompleteN
          * Note that the Edges returned by this method will technically break the Located contract,
          * since they have null bounds.
          */
-        return this.inEdgeIdentifiers == null ? null
-                : this.inEdgeIdentifiers.stream().map(CompleteEdge::new)
-                        .collect(Collectors.toCollection(TreeSet::new));
+        return this.inEdgeIdentifiers == null ? null : this.inEdgeIdentifiers.stream()
+                .map(CompleteEdge::new).collect(Collectors.toCollection(TreeSet::new));
     }
 
     @Override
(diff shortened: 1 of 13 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
```diff
@@ -243,6 +243,12 @@ public CompleteNode withBoundsExtendedBy(final Rectangle bounds)
         return this;
     }
 
+    @Override
+    public CompleteEntity withGeometry(final Iterable<Location> locations)
+    {
+        return this.withLocation(locations.iterator().next());
```
adahn6: can we throw a check here, just to be safe?

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompletePoint.java
```diff
@@ -177,6 +177,12 @@ public CompletePoint withBoundsExtendedBy(final Rectangle bounds)
         return this;
     }
 
+    @Override
+    public CompleteEntity withGeometry(final Iterable<Location> locations)
+    {
+        return this.withLocation(locations.iterator().next());
```
adahn6: same here

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteNode.java
lucaspcram: What kind of check do you have in mind? A `hasNext()` check that throws a `CoreException` if `false`?

lucaspcram: @adahn6 I made a change which I think addresses your concerns.

adahn6: that looks great! sorry I should have been more specific-- I realize we'd get an iterator exception somewhere if this element doesn't exist, but I feel like wrapping it in our CoreException makes it easier to debug / find in logs / etc

matthieun:
Would be nice to add a quick unit test here and there..

lucaspcram:
@matthieun I'll add a couple tests.

PR #434, opened 2019-05-15
Reworked sectioning logic

adahn6:
### Description:

This PR changes the sectioning logic to allow entities that pass both area and edge filter to become both instead of just an edge. This is useful primarily for [pedestrian highways](https://wiki.openstreetmap.org/wiki/Key:area), which are tagged as both highway=pedestrian (thus typically qualifying as Edges) and area=yes (thus should be considered Areas as well).

### Potential Impact:
Limited impact, though now pedestrian areas will appear simultaneously as Edges and as Areas which is a change from previous behavior of only being Edges.

### Unit Test Approach:
Added unit test for expected behavior.

### Test Results:
------


adahn6:
@ranek

PR #441, opened 2019-05-22
Dynamic tile sharding equals function 

jklamer:
### Description:

Useful for knowing if two dynamic tile shadings are the same. 
Does a BFS thought both tree and exits as soon as it finds inequality. Tolerant to different children tile orderings.

### Potential Impact:

None. New Feature.

### Unit Test Approach:

Wrote a script to generate random tree with leaves between 3 and 10. (@ranek and @gitclonefun made sure I did it right). Then modified this tree to be missing children in one file and with different children ordering in another file. 

Created two identical sharding trees with the original file with different pointers and ensured equality. 
Created a tree with the missing children, ensured inequality. 
Created a tree with the different children ordering, ensured equality.

### Test Results:

passed. 

------


src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -114,6 +118,49 @@ public boolean equals(final Object other)
...
+        {
+            final Comparator<Node> nodeCompare = Comparator.comparing(Node::getTile);
+            // BFS through both trees to get equality
+            final Queue<Node> queue = new LinkedList<>();
+            queue.offer(this);
+            queue.offer(other);
+            while (!queue.isEmpty())
+            {
+                final Node node1 = queue.poll();
+                final Node node2 = queue.poll();
```
MikeGost: Maybe I'm not reading this correctly, you're checking that the `queue` is not empty before polling and then polling twice. Is there a chance that the queue only has 1 element and when you poll the second time, `node2` is `null` and a NP happens on line 139 when `node2.getChildren()` is called?

jklamer: Initially I did this, https://github.com/osmlab/atlas/pull/441/commits/3beac046d3331c5a90ee2ef6c20ab6254d29c742) but then sonar complained I wasn't checking for the existence of the first node. At which point I realized I didn't need to worry about it because I check the size of the children to make sure they're equal. So I am always adding two nodes at a time.

jklamer: Less safe if someone changes the function later. But, removes unnecessary checking.

MikeGost: That makes perfect sense, thanks for the explanation! Maybe worth a comment in case the next reader has the same question.

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -114,6 +118,50 @@ public boolean equals(final Object other)
...
+                // We always offer two at a time, so we can poll two at a time.
+                final Node node1 = queue.poll();
+                final Node node2 = queue.poll();
+                if (node1.equals(node2) && node1.getChildren().size() == node2.getChildren().size())
+                {
+                    final List<Node> children1 = node1.getChildren();
+                    final List<Node> children2 = node2.getChildren();
+                    children1.sort(nodeCompare);
+                    children2.sort(nodeCompare);
+                    for (int itr = 0; itr < children1.size(); itr++)
```
lucaspcram: nit: either make this `i` or spell out the abbreviation

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
```diff
@@ -329,6 +377,29 @@ private DynamicTileSharding()
         this.root = new Node();
     }
 
+    @Override
+    public boolean equals(final Object obj)
```
lucaspcram: nit: `obj` -> `object` purely for philosophical consistency :)

src/main/java/org/openstreetmap/atlas/geography/sharding/DynamicTileSharding.java
jklamer: can't make it i for check style I believe

lucaspcram: Checkstyle should allow one letter variable names in loops, per the config we currently have:
```
<module name="LocalVariableName" >
                        <property name="format" value="^([a-z][a-zA-Z0-9]{2,}$|[e])$"/>
                        <property name="allowOneCharVarInForLoop" value="true"/>
</module>
```
If this is not the case (and you are seeing Checkstyle failures) please let us know because that is a potential config issue on our part we will need to address.

lucaspcram: I am going to merge this for now since it's not a huge deal - but again I would like to know if using `i` is causing your `checkstyleMain` to fail because that would be a config issue on our part.

PR #456, opened 2019-06-07
 Added meta-data to FeatureChange

matthieun:
### Description:

Added a string map to `FeatureChange` to serve as meta-data. When two `FeatureChange`s are merged, the meta-data is merged as well. When a key conflicts, the values are merged if identical, or appended and separated by a comma.

Also:
- Gradle Wrapper 5.4.1

### Potential Impact:

None

### Unit Test Approach:

Added new tests, updated existing ones.

### Test Results:

Tests pass

------


src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -522,6 +519,27 @@ public void save(final WritableResource resource)
...
 
+    /**
+     * Add a new key value pair to this FeatureChange's meta-data
+     * 
+     * @param key
+     *            The key
+     * @param value
+     *            The value
+     */
+    public void setMetaData(final String key, final String value)
```
lucaspcram: Maybe call this `addMetaData` instead. I think `set` implies that it will wipe the current metadata which seems not to be the case.

PR #457, opened 2019-06-10
Reworked antimeridian handling to allow for both -180 and 180

adahn6:
### Description:
This PR reworks antemeridian handling to allow for locations at both -180 and 180 longitude. Previously, the maximum longitude was 179.99999, with any values at 180 mapping back to -180. Now, entities with longitudes of 180 will *not* be mapped to -180, allowing for true representations of both the antimeridian east and antemeridian west. In order to maintain connectivity of the edge network across the antemeridian, MultiAtlas will still treat nodes at either -180 or 180 as having equivalent longitude and follow the same behavior as other overlapping nodes to merge the two together and create a connected network. It is worth noting, however, that this behavior lies in MultiAtlas-- should an Atlas be directly built with data from both sides of the antimeridian, the original data will be preserved as-is and not be treated as overlapped nodes.

Screenshots of the antimeridians in Fiji:
Screen Shot 2019-06-10 at 1 04 32 PM
Screen Shot 2019-06-10 at 12 54 21 PM

### Potential Impact:
Possible, but should be minimally impactful in a negative way as MultiAtlas behavior should remain the same across the antimeridian. 

### Unit Test Approach:
N/A -- existing tests covered behavior.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Location.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Location.java
@@ -498,7 +498,6 @@ public class Location
             lambda *= -1;
         }
         return new Location(Latitude.radians(pheta), Longitude.radians(lambda));
-
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/geography/Location.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Location.java
@@ -492,8 +492,13 @@ public class Location
 
         // Normalize to -180/180
         lambda = (lambda + FACTOR_OF_3 * Math.PI) % (2 * Math.PI) - Math.PI;
-
+        if (this.getLongitude().equals(Longitude.MAXIMUM)
+                && that.getLongitude().equals(Longitude.MAXIMUM))
+        {
+            lambda *= -1;
+        }
         return new Location(Latitude.radians(pheta), Longitude.radians(lambda));
+
     }
 
     /**
--- a/src/main/java/org/openstreetmap/atlas/geography/Longitude.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Longitude.java
@@ -3,8 +3,7 @@ package org.openstreetmap.atlas.geography;
 import org.openstreetmap.atlas.utilities.scalars.Angle;
 
 /**
- * A Longitude between -180 degrees included and 180 degrees excluded. A Longitude of 180 degrees is
- * still accepted, but it would create the -180 degrees object.
+ * A Longitude between -180 degrees and 180 degrees included.
  *
  * @author matthieun
  * @author tony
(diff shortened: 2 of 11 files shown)


adahn6:
@ranek: heads up in case you'd like to review!

src/main/java/org/openstreetmap/atlas/geography/Location.java
```diff
@@ -492,8 +492,13 @@ public Location midPoint(final Location that)
...
         // Normalize to -180/180
         lambda = (lambda + FACTOR_OF_3 * Math.PI) % (2 * Math.PI) - Math.PI;
-
+        if (this.getLongitude().equals(Longitude.MAXIMUM)
+                && that.getLongitude().equals(Longitude.MAXIMUM))
+        {
+            lambda *= -1;
+        }
         return new Location(Latitude.radians(pheta), Longitude.radians(lambda));
+
```
MikeGost: Nit: space

src/main/java/org/openstreetmap/atlas/geography/atlas/multi/MultiAtlasOverlappingNodesFixer.java
```diff
@@ -152,6 +153,7 @@ else if (a.getIdentifier() > b.getIdentifier())
         for (final Rectangle bound : bounds)
         {
             others.addAll(Iterables.asSet(this.parent.nodesWithin(bound)));
+
```
MikeGost: Nit: space

MikeGost:
As part of this, can you also clean up `PackedAtlas.nodeIdentifierForLocation()` - it has meridian specific logic that can be cleaned up.

adahn6:
@MikeGost ah good catch! I'll change that and double check that it doesn't cause any issues to remove 👍

src/main/java/org/openstreetmap/atlas/geography/Longitude.java
```diff
@@ -3,8 +3,7 @@
 import org.openstreetmap.atlas.utilities.scalars.Angle;
 
 /**
- * A Longitude between -180 degrees included and 180 degrees excluded. A Longitude of 180 degrees is
- * still accepted, but it would create the -180 degrees object.
+ * A Longitude between -180 degrees and 180 degrees included.
```
ranek:
```suggestion
 * A Longitude between -180 degrees and +180 degrees, inclusive.
```

src/main/java/org/openstreetmap/atlas/geography/Longitude.java
```diff
@@ -134,11 +129,10 @@ public String toString()
     @Override
     protected int assertDm7(final int dm7)
     {
-        if (dm7 == MAXIMUM_DM7)
+        if (dm7 < MINIMUM_DM7 || dm7 > MAXIMUM_DM7)
         {
-            // Make one exception for the antimeridian, and make it the proper allowed value.
-            return MINIMUM_DM7;
+            throw new IllegalArgumentException("Angle dm7 value " + dm7 + " is invalid.");
```
matthieun:
```suggestion
            throw new IllegalArgumentException("Longitude dm7 value " + dm7 + " is invalid.");
```

adahn6:
@matthieun I believe the first one is present in `MultiAtlasOverlappingNodesFixerTest.testOverlappingAtAntimeridian`, but the other ones I will work on!

PR #460, opened 2019-06-14
Slicing fix and raw atlas ingest update

jwpgage:
### Description:

This PR fixes an issue in country slicing that was causing two different nodes to be assigned the same identifier. The issue was that nodes of segments of a line that fell outside a given country boundary were not processed in the same way as other lines, meaning that there were collisions in the point identifier space.

This PR also adds additional logic to the raw atlas ingest. Previously, only lines that had shape points within a given shard bounds were included in the raw atlas. However, this meant that long lines with great stretches between shape points might cross a shard boundary, but not have any nodes fall within the boundary. These lines were therefore filtered out in the process of creating the raw atlas, leaving holes in some entities when stitching atlases together.

Unit tests were updated in this PR because the number of points and lines ingested to the raw atlas increased. The country slicing fix involved processing the nodes of lines beyond the country boundary, meaning that each atlas now includes more total points than before. Also, the raw atlas ingest update now includes extra lines that were filtered out before, increasing the line and point count further.

### Potential Impact:

Raw atlases will now include more shape points and lines.

### Unit Test Approach:

Existing unit tests were updated.

### Test Results:

Updated tests pass.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
@@ -12,6 +12,7 @@ import org.openstreetmap.atlas.geography.GeometricSurface;
 import org.openstreetmap.atlas.geography.Latitude;
 import org.openstreetmap.atlas.geography.Location;
 import org.openstreetmap.atlas.geography.Longitude;
+import org.openstreetmap.atlas.geography.PolyLine;
 import org.openstreetmap.atlas.geography.atlas.Atlas;
 import org.openstreetmap.atlas.geography.atlas.builder.AtlasSize;
 import org.openstreetmap.atlas.geography.atlas.items.Line;
@@ -68,6 +69,10 @@ public class OsmPbfCounter implements Sink
     private final Set<Long> nodeIdentifiersBroughtInByWaysOrRelations = new HashSet<>();
     private final Set<Long> wayIdentifiersToInclude = new HashSet<>();
     private final Set<Long> relationIdentifiersToInclude = new HashSet<>();
+
+    // Keep track of included nodes so that they can be used in calculating if a way intersects the
+    // given shard
+    private final Map<Long, Location> nodeIdToLocation = new HashMap<>();
 
     // Keep track of excluded ways to see if we need to add them later
     private final Map<Long, Way> waysToExclude = new HashMap<>();
@@ -140,6 +145,14 @@ public class OsmPbfCounter implements Sink
 
         if (OsmPbfReader.shouldProcessEntity(this.loadingOption, rawEntity))
         {
+            // store all node locations for calculating way geometry
+            if (rawEntity instanceof Node)
+            {
+                final Node node = (Node) rawEntity;
+                final String nodeLatLon = node.getLatitude() + "," + node.getLongitude();
+                final Location nodeLocation = Location.forString(nodeLatLon);
+                this.nodeIdToLocation.put(rawEntity.getId(), nodeLocation);
+            }
             if (shouldLoadOsmNode(rawEntity))
             {
                 // This node is within the boundary, bring it in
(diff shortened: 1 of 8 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
```diff
@@ -385,17 +398,43 @@ private boolean shouldLoadOsmWay(final Entity entity)
...
+        if (wayNodesLocations.isEmpty())
+        {
+            return false;
+        }
+        final PolyLine wayGeometry = new PolyLine(wayNodesLocations);
+        if (wayGeometry.isPoint() || wayGeometry.isEmpty() || wayGeometry.bounds() == null)
+        {
+            return false;
+        }
+        // Checking the bounds of the polyline instead of the actual geometry may include some
```
adahn6: this is a great way of handling the performance requirements here! 👍

jwpgage: Thanks, it was @matthieun's idea!

adahn6:
@ranek in case you're interested

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
```diff
@@ -69,6 +71,10 @@
     private final Set<Long> wayIdentifiersToInclude = new HashSet<>();
     private final Set<Long> relationIdentifiersToInclude = new HashSet<>();
 
+    // Keep track of included nodes so that they can be used in calculating if a way intersects the
+    // given shard
+    private final Map<Long, Location> nodeIdentifierToLocation = new HashMap<>();
```
MikeGost: What do you think about combining `nodeIdentifiersToInclude` and `nodeIdentifierToLocation` to be a single Map. That way we aren't tracking the identifiers and the identifiers + locations in two separate collections?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -416,16 +416,26 @@ private Atlas cutSubAtlasForOriginalShard(final Atlas atlas)
...
+                if (countrySubAtlas.isPresent())
+                {
+                    // then, filter out all remaining entities based on the shard bounds
+                    final Optional<Atlas> finalSubAtlas = countrySubAtlas.get()
+                            .subAtlas(this.initialShard.bounds(), AtlasCutType.SILK_CUT);
+                    if (finalSubAtlas.isPresent())
+                    {
+                        return finalSubAtlas.get();
+                    }
+                    else
```
MikeGost: Is this else clause needed - can we just return null on line 438 to remove the two else clauses?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/creation/OsmPbfCounter.java
jwpgage: Hmm my comment is misleading, it should read "Keep track of ALL nodes...". `nodeIdentifiersToInclude` tracks only the IDs of nodes that should be included, whereas `nodeIdentifierToLocation` includes all nodes from the pbf (basically a node location store). This is because we need to use the location of nodes beyond the shard boundary to check for intersections. The two sets have different node spaces, so I think they shouldn't be combined to a single map.

MikeGost: This makes sense, thanks. It's essentially every location in the PBF.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
jwpgage: If the first subAtlasing at L#419 was successful but the second at L#424 created an empty atlas, then without this else clause we would return the method's atlas parameter on L#442 instead of a null atlas

PR #467, opened 2019-06-21
Updated slicing logic to include fallback to lower precision

adahn6:
### Description:
While slicing lines, line geometry is clipped against the country boundaries repeatedly until the line has been entirely sliced. The clipped geometry returned after slicing can be much higher precision than the original OSM geometry. While not always a problem, this higher precision can result in a non-noded intersection between the clipped remainder geometry and the country boundary. This issue can be fixed by reducing the precision of the clipped remainder geometry to the same as the original OSM geometry. This allows for certain geometries that were previously throwing errors to now be sliced properly.

### Potential Impact:
Minimal to none, just successful slicing under certain rare conditions.

### Unit Test Approach:
N/A because this condition is extremely hard to design a test case for, as it would rely on predicting the results of geometry errors.

### Test Results:
N/A

------


adahn6:
@ranek

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1347,7 +1350,34 @@ public int size()
...
+            Geometry clipped;
+            try
+            {
+                clipped = target.intersection(candidate);
+            }
+            catch (final TopologyException exc)
+            {
+                logger.error(
+                        "Error while using regular intersection for line {}, attempting again with reduced precision",
+                        identifier);
```
matthieun: Here I would add the exception details too!

```suggestion
                        identifier, exc);
```

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -1347,7 +1350,34 @@ public int size()
...
+                    precisionReducer.setPointwise(true);
+                    precisionReducer.setChangePrecisionModel(false);
+                    target = precisionReducer.reduce(target);
+                    clipped = target.intersection(candidate);
+                }
+                catch (final Exception newExc)
+                {
+                    logger.error(
+                            "Reduced precision still failed for line {}, rethrowing original exception",
+                            identifier);
```
matthieun: Same here:
```suggestion
                            identifier, newExc);
```

PR #468, opened 2019-06-21
Update area filter to allow in certain piers and footways

adahn6:
### Description:

Following PR #434 which allowed certain types of pedestrian highways tagged as `area=yes` to be both Areas and Edges, this PR updates the default Area filter to allow in certain `highway=footway` and `man_made=pier,area=yes` entities to also be Areas. 

### Potential Impact:
Minimal-- some entities will now be added to Areas where previously they were only Edges. 

### Unit Test Approach:
N/A

### Test Results:
N/A

------


adahn6:
@ranek

PR #471, opened 2019-06-24
ComplexRoundabout

Bentleysb:
### Description:

This adds a new ComplexEntity to represent roundabouts. 
Much of the logic is taken from [Atlas-Checks' MalformedRoundaboutCheck](https://github.com/osmlab/atlas-checks/blob/dev/src/main/java/org/openstreetmap/atlas/checks/validation/linear/edges/MalformedRoundaboutCheck.java). As such, there are a number of things that can invalidate a ComplexRoundabout. In general, a roundabout must form a single complete one way car navigable ring route that is directionalized in the correct direction for the country the roundabout is in. 
If a Complex Roundabout is valid, it will have a publicly accessible set of Edges that make up the roundabout and a Route built from those Edges. 

### Potential Impact:
No direct downstream impacts, but this can be used to replace some parts of MalformedRoundaboutCheck.

### Unit Test Approach:
The unit tests check a number of correctly and incorrectly formed roundabouts in both right and left driving countries. They check that the correct invalidation reasons are given, and that both constructors work correctly. The ComplexRoundaboutFinder is also tested.

### Test Results:
Tested running a small county through the finder and converting the resulting ComplexRoundabouts' routes into a geojson. A sampling of the results showed that the roundabouts were valid. 

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundabout.java
```diff
@@ -0,0 +1,302 @@
...
+     * @return CLOCKWISE or COUNTERCLOCKWISE if the majority of the edges have positive or negative
+     *         cross products respectively, and UNKNOWN if all edge cross products are 0 or if the
+     *         roundabout's geometry is malformed
+     */
+    private static RoundaboutDirection findRoundaboutDirection(final Route roundaboutEdges)
+    {
+        int clockwiseCount = 0;
+        int counterClockwiseCount = 0;
+
+        for (int idx = 0; idx < roundaboutEdges.size(); idx++)
```
matthieun: `index` to avoid abbreviations

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundabout.java
```diff
@@ -0,0 +1,302 @@
...
+                        String.format("Invalid source Edge (%s) for a roundabout.",
+                                source.getIdentifier()),
+                        new CoreException(EXCEPTION_MESSAGE));
+                this.invalidationReasons.add(sourceError);
+                throw sourceError.getException();
+            }
+            final String isoCountryCode = source.tag(ISOCountryTag.KEY).toUpperCase();
+
+            // Get all edges in the roundabout
+            this.roundaboutEdgeSet = new SimpleEdgeWalker(source, this.isRoundaboutEdge())
```
matthieun: Has there been any use case where two roundabouts are adjacent and connected?

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundabout.java
```diff
@@ -0,0 +1,302 @@
...
+            {
+                final ComplexEntityError routeError = new ComplexEntityError(this,
+                        INCOMPLETE_ROUTE_INVALIDATION, new CoreException(EXCEPTION_MESSAGE));
+                this.invalidationReasons.add(routeError);
+                throw routeError.getException();
+            }
+
+            // Invalidate the roundabout if any of the edges are not car navigable or master Edges,
+            // or the route does not
+            // form a closed loop.
```
matthieun: Formatting: the comment can be on the previous line.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundaboutFinder.java
```diff
@@ -0,0 +1,45 @@
...
+        {
+            if (!this.checkedIds.contains(edge.getIdentifier()))
+            {
+                final ComplexRoundabout complexRoundabout = new ComplexRoundabout(edge);
+                if (complexRoundabout.getRoundaboutEdgeSet() != null)
+                {
+                    this.checkedIds.addAll(complexRoundabout.getRoundaboutEdgeSet().stream()
+                            .map(Edge::getIdentifier).collect(Collectors.toSet()));
+                }
+                if (complexRoundabout.isValid())
```
matthieun: Is it possible that a roundabout is invalid when starting on Edge X but valid if explored from Edge Y? In that case it might be better to add its edges to the checked IDs only if it is valid?

src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundaboutTestRule.java
```diff
@@ -0,0 +1,501 @@
...
+    private static final String CLOCKWISE_5 = "38.905494761938684,-77.03236699104309";
+
+    private static final String COUNTER_CLOCKWISE_1 = "38.905361177861046,-77.03205585479736";
+    private static final String COUNTER_CLOCKWISE_2 = "38.905528157918816,-77.03158378601074";
+    private static final String COUNTER_CLOCKWISE_3 = "38.905937257400495,-77.031809091568";
+    private static final String COUNTER_CLOCKWISE_4 = "38.90588716371307,-77.03230261802673";
+    private static final String COUNTER_CLOCKWISE_5 = "38.90551980892527,-77.03236699104309";
+
+    // Clockwise roundabout, left driving country
+    @TestAtlas(
```
matthieun: When `@TestAtlas`es become complex like that it might be easier to make them out of a JOSM layer using `loadFromJosmOsmResource` :) But this works too.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundabout.java
Bentleysb: The only example of that I am aware of is the Magic Roundabout in Swindon, UK. However, that is not actually tagged roundabout. In fact, the connecting of two roundabouts in that fashion would go against the practices recommended in the OSM wiki because of the multiple connections at a single node, which is discouraged for routing reasons.

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/roundabout/ComplexRoundaboutFinder.java
Bentleysb: That shouldn't be the case, as the validation of the roundabout is done using the Route built from all the Edges of the roundabout.

matthieun:
There are some Travis errors today... Unrelated to this PR.

PR #478, opened 2019-07-05
Added BUS_TRAP as a new BarrierTag type

AlexHsieh22:
### Description:
Added bus_trap to the BarrierTag enum list.  

The uses of barrier=bus_trap can be found here: https://taginfo.openstreetmap.org/keys/barrier#values 
It is mentioned in the OSM wiki: https://wiki.openstreetmap.org/wiki/Key:barrier

### Potential Impact:
Minimal impact.  If anyone is using the BarrierTag::isBarrier, there's  new data that could be considered a barrier now (anything tagged with barrier=bus_trap).

### Unit Test Approach:
Added test case to make sure barrier=bus_trap gets picked up as a barrier and a blocking barrier properly.

### Test Results:
All current unit tests as well as new unit tests have passed locally.

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/tags/BarrierTagTestCase.java
+++ b/src/test/java/org/openstreetmap/atlas/tags/BarrierTagTestCase.java
@@ -1,0 +1,27 @@
+package org.openstreetmap.atlas.tags;
+
+import org.junit.Assert;
+import org.junit.Test;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
+
+/**
+ * Test for {@link org.openstreetmap.atlas.tags.BarrierTag}
+ *
+ * @author alexhsieh
+ */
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/tags/BarrierTagTestCase.java
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.openstreetmap.atlas.tags.annotations.validation.Validators;
+
+/**
+ * Test for {@link org.openstreetmap.atlas.tags.BarrierTag}
+ *
+ * @author alexhsieh
+ */
+public class BarrierTagTestCase
```
snb3300: 👍 for test case

PR #484, opened 2019-07-10
AtlasChangeGenerator to enforce Node location

matthieun:
### Description:

This is to enforce the presence of a location in `FeatureChange` objects for `Node`s. Since their bounding box is altered, this is to make for easier debugging: get the original node's location.

### Potential Impact:

Not much, slightly larger geojson representation.

### Unit Test Approach:

Added one new unit test.

### Test Results:

N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/AtlasChangeGenerator.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/AtlasChangeGenerator.java
@@ -76,11 +76,12 @@ public interface AtlasChangeGenerator extends Converter<Atlas, Set<FeatureChange
                     newBounds = newBounds.combine(bounds);
                 }
             }
-            CompleteNode newCompleteNode = originalCompleteNode.withBoundsExtendedBy(newBounds);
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/AtlasChangeGenerator.java
```diff
@@ -76,8 +76,13 @@
                     newBounds = newBounds.combine(bounds);
                 }
             }
+            CompleteNode newCompleteNode = originalCompleteNode.withBoundsExtendedBy(newBounds);
+            if (originalCompleteNode.getLocation() == null && originalNode != null)
+            {
+                newCompleteNode = newCompleteNode.withLocation(originalNode.getLocation());
```
jwpgage: Not sure of the context, but is the originalNode guaranteed to have a location?

lucaspcram: Yes, because it comes directly from an `Atlas`. See line 63 of this file.

PR #485, opened 2019-07-15
Upgraded find-id command to support more search criteria

lucaspcram:
### Description:
The `find-id` command is now `find`, and supports numerous additional search criteria: WKT geometry, `TaggableFilter`s, start/end node identifiers, in/out edge id sets, and parent relation identifiers. @adahn6 also requested the ability to search for points within lines. I will leave it up to him to implement this once this PR is merged.

### Potential Impact:
Users of `find-id` will notice that the command name is now `find`. Other than this change, old use cases for `find-id` will operate using the same syntax (barring the name change).

### Unit Test Approach:
Tested locally on the command line

### Test Results:
Tests look good.

------


src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/AtlasSearchCommand.java
```diff
@@ -0,0 +1,456 @@
...
+        this.parentRelations = new HashSet<>();
+    }
+
+    @Override
+    public int finish()
+    {
+        if (this.optionAndArgumentDelegate.hasOption(COLLECT_OPTION_LONG)
+                && !this.matchingAtlases.isEmpty())
+        {
+            final Atlas outputAtlas = new MultiAtlas(this.matchingAtlases);
```
matthieun: Here skip the MultiAtlas step if the matching atlas set is only one element

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/AtlasSearchCommand.java
```diff
@@ -0,0 +1,456 @@
...
+            final Path concatenatedPath = Paths.get(getOutputPath().toAbsolutePath().toString(),
+                    OUTPUT_ATLAS);
+            final File outputFile = new File(concatenatedPath.toAbsolutePath().toString());
+            new PackedAtlasCloner().cloneFrom(outputAtlas).save(outputFile);
```
matthieun: The cloning can be saved too here in case of only one match.

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/AtlasSearchCommand.java
lucaspcram: Updated! I figured leaving it this way made the code a bit simpler at some potential performance costs.

PR #487, opened 2019-07-16
Added a special exception for FeatureChange merge failures

lucaspcram:
### Description:
`FeatureChange` merge failures now throw a special exception type that also contains an enum detailing what type of merge failure occurred.

### Potential Impact:
N/A as is, but downstream code will have an easier time inspecting and handling merge errors

### Unit Test Approach:
Updated some tests to handle the new exception type

### Test Results:
Tests pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/exception/change/FeatureChangeMergeException.java
+++ b/src/main/java/org/openstreetmap/atlas/exception/change/FeatureChangeMergeException.java
@@ -12,7 +12,7 @@ public class FeatureChangeMergeException extends CoreException
 {
     private static final long serialVersionUID = -3583945839922744755L;
 
-    private final transient MergeFailureType failureType;
+    private final MergeFailureType failureType;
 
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/exception/change/FeatureChangeMergeException.java
```diff
@@ -0,0 +1,34 @@
...
+/**
+ * A special exception for {@link FeatureChange} merge errors.
+ * 
+ * @author lcram
+ */
+public class FeatureChangeMergeException extends CoreException
+{
+    private static final long serialVersionUID = -3583945839922744755L;
+
+    private final transient MergeFailureType failureType;
```
matthieun: It would be nice to not make it transient so it can be serialized around...

lucaspcram: Done.

PR #491, opened 2019-07-19
Refactored RawAtlasPointAndLineSlicer

adahn6:
### Description:
Introducing RawAtlasPointAndLineSlicer 2.0! Now includes the use of [ChangeAtlas](https://github.geo.apple.com/maps-osm/atlas/blob/42a0fba62834672d1c07449b23717de65b07860d/src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeAtlas.java) and [FeatureChanges](https://github.geo.apple.com/maps-osm/atlas/blob/42a0fba62834672d1c07449b23717de65b07860d/src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java) in place of the [SimpleChangeSet](https://github.geo.apple.com/maps-osm/atlas/blob/42a0fba62834672d1c07449b23717de65b07860d/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/SimpleChangeSet.java) and [SimpleChangeSetHandler](https://github.geo.apple.com/maps-osm/atlas/blob/42a0fba62834672d1c07449b23717de65b07860d/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/changeset/SimpleChangeSetHandler.java), as well as all of the slicing logic contained in [CountryBoundaryMap](https://github.geo.apple.com/maps-osm/atlas/blob/42a0fba62834672d1c07449b23717de65b07860d/src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java)!

Benefits include: 
* Streamlined slicing code! 
   * All logic for slicing now contained in RawAtlasPointAndLineSlicer or its parent, RawAtlasSlicer!
   * Same FeatureChanges available in other Atlas uses instead of one-off ChangeSet logic!
   * Improved readability -- Sonar now approves of the geometry slicing logic 😂  
   * Newly updated method documentation!
* Fixes for slicing bug where line slice geometries were merged without checking to ensure tags were identical
   * Now, line slice geometries will only be merged if they a) have overlapping endpoints, b) lie in the same country, and c) have identical tagging
* Somewhat improved output consistency!
   * In previous logic, sliced Atlases sometimes contained points for countries outside of the country list being sliced-- no more!

I'm sure there's more I'm forgetting about but this is a good start. 

### Potential Impact:
Hopefully none: the output Atlases are almost entirely identical to the best of my comparisons. There are sometimes fewer points than before, but all entirely points that should have been filtered out anyway -- and these points should be filtered out later during way-sectioning anyway, so the final Atlas during Atlas generation should look the same. 

### Unit Test Approach:
No new tests, a few small updates that are in line with expectations of the new behavior.

### Test Results:
DNM, still doing some final validation.

------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -76,46 +83,79 @@ public Atlas slice()
...
+            }
         }
     }
 
+    /**
+     * Given a list of slices, check that all tags for each slice are the same (i.e. no
+     * SyntheticNearestNeighborCountry tag on one slice but not the others)
+     *
+     * @param slices
+     * @return
```
MikeGost: Nit: missing return doc

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -166,6 +206,18 @@ else if (line.isClosed())
...
+     * previously existing point identifiers in the {@link Atlas} being sliced.
+     *
+     * @param pointIdentifierFactory
+     *            An instantiated factory for generating new {@link Point} identifiers
+     * @param lineIdentifier
+     *            The identifier for the {@link Line} being sliced
+     * @return An identifier that can safely be used to make a new {@link Point} in the
+     *         {@link Atlas} being sliced
+     */
+
```
MikeGost: Nit: extra line

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -182,29 +234,103 @@ private long createNewPointIdentifier(final PointIdentifierFactory pointIdentifi
...
+                CountryBoundaryMap.setGeometryProperty(clipped, ISOCountryTag.KEY, countryCode);
+                addResult(clipped, results);
+
+                // Update target to be what's left after clipping
+                currentTarget = currentTarget.difference(candidate);
+                if (currentTarget.getDimension() == 1
+                        && currentTarget.getLength() < CountryBoundaryMap.LINE_BUFFER
+                        || currentTarget.getDimension() == 2
+                                && currentTarget.getArea() < CountryBoundaryMap.AREA_BUFFER
+                                && new DiscreteHausdorffDistance(currentTarget, candidate)
```
MikeGost: This looks pretty interesting - worth a comment?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -279,150 +417,294 @@ private void processLineSlices(final Line line, final List<Geometry> slices)
...
-            final Location location = JTS_LOCATION_CONVERTER.backwardConvert(coordinate);
-            final Iterable<Point> existingRawAtlasPoints = getStartingAtlas().pointsAt(location);
-            existingRawAtlasPoints
-                    .forEach(point -> this.pointsMarkedForRemoval.add(point.getIdentifier()));
+            final org.locationtech.jts.geom.Polygon candidate = candidateIterator.next();
+            final String countryCode = CountryBoundaryMap.getGeometryProperty(candidate,
+                    ISOCountryTag.KEY);
+            if (Strings.isNullOrEmpty(countryCode))
+            {
+                logger.warn(
```
MikeGost: How often does this happen? Would it be helpful to log the polygon here or at least the identifier?

adahn6: good idea! I'll do a second pass through this whole section and make sure there's better logging

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
adahn6: good idea! I've also moved this out to a helper method, since that's a little easier to read and more reusable.

adahn6:
> These are awesome changes, very clean and easy to follow! What kind of performance gains are you seeing?

yessss... performance gains... definitely... 😛  

in all seriousness, this takes about the same amount of time! which considering the overhaul, is acceptable IMO 😄 that being said, I'm still finalizing a few more changes and will hopefully have a slightly more detailed understanding of the performance impact then.

src/test/resources/log4j.properties
```diff
@@ -10,7 +10,7 @@ log4j.appender.stdout.layout.ConversionPattern=%d{yyyy-MM-dd HH:mm:ss} %-5p [%t]
 log4j.logger.org.reflections=WARN
 
 # Show our logs
-log4j.logger.org.openstreetmap.atlas=TRACE
+log4j.logger.org.openstreetmap.atlas=DEBUG
```
matthieun: Any reason not to have Trace for tests?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -279,150 +417,294 @@ private void processLineSlices(final Line line, final List<Geometry> slices)
...
-                else
-                {
-                    // Extract relevant tag values for this slice
-                    final Map<String, String> lineTags = createLineTags(slice, line.getTags());
-                    createdLines.add(
-                            new TemporaryLine(lineSliceIdentifier, newLineShapePoints, lineTags));
+                    lineChanges.add(FeatureChange.add(newLineSlice, atlas));
+                    for (final Relation relation : line.relations())
+                    {
+                        final CompleteRelation updatedRelation = CompleteRelation.from(relation)
```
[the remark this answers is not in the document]
adahn6: hmm, do the call `CompleteLine.from(line)` not copy that line's relation identifiers?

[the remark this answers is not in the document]
lucaspcram: `CompleteLine.from(line)` will copy that line's relation identifiers, if that line is already a member of a relation. But it seems like you are adding `newLineSlice` to a relation in the `line.relations()` loop. So in that case, you have to make sure to manually point `newLineSlice` back to its parent, via `newLineSlice#withRelationIdentifiers`

src/test/resources/log4j.properties
adahn6: I found that the ChangeAtlas TRACE logs were... excessive. I could turn it off just for that package, I guess?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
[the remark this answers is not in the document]
adahn6: nope, the difference here is that we are subtracting the original Line from the Atlas and replacing it with sliced Lines-- in this case, we are simply adding the new sliced Lines to the Relations that the original Line was a member of, and relying on the `CompleteLine.from(line)` to ensure the the new sliced Lines have the Relation set of the original Line!

adahn6:
@lucaspcram I think I've updated the `.from` and `.shallowFrom` calls appropriately-- but happy to have feedback if you see anything else worth changing!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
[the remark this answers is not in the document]
lucaspcram: In that case this should work as you expect!

PR #503, opened 2019-08-13
Support for GeoHash sharding

matthieun:
### Description:

Add support for GeoHash sharding

### Potential Impact:

None

### Unit Test Approach:

Adding unit tests along with feature.

### Test Results:

Tests pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/sharding/GeoHashSharding.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/sharding/GeoHashSharding.java
@@ -21,6 +21,11 @@ public class GeoHashSharding implements Sharding
         this.precision = precision;
     }
 
+    public int getPrecision()
+    {
+        return this.precision;
+    }
+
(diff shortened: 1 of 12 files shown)


src/main/java/org/openstreetmap/atlas/geography/sharding/GeoHashSharding.java
```diff
@@ -0,0 +1,57 @@
...
+    @Override
+    public Iterable<Shard> neighbors(final Shard shard)
+    {
+        if (shard instanceof GeoHashTile)
+        {
+            return ((GeoHashTile) shard).neighbors();
+        }
+        else
+        {
+            throw new CoreException("Cannot have neighbors from another type of shard.");
```
lucaspcram: What about:
```java
throw new CoreException("shard parameter was of invalid type {}", shard.getClass().getName());
```
This may be slightly more helpful as an error message.

PR #509, opened 2019-08-20
Validation of NodeToEdgeLocationAccuracy: Better error message

matthieun:
### Description:

Improve the error message thrown when there is a mismatch between a node and the connected edge's location it references. Prior to this, it was reversed, and could mis-represent the node's location when taking it from the edge's start/end identifiers.

### Potential Impact:

Less confusion

### Unit Test Approach:

Use existing tests

### Test Results:

Tests pass

------


lucaspcram:
Great change! I ran into some confusion with this error message in the past.

PR #516, opened 2019-08-28
Implemented more CompleteRelation API methods for ease of use

lucaspcram:
### Description:
Added a `withRemovedMember` method to CompleteRelation. Instead of using clunky `stream().filter().collect()` calls, you can now do something like:
```java
final CompleteRelation completeRelation = CompleteRelation.shallowFrom(atlas.relation(1L))
.withMembers(atlas.relation(1L).members())
.withRemovedMember(atlas.point(1L));
```

### Potential Impact:
Downstream code that is using `FeatureChange` to remove relation members can be much simpler now.

### Unit Test Approach:
Added some unit tests to ensure this works.

### Test Results:
Tests OK.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
@@ -342,6 +342,34 @@ public class RelationBean extends AbstractCollection<RelationBeanItem> implement
         return RelationBean.mergeBeans(this, other);
     }
 
+    public void removeItem(final Long identifier, final ItemType itemType)
+    {
+        int firstMatchingIndex = -1;
+        for (int i = 0; i < this.beanItems.size(); i++)
+        {
+            if (this.beanItems.get(i).getIdentifier().equals(identifier)
+                    && this.beanItems.get(i).getType().equals(itemType))
+            {
+                firstMatchingIndex = i;
+            }
+        }
+
+        if (firstMatchingIndex != -1)
+        {
+            this.beanItems.remove(firstMatchingIndex);
+        }
+    }
+
+    public void removeItem(final Long identifier, final String role, final ItemType itemType)
+    {
+        removeItem(new RelationBeanItem(identifier, role, itemType));
+    }
+
+    public void removeItem(final RelationBeanItem item)
+    {
+        this.beanItems.remove(item);
+    }
+
     /**
      * @return The number of members in this {@link RelationBean}
      */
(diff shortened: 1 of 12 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
```diff
@@ -333,50 +328,85 @@ public int hashCode()
...
     public Iterator<RelationBeanItem> iterator()
     {
-        final List<RelationBeanItem> result = new ArrayList<>();
-        for (int index = 0; index < this.memberIdentifiers.size(); index++)
-        {
-            result.add(new RelationBeanItem(this.memberIdentifiers.get(index),
-                    this.memberRoles.get(index), this.memberTypes.get(index)));
-        }
-        return result.iterator();
+        return this.beanItems.iterator();
```
adahn6: 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
```diff
@@ -333,50 +328,85 @@ public int hashCode()
...
+
+        if (firstMatchingIndex != -1)
+        {
+            final String removedRole = this.beanItems.get(firstMatchingIndex).getRole();
+            this.beanItems.remove(firstMatchingIndex);
+            return Optional.of(removedRole);
+        }
+
+        return Optional.empty();
+    }
```
adahn6: :tada

src/main/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteArea.java
```diff
@@ -299,4 +306,13 @@ public CompleteArea withRelations(final Set<Relation> relations)
                 .collect(Collectors.toSet());
         return this;
     }
+
+    @Override
+    public CompleteArea withRemovedRelationIdentifier(final Long relationIdentifier)
+    {
+        this.relationIdentifiers = this.relationIdentifiers.stream()
+                .filter(keepId -> keepId != relationIdentifier.longValue())
+                .collect(Collectors.toSet());
```
adahn6: 💯  omg

src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
```diff
@@ -333,50 +328,85 @@ public int hashCode()
...
+     * If an item is actually removed, this method will return its role wrapped in an
+     * {@link Optional}. If nothing was removed, then the {@link Optional} will be empty.
+     * 
+     * @param identifier
+     *            the id to remove
+     * @param itemType
+     *            the type to remove
+     * @return an {@link Optional} wrapping the role of the removed item
+     */
+    public Optional<String> removeItem(final Long identifier, final ItemType itemType)
```
matthieun: Shouldn't it remove all the roles matching the same item? I know it is uncommon, but sometimes it happens to have the same object with different roles

lucaspcram: Good point. My initial thought was to leave it up to the user to deal with this. If they want to remove a specific role, then they should be using the `removeItem` that allows role specification. I can change the behavior here to do what you said, and then change the name to `removeAllMatchingItems` for clarity.

PR #530, opened 2019-09-13
Fixed an issue with duplicate points

adahn6:
### Description:
This PR addresses an issue in which occasionally, two consecutive Points would be inserted along a Line during Relation slicing. This issue was caused by using the full iterator of `atlas.pointsAt()`, which would occasionally return multiple Points for the same location. The intent behind the code was simply to leverage an existing Point's coordinates, not to add all Points at that location to the Line. Beyond being a minor bug, this logic sometimes introduced differences to the final Relation slicing result based on what Points were included in the Atlas-- this change now ensures the results should be identical.

### Potential Impact:
N/A
### Unit Test Approach:
N/A
### Test Results:
N/A
------


adahn6:
@matthieun done!

PR #541, opened 2019-10-08
Shard listing now uses getName instead of toString

lucaspcram:
### Description:
Now that the `Shard#getName` method has been standardized, any code attempting to write a parsable shard string should be using this method.

See: https://github.com/osmlab/atlas-generator/pull/122

### Potential Impact:
Parsing code expecting `toString` to give parsable output may be broken. This fixes that issue.

### Unit Test Approach:

Added a test that should catch any future getName issues

### Test Results:
N/A

------


lucaspcram:
@matthieun Done. Added a test for `StringToShardConverter` that should catch these issues in the future. It enforces that `Shard` implementations provide a `getName` that is parsable.

PR #551, opened 2019-10-25
LayerTag Calculate Values for Bridges and Tunnels

Bentleysb:
### Description:

This updates LayerTag's `getTaggedOrImpliedValue` method to take into account bridge and tunnel tags. Now if there is not a layer tag value but the feature is marked as a bridge or tunnel the values of 1 or -1 are respectively returned.

### Potential Impact:

Functions that call `LayerTag.getTaggedOrImpliedValue()` may see a change when evaluating some features. For most applications this should be an improvement in accuracy. However, if needed the old logic can be mimicked by using `LayerTag.getTaggedValue(taggable).orElse(impliedValue)`.

### Unit Test Approach:

Unit tests cover cases for all the tags used as well as no tags (to test the default). 

### Test Results:
None

------


src/main/java/org/openstreetmap/atlas/tags/annotations/extraction/LayerExtractor.java
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Extracts layer values as {@link Optional} {@link Long}s based on the {@link LayerTag},
+     * {@link BridgeTag}, and {@link TunnelTag}s. Defaults to 0 if no layer value is found.
+     *
+     * @param taggable
+     *            {@link Taggable} to get a value for
+     * @return {@link Optional} {@link Long} layer value. Will return empty if taggable has a
+     *         {@link LayerTag} that is invalid.
+     */
+    public static Optional<Long> validateAndExtract(final Taggable taggable)
```
MikeGost: There is some overlap between this functionality and the implementation in `LayerTag`. It might be worth keeping the tests and adding some of the bridge and tunnel logic to the existing extraction logic.

sayas01: Looks like we need to include the tunnel and bridge logic in the [areOnSameLayer()] method in LayerTag.(https://github.com/osmlab/atlas/blob/0b6cdf02255ca4f51048faaab72490544b19a196/src/main/java/org/openstreetmap/atlas/tags/LayerTag.java#L35). Currently it defaults to 0 if no layer tag is present.

Bentleysb: I thought about making this an addition to that logic, but was not sure about adding logic to look at other tags from inside a tag class. I am not sure that there is any other tag class where we do that. However, if we think that is fine then I am happy to handle this all from the LayerTag class.

MikeGost: Good callout, my vote is that this is a normal usage pattern. An example of this is in `HighwayTag`, which references `AreaTag`.

Bentleysb: All the logic has been consolidated into `LayerTag.getTaggedOrImpliedValue()`.

Bentleysb: Now that we are using `LayerTag.getTaggedOrImpliedValue()` this is handled.

PR #555, opened 2019-10-29
CountryBoundaryMapArchiver consolidation

matthieun:
### Description:

Consolidate some of the boundary operation commands in `CountryBoundaryMapArchiver`.
Now `CountryBoundaryMapArchiver` can:
- Read from:
  - SHP file
  - Atlas file
  - Serialized boundary file
- Do:
  - Fill in ocean boundaries
  - Create grid index
  - Print Geojson and Wkt for each country

Also:
- Temporary `File`s are now `Closeable` `TemporaryFile`s (meaning it gets deleted when closed)
- `StringList` has a `toArray()` method

### Potential Impact:

Toolchains using `CountryBoundaryMapArchiver` and `OceanCoundtryBoundaryMap` will have to be updated to use `CountryBoundaryMapArchiver` only.

### Unit Test Approach:

Added unit tests and integration test.

### Test Results:

Pass

------


src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMapArchiver.java
```diff
@@ -19,42 +27,119 @@
...
+                    finalBoundaryMap.addCountry(countryCode, (MultiPolygon) countryGeometry);
+                }
+                logger.info("Added Ocean Country {} in {}", countryCode, start.elapsedSince());
+            }
+            else
+            {
+                logger.info("Skipped Ocean Country {} in {}. It is land covered.", countryCode,
+                        start.elapsedSince());
+            }
+            oceanCountryCount++;
```
jwpgage: I used to increment this only if the boundary existed, so our ocean boundaries will have different names. For simplicity could we move this increment back to the if statement?

PR #558, opened 2019-11-01
Fixed bug with FeatureChange JSON serialization

lucaspcram:
### Description:
`FeatureChange` GeoJSON serialization no longer serializes `null` valued properties with `"null"`. It will now instead simply omit the field.

### Potential Impact:
Downstream users depending on `null` valued fields having a value `"null"` will need to adjust.
 
### Unit Test Approach:
Updated relevant unit tests.

### Test Results:
All pass!

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/serializer/FeatureChangeGeoJsonSerializer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/serializer/FeatureChangeGeoJsonSerializer.java
@@ -207,7 +207,6 @@ public class FeatureChangeGeoJsonSerializer
         return result;
     };
 
-    private static final String NULL = "null";
     private final Gson jsonSerializer;
 
     private static <T> void add(final JsonObject result, final String name, final T property,
@@ -215,7 +214,7 @@ public class FeatureChangeGeoJsonSerializer
     {
         if (property == null)
         {
-            result.addProperty(name, NULL);
+            result.addProperty(name, (String) null);
         }
         else
         {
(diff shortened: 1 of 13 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/serializer/FeatureChangeGeoJsonSerializer.java
```diff
@@ -207,15 +207,14 @@ else if (source instanceof Relation)
         return result;
     };
 
-    private static final String NULL = "null";
```
yazad3: Would this be used to add null as a String? As in

```json
{"name": "null"}
```

instead of,

```json
{"name": null}
```

lucaspcram: That is what it was used for before this PR. Now we no longer have need for it. Interestingly enough, it seems like gson is not even adding the null fields. So while the code explicitly sets things to null like `{"tags": null}`, they end up just being excluded from the final GeoJSON object.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/serializer/FeatureChangeGeoJsonSerializer.java
```diff
@@ -207,15 +207,14 @@ else if (source instanceof Relation)
...
     private final Gson jsonSerializer;
 
     private static <T> void add(final JsonObject result, final String name, final T property,
             final Function<T, JsonElement> writer)
     {
         if (property == null)
         {
-            result.addProperty(name, NULL);
+            result.addProperty(name, (String) null);
```
lucaspcram: @yazad3 See here. It seems like explicitly adding a null property is not actually reflected in the final object. Gson just skips the property altogether when converting to string.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/serializer/FeatureChangeGeoJsonSerializer.java
yazad3: Thank you @lucaspcram so will you be removing this constant?

lucaspcram: Yes. It's removed by this PR!

PR #559, opened 2019-11-01
Reworked relation slicing

adahn6:
### Description:
This PR makes a number of notable changes to Relation slicing. 

- Relation slicing now uses FeatureChange and ChangeAtlas, instead of the deprecated (now deleted) ChangeSet and ChangeSetHandler
- Logic previously inside CountryBoundaryMap used exclusively for handling Relation slicing geometry has now been moved to RawAtlasRelationSlicer
- All synthetic Relation members added during multipolygon closure for slicing now use EntityIdentifierGenerator instead of picking an arbitrary Relation member to use an identifier base
-  Slicing logic has been made more consistent: now, only multipolygons will be divided into separately identified Relations ("sliced")
     - Due to Relation data differences often emerging across shards and countries, this process was previously extremely nondeterministic and sometimes led to multiple Relations (sliced and non sliced) from the same base OSM Relation, even within one country
     - Multipolygon Relations will now *always* be sliced if they have members in more than one country, even if they don't form a valid multipolygon
     - In other cases, the Relation will be tagged with the intersection of the ISO country codes being sliced and the ISO country codes of its members, as that will be the view present for the final output Atlas
           - The Relation ID will *NOT* be converted to a sliced identifier
    - During this refactor, all code has been cleaned up to pass Sonar checks!

### Potential Impact:
While the changes to Relation identifier changes sound large, the reality is that most Relations followed this identifier logic anyway. This change simply makes this behavior consistent and expected. 

The change for the synthetic added members using EntityIdentifierGenerator similarly may sound concerning, but previous behavior was nondeterministic and accordingly this should be exclusively an improvement.

### Unit Test Approach:
Significant reworks to the larger, simulated dynamic Relation slicing tests (including adding one for single country slicing) to ensure behavior is as expected. The remaining tests, which all tested multipolygon slicing behavior, were correct and left untouched.

### Test Results:
👍 

------


src/main/resources/org/openstreetmap/atlas/geography/atlas/pbf/atlas-relation-slicing.json
```diff
@@ -1,6 +1,5 @@
 {
   "filters": [
-    "natural->water,coastline",
```
matthieun: 👍

MikeGost:
This is looking very clean, awesome job!

PR #560, opened 2019-11-11
AbstractAtlasShellToolsCommand now supports ignoring unknown options

lucaspcram:
### Description:
`AbstractAtlasShellToolsCommand` now has a workaround that allows atlas commands to ignore unknown options. This will be useful when writing scripts that potentially call older versions of a command.

In order to unlock this feature, simply supply the option `--ignore-unknown-options` as the *first* element to ARGV in your command line. It must be the first element. For example, suppose the following command is failing in your script because it is running against an old version that does not contain the option `--shiny-new-option`

```
# This is failing because the installed version of some-command does not
# recognize the '--shiny-new-option'
$ atlas some-command arg1 --opt1 --shiny-new-option
```

Now, you can do this:
```
$ atlas some-command --ignore-unknown-option arg1 --opt1 --shiny-new-option
```
and the command will simply throw out the `--shiny-new-option`.

### Potential Impact:
There should be no downstream impact. This only adds additional functionality for users.

### Unit Test Approach:
See unit test added to `SimpleOptionAndArgumentParserTest`. Also, I tested from the command line.

### Test Results:
Test success.

------


src/main/java/org/openstreetmap/atlas/utilities/command/abstractcommand/AbstractAtlasShellToolsCommand.java
```diff
@@ -784,15 +783,23 @@ private void printSimpleUsageMenu()
...
     }
 
-    private String[] unpackTailSentinelArguments(final String[] args)
+    private String[] unpackSentinelArguments(final String[] args)
     {
-        // check the last arg to see if we should check for other tail arguments
-        if (args.length > 0 && JAVA_MARKER_SENTINEL.equals(args[args.length - 1]))
+        List<String> argsAsList = Arrays.asList(args);
+        argsAsList = new ArrayList<>(argsAsList);
+        if (!argsAsList.isEmpty() && IGNORE_UNKNOWN_OPTIONS.equals(argsAsList.get(0)))
```
Bentleysb: Why are we limiting this to being the first argument?

lucaspcram: We can change it so this code scans all of ARGV instead. I had it this way just to make it clear that it is a hack. Options are supposed to be handled by `SimpleOptionAndArgumentParser` but I didn't want to clutter that class with this particular workaround, as it is supposed to be for more general use.

PR #561, opened 2019-11-12
Atlas Query Language (AQL)

yazad3:
### Description:

Atlas Query Language or AQL is a new feature that allows developers to write queries directly against the Atlas files. This will allow rapid development where developers can write `select` statements against their Atlas or OSM files locally and then port these queries in their applications (spark or otherwise) to run the same queries at scale.

AQL currently supports `using` (as in use Atlas/OSM files), `select`, `update`, `delete`, explain (as in explain plan), `commit` and `diff` (as in difference between two atlas schemas).

The AQL statements (select, update, delete) support complex where clauses with geospatial and tag querying among other criterion. AQL also supports nested inner queries..

AQL leverages Atlas as the underlying framework and doesn't reinvent the wheels. AQL is efficient in the sense that queries are automatically optimized to benefit from `index`-ing which can be either atlas id based or geospatial based.

Atlas allows working against all 6 major entities, `node`, `point`, `edge`, `line`, `relation`, `area`, exposed as "tables" in an atlas schema.

### Potential Impact:

None expected including for groovy-json as it has no new transitive dependencies.

### Unit Test Approach:

173 test cases are included.

### Test Results:

Tests successful.

------


build.gradle
```diff
@@ -1,5 +1,6 @@
 plugins {
 	id 'java'
+    id 'groovy'
```
matthieun: Small indentation issue here it seems!

src/main/groovy/org/openstreetmap/atlas/geography/atlas/dsl/engine/impl/AbstractQueryExecutorImpl.groovy
```diff
@@ -0,0 +1,113 @@
...
+    private AtlasMediator toAtlasMediator(Atlas atlas) {
+        new AtlasMediator(atlas)
+    }
+
+    private final void validate(String queryAsString) {
+        Valid.notEmpty queryAsString, "Query is EMPTY!"
+
+        final List<Statement> allowedStatements = [Statement.SELECT, Statement.UPDATE, Statement.DELETE]
+
+        Valid.isTrue allowedStatements.stream().filter { statement -> queryAsString.startsWith(statement.closureName()) }.count() == 1, "Invalid Query input, only ${allowedStatements} supported."
```
matthieun: Any way to break up those two lines here so they do not overflow?

yazad3:
Hi @matthieun, addressed your items other than spotless, checkstyle. I have done some investigation in this area in the past and it turns out that a lot of tools we use for code quality for Java are encompassed in one tool named [CodeNarc](https://github.com/CodeNarc/CodeNarc). Currently, however there is some noise in the reporting due to the nature of the DSL being built and used and I will need to spend time to configure it in a way it behaves reasonably against this project. Are you ok with a new PR with CodeNarc support?

PR #568, opened 2019-11-15
Fixed a false positive bug in the Change validation

lucaspcram:
### Description:
`Change` validation was over-zealous when checking reverse `CompleteEdge` nodes. It would throw an error when one `FeatureChange` specified `null` for a start/end node, while the other had a change. We can't actually throw an error here, since it is possible that the reverse edge was freshly added while the original master edge was changed in an unrelated way (e.g. tags).

### Potential Impact:
Fewer false positive errors in downstream code.

### Unit Test Approach:
Added two unit tests to show expected behavior.

### Test Results:
Tests pass.

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeTest.java
@@ -3,8 +3,10 @@ package org.openstreetmap.atlas.geography.atlas.change;
 import org.junit.Assert;
 import org.junit.Test;
 import org.openstreetmap.atlas.exception.CoreException;
+import org.openstreetmap.atlas.geography.PolyLine;
 import org.openstreetmap.atlas.geography.Polygon;
 import org.openstreetmap.atlas.geography.atlas.complete.CompleteArea;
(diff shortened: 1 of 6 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/change/ChangeTest.java
```diff
@@ -49,6 +51,31 @@ public void testAddSameIdentifierMerge()
...
+    @Test
+    public void testEdgeNodeValidation()
+    {
+        final CompleteEdge edge1 = new CompleteEdge(1L, PolyLine.SIMPLE_POLYLINE,
+                Maps.hashMap("key1", "value1", "newKey", "newValue"), null, null, null);
+        final CompleteEdge edge1Reverse = new CompleteEdge(-1L, PolyLine.SIMPLE_POLYLINE.reversed(),
+                Maps.hashMap("key1", "value1", "newKey", "newValue"), 2L, 1L, null);
+        final Change change1 = ChangeBuilder.newInstance()
+                .addAll(FeatureChange.add(edge1), FeatureChange.add(edge1Reverse)).get();
+        Assert.assertEquals(2, change1.changeCount());
```
adahn6: 🎉

PR #570, opened 2019-11-20
DynamicAtlasExpander: Cache covered entities to save time

matthieun:
### Description:

In DynamicAtlasExpander some large polygons and polylines take some time to intersect with shards. This adds some caching to avoid recomputing those every time.

### Potential Impact:

Speedup when using DynamicAtlas

### Unit Test Approach:

Existing tests

### Test Results:

pass

------


matthieun:
@MikeGost Yes, those two are separated to avoid collisions!
There are a lot of cache hits actually, since the expansion is usually following a 2-tier exploration, even in the most restrictive cases, to be able to catch features coming from other shards which intersect the initial shard. So the good thing is it is a 50% cache hit at the minimum!

MikeGost:
Thanks @matthieun!

PR #571, opened 2019-11-20
Enhance caching in DynamicAtlasExpander

matthieun:
### Description:

In some cases, the "atlas entities to consider for expansion" filter is slow and this update follows up on #570 to have it skip already cached entities.

### Potential Impact:

Performance improvement

### Unit Test Approach:

Use existing tests

### Test Results:

Use existing tests

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasExpander.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasExpander.java
@@ -93,7 +93,7 @@ class DynamicAtlasExpander
 
     boolean areaCovered(final Area area)
     {
-        if (this.areaCoveredCache.contains(area.getIdentifier()))
+        if (!entityNotCached(area))
         {
             return true;
         }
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlasExpander.java
```diff
@@ -93,7 +93,7 @@ public DynamicAtlasPolicy getPolicy()
 
     boolean areaCovered(final Area area)
     {
-        if (this.areaCoveredCache.contains(area.getIdentifier()))
+        if (!entityNotCached(area))
```
MikeGost: Double negative --> 🤯

matthieun: I wanted to be able to use the shortcut here ;P https://github.com/osmlab/atlas/pull/571/files#diff-8b62a99e1caf05da028d568d14101c2eR453

MikeGost: Agreed, that makes that part cleaner!

PR #575, opened 2019-12-06
SmoothnessTag

Bentleysb:
### Description:

Adding the OSM smoothness tag. This tag has ordinal values, so positional comparative functions are included.

### Potential Impact:

None, just another tag class.

### Unit Test Approach:

Unit test check that the comparative functions work.

### Test Results:
None.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/tags/SmoothnessTag.java
+++ b/src/main/java/org/openstreetmap/atlas/tags/SmoothnessTag.java
@@ -1,0 +1,45 @@
+package org.openstreetmap.atlas.tags;
+
+import org.openstreetmap.atlas.tags.annotations.Tag;
+import org.openstreetmap.atlas.tags.annotations.TagKey;
+
+/**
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/tags/SmoothnessTag.java
```diff
@@ -0,0 +1,45 @@
...
+@Tag(taginfo = "https://taginfo.openstreetmap.org/keys/smoothness", osm = "https://wiki.openstreetmap.org/wiki/Key:smoothness")
+public enum SmoothnessTag
+{
+    EXCELLENT,
+    GOOD,
+    INTERMEDIATE,
+    BAD,
+    VERY_BAD,
+    HORRIBLE,
+    VERY_HORRIBLE,
```
MikeGost: Would love to see some mapillary street-level images comparing `horrible` to `very_horrible` 😂

PR #576, opened 2019-12-11
pyatlas now supports Python 3! 

lucaspcram:
### Description:
Upgraded `pyatlas` to support Python 3! Note that we no longer support Python 2. If you need a Python 2 version of `pyatlas`, you can install version `5.3.6`. But please note that all future `pyatlas` updates will be Python 3 only.

We now also use protobuf `3.11.1` as opposed to `atlas`'s `2.6.1`. This should not affect `pyatlas`'s ability to read atlases, as everything *should* be backwards compatible.

### Potential Impact:
Downstream users can continue as normal. Those waiting for a Python 3 version are satisfied.

### Unit Test Approach:
Same unit test suite as before.

### Test Results:
Tests pass. Also opened and played with some old atlas files using the new `pyatlas`. Everything seems to be compatible, and protobuf3 is working.

------


pyatlas/README.md
```diff
@@ -6,7 +6,7 @@
 To get setup in a new project folder, run:
 
     $ mkdir newproj && cd newproj
-    $ virtualenv venv --python=python2.7
+    $ virtualenv venv --python=python3
```
adahn6: 🎉

pyatlas/package.sh
```diff
@@ -85,6 +87,21 @@ do
     "$protoc_path" "$protofile" --proto_path="$protofiles_dir" --python_out="$pyatlas_root_dir/$pyatlas_srcdir/autogen" || err_shutdown "protoc invocation failed"
 done < <(find "$protofiles_dir" -type f -name "*.proto" -print0)
 
+# TODO HACK FIXME this is to fix badly generated import statements by proto3 library
```
adahn6: > TODO HACK FIXME

🤣 
but for real, thanks for adding this note here for tracking purposes!

PR #579, opened 2019-12-13
Added in tag summary map method for Change

adahn6:
### Description:

This PR adds in a set of summaries for a Change object. The first is a String that summarizes the number of changes for each ItemType and ChangeType that the Change contains. The second is a complex map of ItemType to ChangeDescriptorType to Tag key, mapping to the number of changes to that Tag for that ItemType and ChangeDescriptorType. For example, if there were 7 areas with the tag `last_edit_user_name` changed, then the map would contain an entry `(AREA, UPDATE, last_edit_user_name) -> 7`. This tag summary provides a direct way to parse and understand tag changes, which are often large in scale and hard to understand otherwise.

### Potential Impact:
None, this is additive.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/change/Change.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/change/Change.java
@@ -3,6 +3,7 @@ package org.openstreetmap.atlas.geography.atlas.change;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.Arrays;
+import java.util.EnumMap;
 import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/change/Change.java
```diff
@@ -253,6 +263,77 @@ public void save(final WritableResource resource, final boolean showDescription)
...
+                                        else
+                                        {
+                                            tagMap.get(itemType)
+                                                    .get(tagChangeDescriptor
+                                                            .getChangeDescriptorType())
+                                                    .put(tagChangeDescriptor.getKey(),
+                                                            new AtomicLong(1));
+                                        }
+                                    }));
+        }
```
lucaspcram: Wow. Haha

lucaspcram: Could we maybe just get a quick comment above this block about why it's here/what it is intended to do? That's a pretty dense cluster of stream operations haha

PR #580, opened 2019-12-19
DynamicAtlas Spatial Matcher Bug

Bentleysb:
### Description:

This fixes a bug in the DynamicAtlas spatial filter functions that take a predicate matcher.
The bug manifests when a predicate matcher is sent to one of these functions and it contains a call to `AtlasEntity.equals()`.  Currently this matcher is sent on through the equivalent spatial filter function for `this.current` (which is a PackedAtlas or MultiAtlas). The call to `AtlasEntity.equals()` then always returns false because it is comparing entities from 2 different atlases (the DynamicAtlas and the Packed/MultiAtlas). See the unit tests added in this PR for examples of this scenario. 
This bug is fixed by filtering features using the matcher after calling the spatial filter functions on `this.current`.

### Potential Impact:

No impact other than being able to safely use `AtlasEntity.equals()` in these DynamicAtlas functions. 

### Unit Test Approach:

Added unit tests for al the functions that were changed. All the tests failed before the changes. 

### Test Results:

Tested this with a version of AtlasChecks using a DynamicAtlas where this was causing an infinite loop. The loop is now fixed and completes as expected. 

------


adahn6:
this is a very good catch! thanks for the fix and tests!

PR #500, opened 2019-08-08
Fix Sonar issues

matthieun:
### Description:

Fix Sonar issues

### Potential Impact:

This should be transparent

### Unit Test Approach:

Using same unit tests, except for `indexBasedIterable` that was moved from `PackedAtlas` to `Iterables` and which now has its own test.

### Test Results:

Pass

------

In doubt: [Contributing Guidelines](CONTRIBUTING.md)

PR #581, opened 2019-12-20
Tippecanoe Bugs and Redundancy

Bentleysb:
### Description:

This fixes some bugs in the atlas to mbtiles system caused by changes in https://github.com/osmlab/atlas/pull/500, and removes some redundant information. 

The TippecanoeExporter and other commands that use the LineDelimitedGeoJsonConverter were not functioning do to a change in the thread pool used to execute the converter. This makes two updates to make this work with the new thread pool. One, ` TippecanoeCommands.concatenate(geojsonDirectory)` is moved to execute after the pool completes. Two, the conversion jobs are now properly queued by moving where the distributing of the atlas files happens. 

This also changes the way relations are listed within the properties of their members. Previously the member of a relation would contain all the properties of the parent relation. This created large amounts of redundant data. Now the members of relations only contain the atlas ids of their parent relations. The relation properties now only exist once, as the properties of the relation geojson feature. 

### Potential Impact:

Calling AtlasEntitiy.getGeoJsonProperties() will no longer return relation properties. 

### Unit Test Approach:

Updated one unit test resource to reflect the changes to how relations are added to member properties. 

### Test Results:

Tested exporting an atlas using TippecanoeExporter. This now functions as expected, and the size of the outputs are reduced due to the drop in redundant relation information. 
Tested generating AtlasChecks tippecanoe results and converting to mbtiles. This also once again works as expected. 

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/AtlasEntity.java
```diff
@@ -131,14 +132,13 @@ public JsonObject getGeoJsonProperties()
         properties.addProperty(GeoJsonUtils.ITEM_TYPE, String.valueOf(getType()));
 
         final Set<Relation> relations = relations();
-        if (relations.size() > 0)
+        if (!relations.isEmpty())
```
lucaspcram: 👍

src/test/resources/org/openstreetmap/atlas/geography/atlas/line-delimited-geojson.txt
```diff
@@ -1,18 +1,18 @@
 {"type":"Feature","geometry":{"type":"Point","coordinates":[-122.472852,37.780574]},"properties":{"identifier":1,"osmIdentifier":0,"itemType":"NODE","inEdges":[],"outEdges":[0]}}
-{"type":"Feature","geometry":{"type":"Point","coordinates":[-122.472242,37.780592]},"properties":{"identifier":2,"osmIdentifier":0,"itemType":"NODE","relations":[{"restriction":"no_left_turn","type":"restriction","identifier":1,"osmIdentifier":0,"itemType":"RELATION","relations":[{"type":"inside_because_of_relation_inside","identifier":5,"osmIdentifier":0,"itemType":"RELATION","relations":[{"type":"inside_because_of_level_2_relation_inside","identifier":4,"osmIdentifier":0,"itemType":"RELATION"}]}]}],"inEdges":[0],"outEdges":[1]}}
+{"type":"Feature","geometry":{"type":"Point","coordinates":[-122.472242,37.780592]},"properties":{"identifier":2,"osmIdentifier":0,"itemType":"NODE","relations":[1],"inEdges":[0],"outEdges":[1]}}
```
lucaspcram: This is much easier to read.

PR #586, opened 2020-01-13
Major refactor of AtlasResourceLoader

lucaspcram:
### Description:
This refactor aims to accomplish:

1. Provide a way to easily load text atlases. All `*load*` methods now fully support atlases in `.atlas.txt` format.
2. Remove the behavior from the `load` methods that returned `null`. `load` methods now throw an exception with a detailed message on failure. If users would like a method that does not throw under any circumstance, they can use the `safeLoad` variants, which return an empty `Optional` on a load fail.
3. Make recursive directory loading an explicit rather than implicit feature. Regular `load` methods are now unable to handle directory `Resource`s. If users would like to load an entire folder of atlases (using `MultiAtlas` under the hood with all the legacy file extension filtering), they should use `loadRecursively`.

A couple other ancillary changes:
1. `Resource`s can no longer have `null` names. By default, all `Resource`s will have `className@hashcode` as their name. Most `Resource` classes provide a way to override this behavior with a custom name.
2. The `FileSuffix#resourceFilter` method no longer auto-passes `Resource`s with `null` names. If a this filter acts on a `Resource` with a name that has been set to `null`, it will throw a `NullPointerException`.

### Potential Impact:
1. Users loading directories using `load` will need to change to `loadRecursively`.
2. Users expecting a `null` on `load` failure will now get a `CoreException`. They will need to either adjust their code accordingly, or switch to the `safeLoad` method and deal with the `Optional`.

@yazad3 I believe some of your groovy code uses `AtlasResourceLoader#load` with directories as arguments. You will need to change those method calls to `loadRecursively`. I can help you with this change if you have any issues.

### Unit Test Approach:
Re-wrote the entire unit test suite to obtain almost 100% coverage.

### Test Results:
All tests pass.

------


yazad3:
Thank you for tagging me @lucaspcram. I'll take a look at the changes.

lucaspcram:
No problem @yazad3 . A quick `grep` revealed only a few places where this method is used, so the changes should be minor. In fact, I believe only one place is actually using `load` to load a directory and not just an individual atlas resource. That would be the place that needs `loadRecursively`.

yazad3:
@lucaspcram - thank you for researching that.

PR #595, opened 2020-01-24
AtlasEntity isCountrySliced Method

Bentleysb:
### Description:
When working with country sliced atlases it is often necessary to control for edge effects by filtering out Atlas entities that have been trimmed at the border. This is frequently done in AtlasChecks when working with Edges and Edge networks by checking for synthetic boundary nodes. 
This seeks to make that process easier and more uniform for all Atlas entities by adding an `isCountrySliced()` method to the abstract AtlasEntity class. Each entity has its own logic for calculation this as detailed below.

isCountrySliced logic:
Area: identifier suffix is an increment of 1000
Edge: is connected to a synthetic boundary node
Line: identifier suffix is an increment of 1000
Node: has a synthetic boundary node tag
Point: has a synthetic boundary node tag
Relation: any of its members are country sliced

### Potential Impact:
Requires an extra method for classes that extend AtlasEntity.
Should make identifying country sliced entities easier.

### Unit Test Approach:

Created unit tests for each type of AtlasEntity that check sliced and non-sliced versions. 

### Test Results:
Tested using the `isCountrySliced()` method as a filter to create a sub atlas from a country sliced atlas. The resulting sub atlas contained all of the AtlasEntity types, and sampling the entities showed them to all have been country sliced.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Area.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Area.java
@@ -2,7 +2,6 @@ package org.openstreetmap.atlas.geography.atlas.items;
 
 import java.util.Map;
 import java.util.Optional;
-import java.util.regex.Pattern;
 
 import org.openstreetmap.atlas.geography.GeometricSurface;
 import org.openstreetmap.atlas.geography.Location;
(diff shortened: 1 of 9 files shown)


adahn6:
hey @Bentleysb, I like this idea a lot, but I think we may want to just make this a data member of AtlasEntity itself that defaults to no and is changed directly during country slicing. While the logic you wrote works for most AtlasEntities, there are some cases where we add synthetic Relation members for multipolygons to preserve their closed geometry along country borders-- these members will have an AtlasEntity with an ID generated using `EntityIdentifierGenerator`. In this case, these are sliced AtlasEntities, but `.isCountrySliced()` will be nondeterministic. Using a data member directly would solve this problem, and also protect against any future changes to methodology for determining if an AtlasEntity was sliced.

Bentleysb:
Closing to investigate other solutions.

PR #601, opened 2020-02-14
Improved slicing logic

adahn6:
### Description:
This PR makes two major changes to slicing logic. For Point and Line slicing, a bug has been addressed in which trivial slice geometry was not consistently being handled. In these cases, very small pieces of geometry were being included despite being "trivial" in size. While this has been fixed, the conditions for what is considered "trivial" have also been increased for polygonal geometry slice results, as the previous conditions were sometimes excluding valuable geometry.

For Relation slicing, a major change has been made to the handling of synthetic relation members. During multipolygon slicing, the slicer will add a synthetic relation member along the country borders to preserve slice the multipolygon into separate closed multipolygon relations with area contained entirely in each country. Previously, these synthetic Lines were being added by taking the intersection LineString, walking through each coordinate and either matching it to an existing Point in the atlas or rounding it to 6 digits of precision and adding a new SyntheticBoundaryNode tagged point. Unfortunately, this approach was nondeterministic as the underlying Points in a given Atlas could vary with loading options, shard location, etc. 
The new methodology resolves this issue by using only the LineString directly, with no reliance on Points in the Atlas. Additionally, before the Line is created, the slicer will attempt to match the beginning and ending location to existing outer members for the relation. This "snapping" logic is because the geometry intersection calls sometimes have slight precision issues that cause the outers to have a very small gap (and thus not be mergeable into a closed multipolygon).

Lastly, a small bug has also been fixed in which SyntheticRelationMembersAdded tag was being added to all multipolygons, even those which were not sliced (in those cases, the tag value was empty). Now, the tag will only be added if there were actually synthetic members added to the Relation.

### Potential Impact:
While the results of slicing will change, the results are not significantly different to cause serious concern.
### Unit Test Approach:
Some unit test values changed as expected. In these cases: Point values dropped, as we no longer add a Point for each coordinate in a synthetic Relation member, and several synthetic Relation member IDs changed because they are calculated with the EntityIdentifierGenerator which uses geometry values.
### Test Results:

------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -167,34 +166,26 @@ private boolean allLineTagsEqual(final List<Geometry> slices)
...
 
         // Slice the JTS Geometry
-        result = sliceGeometry(geometry, line);
+        List<Geometry> result = sliceGeometry(geometry, line);
 
         final boolean multipolygonResult = Iterables.stream(result)
                 .anyMatch(geom -> geom instanceof org.locationtech.jts.geom.Polygon
                         && ((org.locationtech.jts.geom.Polygon) geom).getNumInteriorRing() > 0);
-        if ((result == null || result.isEmpty()) && line.isClosed() || multipolygonResult)
+        if (result.isEmpty() && line.isClosed() || multipolygonResult)
```
matthieun: Just checking that the null-check removal here was intentional.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -631,8 +631,9 @@ private boolean relateCandidates(final long identifier,
             }
 
             final Geometry target = geometry;
-            List<org.locationtech.jts.geom.Polygon> candidates = boundary
-                    .query(target.getEnvelopeInternal());
+            final List<org.locationtech.jts.geom.Polygon> candidates = boundary
```
matthieun: Maybe make it a `Set` instead?

matthieun: Although I see it is ported over from below, so not required by any means.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -287,68 +293,81 @@ private boolean containsSlicedMember(final Iterable<RelationMember> members)
...
+            logger.error(
+                    "For relation {}, snapped first location by {} feet, not adding synthetic line",
+                    relationIdentifier, newLineLocations.get(0).distanceTo(originalFirst).asFeet());
+            return null;
+        }
+
+        if (newLineLocations.get(newLineLocations.size() - 1).distanceTo(originalLast)
+                .isGreaterThan(SNAP_DISTANCE))
+        {
+            logger.warn(
```
matthieun: Any reason this one is a `warn` when the other one is an `error`?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
```diff
@@ -1149,8 +1162,11 @@ private void splitRelation(final CompleteRelation currentRelation)
                     syntheticIds.add(Long.toString(member.getEntity().getIdentifier()));
                 }
             });
-            newRelation.withAddedTag(SyntheticRelationMemberAdded.KEY,
-                    SyntheticRelationMemberAdded.join(syntheticIds));
+            if (!syntheticIds.isEmpty())
```
matthieun: 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -58,7 +59,8 @@
     private static final PrecisionModel PRECISION_MODEL = new PrecisionModel(
             SEVEN_DIGIT_PRECISION_SCALE);
     protected static final PrecisionReducerCoordinateOperation PRECISION_REDUCER = new PrecisionReducerCoordinateOperation(
-            PRECISION_MODEL, false);
+            PRECISION_MODEL, true);
+    protected static final Distance SNAP_DISTANCE = Distance.inches(24);
```
matthieun: Maybe a comment here to explain to future generations why 24 inches?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
adahn6: it was! refactor included going ahead and returning an empty collection if slicing was unnecessary, rather than a mix of null or empty

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasRelationSlicer.java
adahn6: nope, oversight!

PR #616, opened 2020-03-16
Fix NullPointerException in Line Slicing

matthieun:
### Description:

When a way is made of two nodes that are at the same location (usually this is bad data) the line slicer was throwing a NPE. This PR fixes it.

Example: https://www.openstreetmap.org/way/408600053

image

### Potential Impact:

No NPE in line slicer

### Unit Test Approach:

Added one test for the case

### Test Results:

Pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
@@ -202,7 +202,9 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
         // slicing)
         if (result.isEmpty())
         {
-            if (CountryBoundaryMap.getGeometryProperty(geometry, ISOCountryTag.KEY).isEmpty())
+            final String countryName = CountryBoundaryMap.getGeometryProperty(geometry,
+                    ISOCountryTag.KEY);
+            if (countryName == null || countryName.isEmpty())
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -202,7 +202,9 @@ private boolean allLineTagsEqual(final List<Geometry> slices)
         // slicing)
         if (result.isEmpty())
         {
-            if (CountryBoundaryMap.getGeometryProperty(geometry, ISOCountryTag.KEY).isEmpty())
+            final String countryName = CountryBoundaryMap.getGeometryProperty(geometry,
+                    ISOCountryTag.KEY);
+            if (countryName == null || countryName.isEmpty())
```
adahn6: good catch!

src/test/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/LineAndPointSlicingTest.java
```diff
@@ -450,4 +450,12 @@ public void testNearestNeighborTagAssignment()
                     point.getTag(SyntheticNearestNeighborCountryCodeTag.KEY).isPresent());
         });
     }
+
+    @Test
+    public void testSingleNodeLine()
```
adahn6: thanks for replicating this in a unit test!!

PR #618, opened 2020-03-18
TurnRestrictions should be Taggable

AlexHsieh22:
### Description:
Let's make TurnRestrictions a Taggable.  The underlying relation is already stored in the TurnRestriction objects - so to make TurnRestrictions a Taggable, we can return the tags from the relation.

I've also fixed a few stylistic issues I found while reading through the file.

### Potential Impact:
None - this is new functionality not used yet.

### Unit Test Approach:
I will update unit tests to make sure TurnRestrictions can return tags from the underlying relation.

### Test Results:
I've updated the ComplexTurnRestrictionTest (where most TurnRestriction testing seems to be happening) and have added a unit test to fetch the tags and make sure the tags are present.

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/restriction/ComplexTurnRestrictionTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/restriction/ComplexTurnRestrictionTest.java
@@ -192,6 +192,23 @@ public class ComplexTurnRestrictionTest
     }
 
     @Test
+    public void testTurnRestrictionTags()
+    {
+        final Atlas testAtlas = this.rule.getAtlasNoUTurn();
+
+        final Optional<TurnRestriction> possibleTurnRestriction = TurnRestriction
+                .from(testAtlas.relation(1L));
+
+        Assert.assertTrue(possibleTurnRestriction.isPresent());
+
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/geography/atlas/items/complex/restriction/ComplexTurnRestrictionTest.java
```diff
@@ -191,6 +191,23 @@ public void testTurnRestrictionNoUTurn()
...
+        final Optional<TurnRestriction> possibleTurnRestriction = TurnRestriction
+                .from(testAtlas.relation(1L));
+
+        Assert.assertTrue(possibleTurnRestriction.isPresent());
+
+        final TurnRestriction turnRestriction = possibleTurnRestriction.get();
+
+        // Make sure both tags exist
+        Assert.assertEquals(2, turnRestriction.getTags().size());
+        Assert.assertEquals("no_left_turn", turnRestriction.getTag("restriction").get());
```
matthieun: Maybe use `TurnRestrictionTag` here instead of strings?

AlexHsieh22: Oh - great find! Yes - I will update this to use the TurnRestrictionTag statics!

PR #619, opened 2020-03-19
Minor bug fixes for line slicing and CountryBoundaryMap

adahn6:
### Description:

This PR fixes two minor issues. First, in certain extremely rare cases, the difference operation during slicing returns a heterogeneous geometry collection. This isn't something we can continue to perform slicing operations on, so we will now correctly abort slicing and log the error should this happen. It's exceptionally rare and hard to predict, so there's no unit test for this case.

Second is a very minor bug in CountryBoundaryMap.addCountry(Polygon) where the country code property was not being initialized as it was in the multipolygon version of the same method.
### Potential Impact:
None, just bug fixes.
### Unit Test Approach:
N/A
### Test Results:
N/A
------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
@@ -202,9 +202,7 @@ public class RawAtlasPointAndLineSlicer extends RawAtlasSlicer
         // slicing)
         if (result.isEmpty())
         {
-            final String countryName = CountryBoundaryMap.getGeometryProperty(geometry,
-                    ISOCountryTag.KEY);
-            if (countryName == null || countryName.isEmpty())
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasPointAndLineSlicer.java
```diff
@@ -308,10 +308,17 @@ private Geometry cutGeometry(final long identifier,
...
-                // Update target to be what's left after clipping
-                currentTarget = currentTarget.difference(candidate);
-
+                try
+                {
+                    currentTarget = currentTarget.difference(candidate);
+                }
+                catch (final IllegalArgumentException exception)
+                {
+                    logger.warn(
```
matthieun: If that usually means there is funky data in OSM, should we make that an ERROR? Or is that purely JTS?

adahn6: in this case, it's really just an extremely rare possibility of good data for this particular geometry operation, so I think WARN is appropriate.

PR #626, opened 2020-04-24
Expose underlying expander map in dynamic atlas

lucaspcram:
### Description:
Expose underlying expander `Map<Shard, Atlas>` in `DynamicAtlas`. Some users of `DynamicAtlas` may want access to this to inspect what shards have been loaded, and then to fetch those atlases if necessary.

### Potential Impact:
None, other than giving users more functionality.

### Unit Test Approach:
Added a quick unit test to make sure the getter works.

### Test Results:
It works.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
@@ -1,6 +1,8 @@
 package org.openstreetmap.atlas.geography.atlas.dynamic;
 
+import java.util.HashMap;
 import java.util.Iterator;
+import java.util.Map;
 import java.util.Map.Entry;
 import java.util.Objects;
 import java.util.Set;
@@ -186,6 +188,15 @@ public class DynamicAtlas extends BareAtlas // NOSONAR
     public DynamicAtlasPolicy getPolicy()
     {
         return this.expander.getPolicy();
+    }
+
+    /**
+     * @return A copy of the {@link Shard} to {@link Atlas} {@link Map} populated by the underlying
+     *         {@link DynamicAtlasExpander}
+     */
+    public Map<Shard, Atlas> getShardToAtlasMap()
+    {
+        return new HashMap<>(this.expander.getLoadedShards());
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java
```diff
@@ -188,6 +190,15 @@ public DynamicAtlasPolicy getPolicy()
         return this.expander.getPolicy();
     }
 
+    /**
+     * @return A copy of the {@link Shard} to {@link Atlas} {@link Map} populated by the underlying
+     *         {@link DynamicAtlasExpander}
+     */
+    public Map<Shard, Atlas> getShardToAtlasMap()
+    {
+        return new HashMap<>(this.expander.getLoadedShards());
```
matthieun, line 199: The expander will have a bunch of null Atlases here for all the shards it searched but could not find an Atlas for. Since this was internal it was fine, however exposing it this way, it would be nice to filter all the entries with null value first.

lucaspcram: Is there ever a case where users may want those `null` atlases though? For example doesn't that happen when `DynamicAtlas` tries to load a valid shard, but fails to find an atlas? Perhaps we don't want to erase that information by wiping out the null mappings. We can leave it to users to wipe the null mappings. Or maybe this is overthinking it? Your thoughts?

matthieun: So this method here has all the shards already. Would that be enough to know which ones we got and which ones we do not?
https://github.com/osmlab/atlas/blob/6.1.4/src/main/java/org/openstreetmap/atlas/geography/atlas/dynamic/DynamicAtlas.java#L195
I am really weary we would get unwanted NPEs here.

lucaspcram: Ah ok cool. As long as there is a way to get that info then we are good. I will make sure to note in the javadoc that the filtering is done for you.

PR #627, opened 2020-05-01
Refactored slicing

adahn6:
### Description:
This PR is a major refactor of the slicing logic. In summary:

* all slicing has been condensed to one class, RawAtlasSlicer, which slices all entities at the same time (previously was split into RawAtlasPointAndLineSlicer, and RawAtlasRelationSlicer)
* slicing for all entities now use a shared method for JTS intersection calls, which will downgrade the precision of returned intersection geometry appropriately to the 7 digit precision we use in Atlas
     * previously these were rounded afterward, but upon review the pre-emptive precision reduction improves geometry operation consistency 
* relation slicing now uses a new process to build synthetic members
     * the slicer first builds the relation polygons from raw atlas geometry
     * then, creates the sliced country multipolygons using the intersection of that raw multipolygon with intersecting country boundary polygons
     * then, “subtracts” out the sliced line members to determine what extra geometry needs to be added in as synthetic relation members
* SyntheticBoundaryNodes are only created for the endpoints of sliced lines that qualify as edges-   
     * previously we were adding them to all points from any sliced geometry, include Areas (which are never edges)
* nearest neighbor functionality has been removed-- it is no longer supported by the shared JTS intersection method, but also it’s no longer necessary 
* filter for “forced slicing” has been removed as it is unused and cannot be supported with the new slicing logic
* point removal is improved, further filtering unnecessary points during slicing
 
### Potential Impact:

*ANYTHING* that uses slicing will be impacted. That being said, I've reviewed the output of the new slicing logic extensively, and believe it to be exclusively improved for all use cases. 

### Unit Test Approach:
No new unit tests, a few were deleted (nearest neighbor, forced slicing), and a few were updated to use more representative assertions. Most were just API changes though.
### Test Results:
------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+        {
+            final String countryCode = slices.keySet().iterator().next();
+            this.stagedLines.get(line.getIdentifier()).withAddedTag(ISOCountryTag.KEY, countryCode);
+            if (!this.isInCountry.test(this.stagedLines.get(line.getIdentifier())))
+            {
+                removeLine(line);
+            }
+            return;
+        }
+        else if (slices.values().size() > AbstractIdentifierFactory.IDENTIFIER_SCALE_DEFAULT)
```
robertstack, line 1070: This looks like a case of very large features with many vertices and/or an intersecting country boundary with a very large number of vertices.  Maybe comment here around how often this case is expected to be triggered, how severe the error is considered to be, and whether it is indicative of bad data or just complex or highly digitized data?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+        return results;
+    }
+
+    /**
+     * Slice a Line by converting it to 1d geometry, calculating its slices, and creating the new
+     * sliced Lines. If it belongs to just one country or cannot be sliced, update the ISOCountryTag
+     * appropriately instead
+     *
+     * @param line
+     *            The Line to slice as an Area
```
robertstack, line 1155: not "as an Area" here, right?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+        {
+            final String countryCode = slices.keySet().iterator().next();
+            this.stagedLines.get(line.getIdentifier()).withAddedTag(ISOCountryTag.KEY, countryCode);
+            if (!this.isInCountry.test(this.stagedLines.get(line.getIdentifier())))
+            {
+                removeLine(line);
+            }
+            return;
+        }
+        else if (slices.values().size() > AbstractIdentifierFactory.IDENTIFIER_SCALE_DEFAULT)
```
robertstack, line 1203: Similar comment as under sliceArea() about how severe is this error.  Is this case significantly more rare than the similar error for areas, since slicing up one dimensional features should involve many fewer resulting vertices?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+        {
+            final String countryCode = entry.getKey();
+            final Set<org.locationtech.jts.geom.Polygon> slicedPolygons = new HashSet<>();
+            entry.getValue().forEach(geometry ->
+            {
+                if (geometry instanceof org.locationtech.jts.geom.Polygon)
+                {
+                    if (((org.locationtech.jts.geom.Polygon) geometry).getNumInteriorRing() > 0)
+                    {
+                        logger.error("Discarding multipolygon slicing result");
```
robertstack, line 1518: Is this expected to be common?  If not, should there be more description of the geometry to troubleshoot?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+
+    /**
+     * Slice a Line by converting it to 1d geometry, calculating its slices, and creating the new
+     * sliced Lines. If it belongs to just one country or cannot be sliced, update the ISOCountryTag
+     * appropriately instead
+     *
+     * @param line
+     *            The Line to slice as an Area
+     */
+    private void sliceLine(final Line line)
```
robertstack, line 1157: This looks similar to sliceArea(), just with lines vs areas; would it benefit from a helper method / class to cover the identical processing?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
+                polylineForSlice = polylineForSlice.reversed();
+            }
         }
+        return polylineForSlice;
+    }
 
-        // Assume it's inside the bound
-        return false;
+    /**
+     * Given a multipolygon Relation, purge any members that don't meet the criteria in the OSM
```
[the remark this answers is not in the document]
adahn6: Ah, I mentioned this in the README but I'll put the link here too!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: yeah this is a good catch! it originally was one method, but the reality is that I found the forking logic within one method harder to read (and it was impossible to keep at sonar's standards for complexity, too). this approach does lead to some duplication, but long term I think that's easier to maintain.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: I don't believe these are common cases for either type of object, but I'll do a bit of inspecting to confirm that!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: This case is very rare, but it's important to discard. I tried to design a unit case around this but struggled to get it to replicate properly, basically it happens when a closed Line that will become an Area has an inner hole introduced by a boundary. In that case, our Area representation of a simple Polygon with no holes won't work, and we're forced to either discard the slicing result or include it without the hole (which seems much worse!).

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/README.md
```diff
@@ -1,79 +1,64 @@
...
+3. Slice all `Point` entities
+4. Filter any remaining `Relation` entities-- because we don't expand on all `Relations`, it's impossible to deterministically "slice" these `Relations`, so instead we just filter out any members that are outside the country code set being sliced and update the country tag for the `Relation` to be the sum of its remaining members.
+5. Add all `CompleteEntities` in the staged entity maps as `FeatureChange.ADD`-- these are either existing features being updated or new entities being added, so `FeatureChange.ADD` is always appropriate
+6. Build a new `ChangeAtlas` out of these changes, then cut out any entities that lay outside the shard bounds (frequently happens for data loaded in durin the `DynamicAtlas`expansion for `Relations`)
+
+### Geometry Slicing
+Slicing of all entities follows the same general approach. 
+1. The data for the entity is constructed into a relevant JTS geometry. 
+    * For example, for a closed `Line` that meets the criteria for an `Area`, a JTS `Polygon` is created
+2. This internal envelople for this geometry is checked against the spatial index of the boundary map to calculate which country boundary polygons it intersects
```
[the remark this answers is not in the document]
adahn6: lmao wow I really can't type 🤣

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: good catch!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: I actually can't find any confirmed cases of this at the moment, but I have seen it in the past-- clearly a rare case, but something we should still check for in the code

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1,258 +1,767 @@
...
-        this.shardOrAtlasName = getShardOrAtlasName(startingAtlas);
+        final Optional<String> countryCodeTag = newRelation.getTag(ISOCountryTag.KEY);
+        if (countryCodeTag.isEmpty())
+        {
+            throw new CoreException("Could not find country code for split relation {}",
+                    newRelation.getIdentifier());
+        }
+        final String countryCode = countryCodeTag.get();
+        final Set<String> synetheticIdsAdded = new HashSet<>();
+        for (int i = 0; i < remainder.getNumGeometries(); i++)
```
matthieun, line 405: One of the rare places where Checkstyle allows less than 3 characters in a variable!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1,258 +1,767 @@
...
+        final Predicate<AtlasEntity> expandPredicate = entity -> entity.getType()
+                .equals(ItemType.LINE)
+                && entity.relations().stream()
+                        .anyMatch(relation -> this.relationPredicate.test(relation)
+                                && initialAtlas.relation(relation.getIdentifier()) != null);
+        final Time dynamicAtlasTime = Time.now();
+        logger.info(DYNAMIC_ATLAS_CREATION_TASK, initialShard.getName());
+        final DynamicAtlasPolicy policy = new DynamicAtlasPolicy(atlasFetcher, sharding,
+                initialShard, Rectangle.MAXIMUM).withDeferredLoading(true)
+                        .withExtendIndefinitely(true)
```
matthieun, line 448: Is it expected here to extend indefinitely? Especially if the bounds are MAXIMUM?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1,258 +1,767 @@
...
+                            });
+                }
+                continue;
+            }
+            final Optional<String> countryCodeTag = stagedRelationMember.getTag(ISOCountryTag.KEY);
+            if (countryCodeTag.isEmpty())
+            {
+                throw new CoreException("Untagged country value for entity {} for relation {}",
+                        stagedRelationMember, relation.getIdentifier());
+
```
matthieun, line 744: whitespace

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -320,44 +875,656 @@ protected boolean isInsideWorkingBound(final AtlasEntity entity)
...
-        // Assume it's inside the bound
-        return false;
+    /**
+     * Given a multipolygon Relation, purge any members that don't meet the criteria in the OSM
+     * specification
+     *
+     * @param relation
+     *            The Relation to check
+     */
+    private void purgeInvalidMultiPolygonMembers(final CompleteRelation relation)
```
matthieun, line 958: More of a general question. What is the thinking behind removing bad member roles? Wouldn't that blind any future AtlasCheck for that specific error?
My initial gut feel is that even if some members are not inner nor outer, they could be ignored when slicing the relation. However if that makes the code significantly more complicated, I am all for keeping this as is.

matthieun:
Also, if you think the 3 cognitive complexity code smells are un-warranted, you can `// NOSONAR` them...
also there seems to be duplication here that seems easy to remove?
https://sonarcloud.io/code?id=org.openstreetmap.atlas%3Aatlas&pullRequest=627&selected=org.openstreetmap.atlas%3Aatlas%3Asrc%2Fmain%2Fjava%2Forg%2Fopenstreetmap%2Fatlas%2Fgeography%2Fatlas%2Fraw%2Fslicing%2FRawAtlasSlicer.java&line=486
image

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: yep. worked on this for a while, this is necessary.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: honestly, the justification for this is twofold. first off, we're going to ignore geometry from any of these anyway. but second, and more importantly, if we do this, we're guaranteed the important condition that this relation is *lowest order* which for a raw atlas very nearly means *line members only*. without that condition, we have handle the case that we've already sliced a child relation member, which gets into nested misery of changes for splitting this and finding which child split relation maps to this relation's new split relation. it's awful and not even conformant to the specification, so we just don't support it. 

there's nothing really stopping us from including the points or lines without valid roles, but if we're going so far as to filter these down to lowest order because of this, I feel like we should just go all out and make the relation fully conform to expectations.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
robertstack: That seems like a reasonable outcome.  The open question is more -- can the geometry or at least some location of the feature get logged to help rare case future troubleshooting?  Most other error logs here give more hints about the the geometry or feature in trouble.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
robertstack: Interesting, so it's a Sonar complexity versus duplication tradeoff.  I trust your maintenance assessment here.

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: in light of this discussion, I've changed the behavior-- now if this happens, we throw an exception, catch it upstream a bit, and slice the feature linearly instead of polygonally. this should resolve any concerns about bad geometry for this kind of issue.

robertstack: Oh nice!  Thank you @adahn6

PR #630, opened 2020-05-18
Fix slicing bugs, add new synthetic tests, and overhaul slicing tests

adahn6:
### Description:

This PR addresses a number of requests from the slicing changes in #627 . First, the AtlasLoadingOption has been changed to only allow a single country instead of a set. The multiple countries setting was previously only used in tests, and is now removed.
Next, geometry calculations are now done without SnapOverlayOp, as in this has been determined unnecessary and in rare cases introduced invalid computational results that affected the quality of the output.
Additionally, new synthetic tags have been added to better track certain behaviors during slicing. For example, SyntheticInvalidMultiPolygonRelationMembersRemovedTag will add a tag to any multipolygon relations that are sliced and have had invalid members removed, with comma separated identifiers for the removed members.
Lastly, testing for slicing has been consolidated to a single test class and rule, with assertion quality dramatically improved and several tests added or reworked.

### Potential Impact:
Limited overall impact as changes here are mostly non-breaking. AtlasLoadingOption API for setting countries has changed, though, and will need to be updated in other projects that use this functionality.
### Unit Test Approach:
A lot-- nearly every slicing test's output has been examined and assertions have been reformulated for them. Nearly all assertions compute the expected values based on the data in the atlas, though a few values are still hardcoded. Importantly, this means that tag values for tags like SyntheticRelationMembersAdded are computed, not hardcoded, which should increase the robustness of the test. Several redundant tests were removed, and at least one new test was added.
### Test Results:

------


adahn6:
oh, and two of the three //nosonar comments in RawAtlasSlicer have been removed :)

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1483,15 +1403,16 @@ private void sliceMultiPolygonRelation(final CompleteRelation relation) // NOSON
         final Map<String, CompleteRelation> newRelations = new HashMap<>();
         clippedMultiPolygons.keySet().forEach(countryCode ->
         {
-            final long newRelationId = relationIdentifierFactory.nextIdentifier();
-            newRelationIds.add(newRelationId);
-            if (this.getCountries().contains(countryCode))
+            final CompleteRelation newRelation = CompleteRelation.shallowFrom(relation)
+                    .withIdentifier(relationIdentifierFactory.nextIdentifier())
+                    .withTags(relation.getTags())
+                    .withAddedTag(SyntheticGeometrySlicedTag.KEY,
```
matthieun, line 1409: Should there be a toString here too?

src/main/java/org/openstreetmap/atlas/tags/SyntheticGeometrySlicedTag.java
```diff
@@ -0,0 +1,27 @@
...
+ * @author samg
+ */
+
+@Tag(synthetic = true)
+public enum SyntheticGeometrySlicedTag
+{
+    YES;
+
+    @TagKey
+    public static final String KEY = "synthetic_geometry_sliced";
```
matthieun, line 21: 👍

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: no-- the key is a private member that is already a string, the YES is the enum value so `toString()` has to be explicitly invoked

adahn6:
not moving the @TAG_KEY to an enum, that seems weird, so this is ready for merge

src/main/java/org/openstreetmap/atlas/tags/SyntheticSyntheticRelationMemberTag.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Tag to indicate an entity is a synthetic relation member added during country slicing.
+ * <p>
+ * This is not an OSM tag.
+ *
+ * @author samg
+ */
+@Tag(synthetic = true)
+public enum SyntheticSyntheticRelationMemberTag
```
MikeGost, line 15: Nit: Synthetic repeated twice.

adahn6: I waffled on this, the format seems to be `Synthetic[description]Tag` and the description in this case is `SyntheticRelationMember`-- i.e., it's a Synthetic Tag tracking Synthetic Relation Members. I went with this, but if there's strong feelings towards the other name I'm happy to do that.

PR #636, opened 2020-05-27
Refactored File to be file system agnostic

lucaspcram:
### Description:

The old `File` implementation was directly tied to the default `FileSystem`. `File` is now completely file system agnostic. This means that all code which uses files is now inherently much more unit testable. We no longer have to create tons of temp files at the system default location, which was often error prone especially on unknown testing environments. See the new unit tests for `File` which make use of [jimfs](https://github.com/google/jimfs), a nice in-memory `FileSystem` implementation.

This PR also deprecates some old methods which explicitly failed to specify a `FileSystem`. The hope is that downstream users will be encouraged to write more modular code, code that does not make any assumptions about the underlying `FileSystem`. Ideally, code that deals with files should work with `Path`s and `File`s sourced from the `FileSystem` of the caller's choosing. That way, all code that utilizes this new `File` class will be unit testable with a drop-in file system.

I have added a README to the `org.openstreetmap.atlas.streaming.resource` package with details about this refactor, as well as a simple guide to refactoring code that uses now-deprecated `File` methods.

Please note three method names have changed, but their functionality remains otherwise intact:
1. `getAbsolutePath()` -> `getAbsolutePathString()`
2. `getParent()` -> `getParentPathString()`
3. `getPath()` -> `getPathString()`

### Potential Impact:
Some `File` methods are now deprecated.

### Unit Test Approach:
Added a big unit test class for `File`, previously this core library class had no tests!

### Test Results:
Tests look good! Coverage is pretty solid. There are still some nooks and crannies uncovered (mostly just exceptions that are hard to reproduce, as well as some of the deprecated code).

------


src/main/java/org/openstreetmap/atlas/streaming/resource/File.java
```diff
@@ -2,155 +2,485 @@
...
+
+    /**
+     * Create a new {@link File} from a given path string, using the default {@link FileSystem}
+     * (i.e. {@link FileSystems#getDefault()}) to resolve the path string into an actual
+     * {@link Path}. Automatically create all necessary parent directories.
+     *
+     * @param pathString
+     *            the path string to the file
+     */
+    @Deprecated
```
matthieun, line 370: Can you add a deprecated comment above, with a description on how to migrate any code using this method to the new model?

src/main/java/org/openstreetmap/atlas/streaming/resource/TemporaryFile.java
```diff
@@ -1,25 +1,55 @@
...
+import java.nio.file.Path;
 
 /**
+ * A special type of {@link File} that automatically cleans up after itself. Please use
+ * {@link File#temporary(FileSystem)}, {@link File#temporaryFolder(FileSystem)}, and the related
+ * methods to obtain {@link TemporaryFile}s. You may take advantage of the automatic cleanup by
+ * using it in a try-with-resources like so:
+ *
+ * <pre>
+ * try (TemporaryFile directory = File.temporaryFolder(fileSystem))
```
matthieun, line 14: Nice example!

src/test/java/org/openstreetmap/atlas/streaming/resource/FileTest.java
```diff
@@ -0,0 +1,600 @@
...
+ */
+public class FileTest
+{
+    @Rule
+    public final ExpectedException expectedException = ExpectedException.none();
+
+    @Test
+    public void testChild()
+    {
+        try (FileSystem filesystem = Jimfs.newFileSystem(Configuration.osX()))
```
matthieun, line 35: Not familiar with Jimfs... Does that mean this test wont work or will work differently on other platforms?

src/test/java/org/openstreetmap/atlas/streaming/resource/FileTest.java
```diff
@@ -0,0 +1,600 @@
...
+import org.openstreetmap.atlas.streaming.compression.Decompressor;
+import org.openstreetmap.atlas.utilities.collections.Sets;
+
+import com.google.common.jimfs.Configuration;
+import com.google.common.jimfs.Jimfs;
+
+/**
+ * @author lcram
+ */
+public class FileTest
```
matthieun, line 27: Great examples here, will help propagate better testing to other projects!

src/test/java/org/openstreetmap/atlas/streaming/resource/FileTest.java
lucaspcram: No - it only affects the way the filesystem behaves under the various file operations. See the Javadoc comment [here](https://github.com/google/jimfs/blob/c934a4b71f8bc396107e40cfc7bfad7c1d45359e/jimfs/src/main/java/com/google/common/jimfs/Configuration.java#L56) as well as the explanations for the various configurations - [Unix configuration](https://github.com/google/jimfs/blob/c934a4b71f8bc396107e40cfc7bfad7c1d45359e/jimfs/src/main/java/com/google/common/jimfs/Configuration.java#L84), [osX configuration](https://github.com/google/jimfs/blob/c934a4b71f8bc396107e40cfc7bfad7c1d45359e/jimfs/src/main/java/com/google/common/jimfs/Configuration.java#L132), etc.

lucaspcram: I just picked `osX` for no particular reason. Could have just as well used the `unix` configuration for the purposes of these unit tests.

src/main/java/org/openstreetmap/atlas/streaming/resource/File.java
lucaspcram: Done.

PR #642, opened 2020-06-17
Bumped log in CountryBoundaryMap down to debug

lucaspcram:
### Description:
This log message does not need to be `ERROR` level.

### Potential Impact:
Less log noise.

### Unit Test Approach:
N/A

### Test Results:
N/A

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
@@ -363,7 +363,7 @@ public class CountryBoundaryMap implements Serializable, GeoJson
                 // Trying to override an existing value - this shouldn't happen!
                 if (!Objects.equals(existingValue, value))
                 {
-                    logger.error(
+                    logger.debug(
                             "Trying to override existing '{}' key's value of '{}' with '{}' for geometry {}",
                             key, existingValue, value, geometry.toString());
                 }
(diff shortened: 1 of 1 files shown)


src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -363,7 +363,7 @@ public static void setGeometryProperty(final Geometry geometry, final String key
                 // Trying to override an existing value - this shouldn't happen!
                 if (!Objects.equals(existingValue, value))
```
matthieun, line 364: How about this, so the geometry is not computed? Or is it going too far?
```suggestion
                if (!Objects.equals(existingValue, value) && logger.isDebugEnabled())
```

src/main/java/org/openstreetmap/atlas/geography/boundary/CountryBoundaryMap.java
```diff
@@ -363,7 +363,7 @@ public static void setGeometryProperty(final Geometry geometry, final String key
...
                 {
-                    logger.error(
+                    logger.debug(
                             "Trying to override existing '{}' key's value of '{}' with '{}' for geometry {}",
                             key, existingValue, value, geometry.toString());
```
matthieun, line 368: Or other option is this, since logger will do the toString anyway:
```suggestion
                            key, existingValue, value, geometry);
```

PR #644, opened 2020-06-25
Fixes for slicing certain OSM valid polygons that are OGC invalid

adahn6:
### Description:

This PR adds in support for slicing certain multipolygon Relations that are OSM valid but not OGC valid. Specifically, the OSM multipolygon standard explicitly allows [overlapping inner rings](https://wiki.openstreetmap.org/wiki/Relation:multipolygon#Touching_inner_rings) as long as those inners have differing tags. However, the OGC standard for polygons does _not_ allow this: see [this StackOverflow post](https://gis.stackexchange.com/questions/301619/can-ogc-simple-feature-polygons-interior-boundaries-share-an-edge) for a good explanation.

Attempting to slice multipolygons that are not OGC valid frequently (possibly always?) results in topology exceptions, which means these features cannot be sliced even though they are OSM valid. The proposed fix here is for a specific subset of these multipolygons. The algorithm works like this:
*Build the multipolygon out of raw OSM data and check if it's valid
*IF it isn't, check if there are overlapping inners that *are not sliced* (i.e. don't cross any boundaries), and if there are, remove them from the multipolygon
*Check if the resulting multipolygon is now valid-- fail out if it isn't, but continue if it is
*When creating the per-country sliced relations, add the overlapping inners back in.

This algorithm works because any inners that aren't sliced won't affect the overall results of the slicing operation-- they're entirely contained within the area of the multipolygon that is contained within a country already, so they can temporarily be removed and then re-added. This removal allows the slicing operation to continue without a topology failure, while still allowing the end result to be a sensible slicing of the entire OSM relation data.

In addition, a new synthetic tag has been added, SyntheticInvalidGeometryTag-- this tracks any OSM data found during slicing that is not OGC valid. This tag will primarily be applied to entities that are candidates for slicing and should *not* be relied upon for any other usage; it simply tracks which entities were attempted to be sliced but could not do to OGC noncompliance.

### Potential Impact:
None in particular, better support for slicing certain multipolygon Relations.
### Unit Test Approach:

Two unit tests for the relevant types of multipolygons with overlapping inners.

### Test Results:
------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1026,6 +1028,68 @@ private void removeLine(final Line line)
...
+                for (int k = j + 1; k < currentPolygon.getNumInteriorRing(); k++)
+                {
+                    final Geometry comparisonInner = currentPolygon.getInteriorRingN(k);
+                    if (currentInner.intersects(comparisonInner))
+                    {
+                        // okay, it might be a candidate for removal
+                        remove = true;
+
+                        // but now validate that it's not composed of sliced lines
+                        for (final Line innerLine : innerLines)
```
matthieun, line 1060: 4-deep for loop!

adahn6: 😢

PR #646, opened 2020-06-29
Added reverse functionality to Angle

adahn6:
### Description:

This PR adds in a `reverse()`method to the Angle class to enable an API for obtaining the reverse of a given Angle.

### Potential Impact:

None, as it's a new API.

### Unit Test Approach:

Unit test added and passing.

### Test Results:

N/A

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/utilities/scalars/AngleTest.java
+++ b/src/test/java/org/openstreetmap/atlas/utilities/scalars/AngleTest.java
@@ -16,7 +16,7 @@ public class AngleTest
     private static final Logger logger = LoggerFactory.getLogger(AngleTest.class);
 
     @Test
-    public void testAddSubstract()
+    public void testAddSubtract()
     {
         logger.info("Small: " + Angle.dm7(2320));
         final Angle ninety = Angle.degrees(90);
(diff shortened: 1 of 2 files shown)


src/test/java/org/openstreetmap/atlas/utilities/scalars/AngleTest.java
```diff
@@ -16,7 +16,7 @@
     private static final Logger logger = LoggerFactory.getLogger(AngleTest.class);
 
     @Test
-    public void testAddSubstract()
```
lucaspcram, line 19: Nice catch.

src/test/java/org/openstreetmap/atlas/utilities/scalars/AngleTest.java
```diff
@@ -153,6 +153,16 @@ public void testPositiveRadians()
...
     }
 
+    @Test
+    public void testReverse()
+    {
+        Assert.assertEquals(Angle.degrees(0), Angle.MINIMUM.reverse());
+        Assert.assertEquals(Angle.degrees((double) -1 / Angle.DM7_PER_DEGREE),
+                Angle.MAXIMUM.reverse());
+        Assert.assertEquals(Angle.degrees(90), Angle.degrees(-90).reverse());
+        Assert.assertEquals(Angle.degrees(-90), Angle.degrees(90).reverse());
```
matthieun, line 163: Can you also add test cases where there is some modulo happening? Like
`Assert.assertEquals(Angle.degrees(90), Angle.degrees(-450).reverse());`

PR #647, opened 2020-07-02
Lazy Cache id to FC map in Change

matthieun:
### Description:

For large change objects, it can be a drain to compute the map of AtlasEntityKey to FeatureChange. To alleviate this concern, the map is now cached, so it is processed only once.

### Potential Impact:

Performance improvement

### Unit Test Approach:

Added simple unit test

### Test Results:

Pass

------


jwpgage:
LGTM! Thanks for the new test

PR #651, opened 2020-07-16
StringToPredicateConverter and ConfiguredFilter now support speedups

lucaspcram:
### Description:
The two classes mentioned in the title now support some optimizations for faster predicates. These optimizations come at the cost of decreased code security. They should only be used in very limited cases.

### Potential Impact:
Performance boost in limited cases.

### Unit Test Approach:
Added test class for ConfiguredFilter, also added more StringToPredicate tests.

### Test Results:
👍

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/conversion/StringToPredicateConverter.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/conversion/StringToPredicateConverter.java
@@ -81,10 +81,10 @@ public class StringToPredicateConverter<T> implements Converter<String, Predicat
     /**
      * Similar to {@link StringToPredicateConverter#convert(String)}, but uses some hacks to improve
      * runtime in certain cases. This version is EXTREMELY INSECURE and should never be used. Please
-     * use {@link StringToPredicateConverter#convert(String)} unless you are extremely sure you want
-     * unsafe conversion and understand what security features you are sacrificing. Note also that
-     * due to the limitations of the improvement hacks, this conversion only supports single
-     * statement expressions.
+     * use {@link StringToPredicateConverter#convert(String)} unless you are sure you want unsafe
+     * conversion and understand what security features you are sacrificing. Note also that due to
+     * the limitations of the improvement hacks, this conversion only supports single statement
+     * expressions.
      *
      * @param booleanExpressionString
      *            a boolean expression involving the object 'e' under consideration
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/utilities/conversion/StringToPredicateConverter.java
```diff
@@ -108,10 +140,67 @@ public StringToPredicateConverter()
...
+     * @return the updated converter
+     */
+    public StringToPredicateConverter<T> withClearedStarImportPackages()
+    {
+        this.additionalWhitelistPackages.clear();
+        return this;
+    }
+
+    @SuppressWarnings("unchecked")
+    private Class<Script> checkExpressionSafety(final String booleanExpressionString)
```
matthieun, line 163: Nice to break this method out!

src/test/java/org/openstreetmap/atlas/utilities/conversion/StringToPredicateConverterTest.java
```diff
@@ -52,4 +118,84 @@ public void testConversion()
...
+
+        Assert.assertTrue(predicate5.test(5));
+        Assert.assertTrue(predicate6.test(5));
+    }
+
+    @Test
+    public void testImportInjectionProtection1()
+    {
+        this.expectedException.expect(CoreException.class);
+        this.expectedException.expectMessage("Invalid import");
```
matthieun, line 161: 👍

src/test/java/org/openstreetmap/atlas/utilities/conversion/StringToPredicateConverterTest.java
```diff
@@ -52,4 +118,84 @@ public void testConversion()
...
+    public void testImportInjectionProtection2()
+    {
+        this.expectedException.expect(CoreException.class);
+        this.expectedException.expectMessage("Invalid import");
+        new StringToPredicateConverter<Integer>()
+                .withAddedStarImportPackages(Collections.singletonList("System.out.println()"))
+                .convertUnsafe("e.intValue() == RandomTagsSupplier.randomTags(5).size()");
+    }
+
+    private void speedTest()
```
matthieun, line 178: Maybe invoke this one in a main method so it can easily be run locally?

lucaspcram: Whoops I actually meant to remove this! Do you think it's something we should leave in? Otherwise I'll just remove it.

matthieun: Sure, removing it is fine if you're ok with that!

PR #659, opened 2020-07-31
Made getTTYStringBuilder public and added toJson to ConfiguredFilter

lucaspcram:
### Description:
Made the `getTTYStringBuilderX` methods in `AbstractAtlasShellToolsCommand` public so that other classes can easily get configured string builders. Also added a `toJson` method to `ConfiguredFilter`.

### Potential Impact:
More methods available to atlas users.

### Unit Test Approach:
Unit tests included for the toJson method!

### Test Results:
👍

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/utilities/configuration/ConfiguredFilter.java
+++ b/src/main/java/org/openstreetmap/atlas/utilities/configuration/ConfiguredFilter.java
@@ -19,6 +19,9 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.google.common.collect.Lists;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonPrimitive;
 
 /**
  * This class reads in a configuration file with a specific schema and creates filters based on the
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/utilities/configuration/ConfiguredFilter.java
```diff
@@ -27,6 +30,13 @@
     public static final String DEFAULT = "default";
     public static final ConfiguredFilter NO_FILTER = new ConfiguredFilter();
     public static final String CONFIGURATION_ROOT = CONFIGURATION_GLOBAL + ".filters";
+    public static final String TYPE_JSON_PROPERTY_VALUE = "_filter";
+    public static final String NAME_JSON_PROPERTY = "name";
+    public static final String PREDICATE_JSON_PROPERTY = "predicate";
+    public static final String UNSAFE_PREDICATE_JSON_PROPERTY = "unsafePredicate";
```
matthieun, line 36: How about re-using those values in the private static values below?

lucaspcram: Haha I knew you would mention this. We can do that for now I suppose. I'd like the values to be public in case future code wants to parse the JSON and wants to know the key names. So maybe I can change those to public? Also, I still need to add unit tests...

matthieun:
Oh and also it'd be nice to have a unit test on the json function!

lucaspcram:
@matthieun see my reply

src/main/java/org/openstreetmap/atlas/utilities/configuration/ConfiguredFilter.java
matthieun: Yes, making all those public seems fine to me!

PR #661, opened 2020-08-03
Minor fix for fullyGeometricallyEncloses

adahn6:
### Description:

This PR modifies `Polygon.fullyGeometricallyEncloses(Rectangle)` to ensure that JTS is used should the AWT implementation return `false`. The change is motivated by the AWT implementation's behavior of returning false if it detects the calculation is too costly; in this case, falling back to JTS results in a slower calculation but ensures that the result is always accurate.

### Potential Impact:

Should be limited-- could impact performance somewhat for some users, but will guarantee more accurate results. 

### Unit Test Approach:
Added a unit test to ensure that a previously failing rectangle contained within a polygon now works properly.

### Test Results:

------


src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -255,17 +255,15 @@ public boolean fullyGeometricallyEncloses(final Rectangle rectangle)
             return false;
         }
         // The item is within the bounds of this Polygon
-        // if this value overflows, use JTS to correctly calculate covers
-        if (awtOverflows())
+        // if this value overflows, or if the much faster AWT returns false,
+        // use JTS to correctly calculate covers
+        if (awtOverflows() || !awtArea().contains(rectangle.asAwtRectangle()))
```
matthieun, line 260: Here that would fall back to JTS for all cases where the area does not contain the rectangle, even when the area is half a world away from the rectangle...
I would fall back only if the area's bounds overlap with the rectangle, and the awt contains is false.

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
```diff
@@ -254,18 +254,23 @@ public boolean fullyGeometricallyEncloses(final Rectangle rectangle)
...
         {
             final org.locationtech.jts.geom.Polygon polygon = JTS_POLYGON_CONVERTER.convert(this);
             return polygon.covers(JTS_POLYGON_CONVERTER.convert(rectangle));
         }
-        // for most cases use the faster awt covers
-        else
+
+        // If AWT contains is false, then we want to recheck with JTS as it's less performant
+        // but more accurate-- AWT contains returns false if the calculation is too expensive
+        if (!awtArea().contains(rectangle.asAwtRectangle()))
```
matthieun, line 267: It seems this does the same thing, i.e. check JTS even if an area is really far from the Rectangle, no?

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
adahn6: wait maybe I'm confused, doesn't that happen on line 252 already?

src/main/java/org/openstreetmap/atlas/geography/Polygon.java
matthieun: Aaah yes, it is already checked before!

PR #662, opened 2020-08-04
Use ConfiguredFilter in AtlasLoadingOption

matthieun:
### Description:

Update `AtlasLoadingOption` to take advantage of `ConfiguredFilter` when possible (i.e. for non-pbf entities). The `ConfiguredTaggableFilter` is still an option for backwards compatibility, and so that everything remains `Serializable`, this pull request introduces a `BridgeConfiguredFilter` which can distinguish what input is passed in.

Also:
- Updated `ConfiguredFilter` to allow for configuration root definitions other than `global.filters`. 

### Potential Impact:

More options for filtering ways that are way sectioned, what relations are sliced, etc. It is still backwards compatible, so it should not have a negative impact.

### Unit Test Approach:

Added new unit tests for the `BridgeConfiguredFilter` and updated ones for `OsmPbfIngest` and `ConfiguredFilter`

### Test Results:

Pass

------


src/main/java/org/openstreetmap/atlas/geography/atlas/pbf/AtlasLoadingOption.java
```diff
@@ -17,20 +18,31 @@
 {
     private static final long serialVersionUID = 1811691207451027561L;
 
-    private static final ConfiguredTaggableFilter DEFAULT_EDGE_FILTER = new ConfiguredTaggableFilter(
-            new StandardConfiguration(new InputStreamResource(
-                    () -> AtlasLoadingOption.class.getResourceAsStream("atlas-edge.json"))));
-    private static final ConfiguredTaggableFilter DEFAULT_AREA_FILTER = new ConfiguredTaggableFilter(
-            new StandardConfiguration(new InputStreamResource(
-                    () -> AtlasLoadingOption.class.getResourceAsStream("atlas-area.json"))));
+    public static final String ATLAS_EDGE_FILTER_NAME = "atlas-edge";
```
adahn6, line 21: making these name strings here is a good idea! 👍

adahn6:
this looks really good! 🚀

PR #664, opened 2020-08-05
Update FeatureChange hashCode method

seancoulter:
### Description:

Currently the hashCode method for FeatureChange considers 2 things: the changeType (ADD, REMOVE) of the FC, and the hashcode of the afterView of the entity being changed, which is a hash of its Atlas identifier and the feature class. This seems to conform to the general hashCode contract, but it presents some performance issues due to hash collisions in hash-based data structures (HashSet, HashMap etc)

Example potential collision: consider a FeatureChange containing a Relation with ID=1 which is being ADDed to the Atlas via a new relation member X. Also consider a different FeatureChange with this same relation, except it's being ADDed to the atlas via some different new member Y. These 2 relations produce the same hash value despite having different members being added. Some data structure relying on the hash of these FeatureChanges would nick performance by putting both of these FCs in the same hash bucket.

This PR updates the FeatureChange hashCode() method to do the following: if the afterview of the FC is a...
-Relation: include its Relation members in the hash
-Node: include its inEdges and outEdges in the hash
-All else: include the object's tags in the hash

Generally, this PR aims to satisfy the 2nd sentence under the 3rd bullet point of the general contract for hashCode: https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode%28%29

### Potential Impact:

Improved performance for hash-based data structures that store FeatureChanges

### Unit Test Approach:

New unit test confirming that two separate FeatureChanges operating on the same Relation but with different afterViews (with respect to members) produce different hash codes

### Test Results:

Pass

------


src/main/java/org/openstreetmap/atlas/geography/atlas/change/FeatureChange.java
```diff
@@ -403,7 +403,8 @@ public ItemType getItemType()
     @Override
     public int hashCode()
     {
-        return Objects.hash(this.changeType, this.afterView);
+        return Objects.hash(this.changeType, this.afterView,
+                this.afterView instanceof Relation ? ((Relation) this.afterView).members() : null);
```
matthieun, line 407: It is a great find, however this would also happen for node's in/out edge lists, no? It would be nice to account for that case here too, if possible!

seancoulter: Yes good call. Also included a catch-all for any other Atlas Entity afterView that hashes that entity's tags. Hopefully this will help collisions that could result from hashing 2 FC ADDS on an entity with different tag updates.

This is honestly a pretty arbitrary way to do it since the ADD could be a geometry change, or in the case of Nodes, a tag change, and so on. But performance already seems to have improved in cases where we have large hash collections, so we might get away with just picking some property and sticking with it !

PR #665, opened 2020-08-07
Fix sonar vulnerability in Extractor

matthieun:
### Description:

Fix this vulnerability: https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AXO3-pDGXm9neesF5xit&open=AXO3-pDGXm9neesF5xit

### Potential Impact:

Pass quality gate

### Unit Test Approach:

Leave existing tests coverage. The update did not touch the functionality of the methods.

### Test Results:

No test results.

------


src/main/java/org/openstreetmap/atlas/utilities/archive/Extractor.java
```diff
@@ -197,7 +195,7 @@ else if (outputFile.exists() && this.skipExisting)
                 }
                 else if (current.isDirectory())
                 {
-                    continue;
+                    // Continue
```
adahn6, line 198: should this be commented or just removed?

matthieun: Well, sonar complained about using continue here since it does not change the behavior. But when I removed it it complained about an empty method body. So here we go :)

src/main/java/org/openstreetmap/atlas/utilities/archive/Extractor.java
```diff
@@ -187,6 +179,12 @@ public Extractor extract(final File inputFile) throws ArchiveException, IOExcept
             {
                 final File outputFile = new File(this.outputDirectory, current.getName());
 
+                if (current.getName().contains(".."))
```
matthieun, line 182: This solves the vulnerability by checking for parent folder injection... Not sure if it is the best way.

PR #666, opened 2020-08-07
Improvements to IsoCountryCodeCommand, WKTShardCommand, and AtlasSearchCommand

lucaspcram:
### Description:
+ Better country name matching when country names have multiple words. This makes it easier to search for countries if you only remember part of the name, e.g. if you are looking for St. Martin or Sint Maarten, and you only remember "Martin".

+ Fixed an unintuitive option in `WKTShardCommand` and updated the unit tests to use in-memory file system

+ Added some new options (`--predicate`, `--bounding-polygons`, `--sub-geometry`) to the `AtlasSearchCommand`.
        + in service of this, added an `entitiesWithin` method to `Atlas`

### Potential Impact:
Better commands.

### Unit Test Approach:
See added unit tests.

### Test Results:
👍

------


src/main/java/org/openstreetmap/atlas/locale/IsoCountryFuzzyMatcher.java
```diff
@@ -99,41 +65,46 @@
...
+    /*
+     * This algorithm could definitely be improved for cases with multi-word country matching.
+     */
     private static List<String> closestIsoCountries(final int number, final String displayCountry)
     {
         if (number <= 0 || number > IsoCountry.ALL_DISPLAY_COUNTRIES.size())
         {
             throw new CoreException("number " + number + " out of range (0, "
-                    + IsoCountry.ALL_DISPLAY_COUNTRIES.size() + ")");
+                    + IsoCountry.ALL_DISPLAY_COUNTRIES.size() + "]");
```
matthieun, line 76: Is that intended?

lucaspcram: Yes! I was trying to make it comply with this kind of notation: https://en.wikipedia.org/wiki/Bracket_(mathematics)#Intervals

lucaspcram:
@matthieun please don't merge yet! Adding one more thing.

PR #670, opened 2020-08-26
Avoid long overflow while comparing Edges

matthieun:
### Description:

When comparing `Edge`s with high Long identifiers, there was the potential for overflows when using the "difference" method. This is a simple fix for that!

### Potential Impact:

Anything relying on Edge ordering is fixed now. Example: AtlasDiff was randomly creating empty diffs, thinking that the in/out edge lists of some nodes were different because of the ordering bug.

### Unit Test Approach:

Added one unit test

### Test Results:

Pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java
@@ -50,19 +50,7 @@ public abstract class Edge extends LineItem implements Comparable<Edge>
     @Override
     public int compareTo(final Edge other) // NOSONAR
     {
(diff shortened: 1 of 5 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/items/Edge.java
```diff
@@ -50,19 +50,7 @@ protected Edge(final Atlas atlas)
...
-        }
-        else if (difference < 0)
-        {
-            return -1;
-        }
-        else
-        {
-            return 0;
-        }
+        return Long.compare(this.getIdentifier(), other.getIdentifier());
```
adahn6, line 53: 👍

PR #671, opened 2020-08-31
Capped Rectangle expansion at geographic poles

lucaspcram:
### Description:
`Rectangle#expand` and `Rectangle#expandVertically` are now capped at the North and South Poles. This prevents the expansion code from failing when large distances are applied to rectangles near the poles. The expansion capping uses a 1 meter buffer from the pole, so this fix will fail if the original rectangle has a corner already within 1 meter of a pole.

### Potential Impact:
Rectangle users no longer need to worry about latitude of rectangle corners when expanding.

### Unit Test Approach:
Added a unit test.

### Test Results:
👍

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
@@ -348,7 +348,7 @@ public final class Rectangle extends Polygon
     public Rectangle expandVertically(final Distance distance)
     {
         final double degreesLatitudeToShift = distance.asMeters()
-                / Distance.DISTANCE_PER_DEGREE_LATITUDE.asMeters();
+                / Distance.APPROXIMATE_DISTANCE_PER_DEGREE_AT_EQUATOR.asMeters();
         final long meterBuffer = 1;
 
         final Location oldLowerLeft = this.lowerLeft;
@@ -357,15 +357,15 @@ public final class Rectangle extends Polygon
         if (oldLowerLeft.getLatitude().asDegrees() - degreesLatitudeToShift <= Latitude.MINIMUM
                 .asDegrees())
         {
-            logger.error(
+            logger.warn(
                     "Provided distance {} would have shifted past the South Pole, truncating southward expansion...",
                     distance);
             final double degreesToHitMinimum = -1
                     * (Latitude.MINIMUM.asDegrees() - oldLowerLeft.getLatitude().asDegrees());
             // subtract a small buffer off the distance to just miss the pole.
             southShiftDistance = Distance
-                    .meters((Distance.DISTANCE_PER_DEGREE_LATITUDE.asMeters() * degreesToHitMinimum)
-                            - meterBuffer);
+                    .meters((Distance.APPROXIMATE_DISTANCE_PER_DEGREE_AT_EQUATOR.asMeters()
+                            * degreesToHitMinimum) - meterBuffer);
         }
         final Location lowerLeftShiftedSouth = oldLowerLeft.shiftAlongGreatCircle(Heading.SOUTH,
                 southShiftDistance);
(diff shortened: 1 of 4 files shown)


src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
```diff
@@ -350,10 +347,49 @@ public Rectangle expandHorizontally(final Distance distance)
...
+                / Distance.DISTANCE_PER_DEGREE_LATITUDE.asMeters();
+        final long meterBuffer = 1;
+
+        final Location oldLowerLeft = this.lowerLeft;
+        Distance southShiftDistance = distance;
+        // If the lowerLeft is about to be shifted south of the South Pole, stop it!
+        if (oldLowerLeft.getLatitude().asDegrees() - degreesLatitudeToShift <= Latitude.MINIMUM
+                .asDegrees())
+        {
+            logger.error(
```
matthieun, line 360: I would make it a WARN since that has the potential to flood a lot when anyone runs it on a lot of rectangles!

src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
```diff
@@ -350,10 +347,49 @@ public Rectangle expandHorizontally(final Distance distance)
      */
     public Rectangle expandVertically(final Distance distance)
```
matthieun, line 348: I think this fix should also be applied to the expandHorizontally function, so it stops at the antimeridian? Or is that already the case?

src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
```diff
@@ -350,10 +347,49 @@ public Rectangle expandHorizontally(final Distance distance)
...
+        final Location lowerLeftShiftedSouth = oldLowerLeft.shiftAlongGreatCircle(Heading.SOUTH,
+                southShiftDistance);
+
+        final Location oldUpperRight = this.upperRight;
+        Distance northShiftDistance = distance;
+        // If upperRight is about to be shifted north of the North Pole, stop it!
+        if (oldUpperRight.getLatitude().asDegrees() + degreesLatitudeToShift >= Latitude.MAXIMUM
+                .asDegrees())
+        {
+            logger.error(
```
matthieun, line 379: Same

src/main/java/org/openstreetmap/atlas/utilities/scalars/Distance.java
```diff
@@ -26,30 +25,38 @@
...
-    public static final Distance AVERAGE_EARTH_RADIUS = Distance.kilometers(6371);
     public static final Distance ZERO = Distance.millimeters(0);
+    public static final Distance MAXIMUM = Distance.millimeters(Long.MAX_VALUE);
+    public static final Distance AVERAGE_EARTH_RADIUS = Distance.kilometers(6371);
     public static final Distance TEN_MILES = Distance.miles(10);
     public static final Distance FIFTEEN_HUNDRED_FEET = Distance.feet(1500);
     public static final Distance ONE_METER = Distance.meters(1);
-    public static final Distance MAXIMUM = Distance.millimeters(Long.MAX_VALUE);
-
+    public static final Distance DISTANCE_PER_DEGREE_LONGITUDE_AT_EQUATOR = Distance.meters(111321);
```
matthieun, line 42: I think this one should be derived from the earth radius... This way we have only one hard coded constant.

src/main/java/org/openstreetmap/atlas/utilities/scalars/Distance.java
```diff
@@ -26,30 +25,38 @@
...
+    public static final Distance DISTANCE_PER_DEGREE_LATITUDE = Distance.meters(111120);
```
matthieun, line 43: and here too!
I am actually surprised this is not the same. I truth the earth is not perfectly spherical, but we have assumed it is here.

src/main/java/org/openstreetmap/atlas/geography/Rectangle.java
lucaspcram: 👍

src/main/java/org/openstreetmap/atlas/utilities/scalars/Distance.java
lucaspcram: Hmmm I should probably include more comments to clarify what is going on here. `DISTANCE_PER_DEGREE_LATITUDE` here is an approximation because you are correct, the earth is not perfectly spherical so this number varies based on your current latitude. I was unable to find a definition for the `distance_per_degree_latitude(currentLatitude)` function so I just used this approximation. I'll make it clear that this is what I am doing.

matthieun (approved):

PR #677, opened 2020-09-14
Added regex taggable filter

mm-ciub:
### Description:

During the extension of the **`InvalidTagsCheck`** from atlas-checks arose the need to validate certain tags based on regex patterns instead of tag-value combination. For a specific example, the check had to be extended to verify if source tags contain illegal values, like private company names. 

For this purpose, the **`RegexTaggableFilter`** similarly extends **`Predicate<Taggable>`** and checks if a set of given tag names match at least one pattern from a set of regex patterns and are not an exception to the general rule.

### Potential Impact:

This would not break backwards compatibility. Atlas-checks will have access to the new filter.

### Unit Test Approach:

The filter has been tested with:
- simple configuration containing just a tag and pattern with true and false cases
- extended configuration containing a set of tags and a set of patterns where either none, some or all should match 
- exception configuration with extended configuration where a matched case was also an exception
- tested the method returning the tag names that matched the patterns by comparing with the expected String

### Test Results:

Tested integration with the atlas-checks plugin using the bad source tag flow described above with positive results.

------


lucaspcram (approved):
Love it, and nice tests! @matthieun ?

src/main/java/org/openstreetmap/atlas/tags/filters/README.md
```diff
@@ -154,3 +154,28 @@ Here it is equivalent to:
...
+The filter also accepts a map of values that are excepted from the regex patterns. 
+
+In order to create `RegexTaggableFilter` one must provide:
+- a set of String values representing the tag names that need to be checked
+- a set of String values representing the regex used to match the values
+- optionally, a map of tag names and sets of excepted values.
+
+For example: 
+
+```
```
matthieun, line 170:
```suggestion
```java
```

src/main/java/org/openstreetmap/atlas/tags/filters/RegexTaggableFilter.java
```diff
@@ -0,0 +1,87 @@
...
+import org.openstreetmap.atlas.tags.Taggable;
+
+/**
+ * This {@link Taggable} filter relies on regex patterns to verify specific tag values.
+ * 
+ * @author mm-ciub on 11/09/2020.
+ */
+public class RegexTaggableFilter implements Predicate<Taggable>, Serializable
+{
+
```
matthieun, line 23:
```suggestion
```

src/main/java/org/openstreetmap/atlas/tags/filters/RegexTaggableFilter.java
```diff
@@ -0,0 +1,87 @@
...
+                if (match.isPresent() && !(this.exceptions.containsKey(tagName)
+                        && this.exceptions.get(tagName).contains(tagValue.get())))
+                {
+                    matchedTags.add(tagName);
+                }
+            }
+        }
+        return matchedTags;
+    }
+
```
matthieun, line 86:
```suggestion
```

src/test/java/org/openstreetmap/atlas/tags/filters/RegexTaggableFilterTest.java
```diff
@@ -0,0 +1,85 @@
...
+import org.junit.Assert;
+import org.junit.Test;
+import org.openstreetmap.atlas.tags.Taggable;
+
+/**
+ * @author mm-ciub on 14/09/2020.
+ */
+public class RegexTaggableFilterTest
+{
+
```
matthieun, line 19:
```suggestion
```

src/test/java/org/openstreetmap/atlas/tags/filters/RegexTaggableFilterTest.java
```diff
@@ -0,0 +1,85 @@
...
+        final Set<String> tagNames = new HashSet<>(Collections.singletonList("source"));
+        final Set<String> regex = new HashSet<>(Collections.singletonList(".*(?i)\\bmap\\b.*"));
+        final RegexTaggableFilter filter = new RegexTaggableFilter(tagNames, regex, null);
+
+        Assert.assertFalse(filter.test(Taggable.with()));
+        Assert.assertFalse(filter.test(Taggable.with("highway", "primary")));
+        Assert.assertFalse(filter.test(Taggable.with("source", "local knowledge")));
+        Assert.assertTrue(filter.test(Taggable.with("source", "illegal map")));
+    }
+
```
matthieun, line 84:
```suggestion
```

matthieun (changes requested):
This is a great addition, thanks for your contribution @mm-ciub!
Just a few cosmetic suggestions.. If you'd rather keep as-is, that is fine too, and we can approve/merge.
Next step we can update the `ConfiguredFilter` to take the options for a regex filter using that class!

matthieun (approved):

PR #686, opened 2020-09-29
MultiPolygon GeoJson Parser Fix

Bentleysb:
### Description:

The multipolygon data structure used to deserialize geojson multipolygons is based on the geojson polygon structure (one outer ring with possible inner rings) instead of the geojson multipolygon structure (one or more outer rings with possible inner rings). This causes the deserialization to fail on multipolygons.
This corrects the multipolygon structure to follow the geojson multipolygon format by expecting the extra level of nesting to allow for multiple outer rings. It also updates the `.toAtlasGeometry()` method to return an atlas MultiPolygon instead of a List of Atlas Polygons. 

### Potential Impact:

This is a breaking change for anything that uses the GeoJsonParserGsonImpl class or the associated helper classes for multipolygon deserialization. However, anything that was using this for multipolygons must not have been using true geojson for it to have worked. 

### Unit Test Approach:

Updated all unit tests for multipolygon deserialization in the GeoJsonParserGsonImplTest  & CoordinatesTest classes and their resource files. True geojson multipolygons are now tested, as are multiple levels of outer and inner rings. The tests were also updated to expect and evaluate Altas MultiPolygons from the `.toAtlasGeometry()` method.

### Test Results:

Only unit tests. 

------


src/main/java/org/openstreetmap/atlas/geography/geojson/parser/domain/geometry/MultiPolygon.java
```diff
@@ -1,53 +1,49 @@
...
  * @author Yazad Khambata
  */
 @SuppressWarnings("squid:S2160")
-public class MultiPolygon
-        extends AbstractGeometryWithCoordinateSupport<List<List<Position>>, List<Polygon>>
+public class MultiPolygon extends
+        AbstractGeometryWithCoordinateSupport<List<List<List<Position>>>, org.openstreetmap.atlas.geography.MultiPolygon>
 {
-    private MultiLineString value;
+    private Coordinates<List<List<List<Position>>>> coordinates;
```
matthieun, line 21: It seems this needs to be transient or serializable: https://sonarcloud.io/project/issues?id=org.openstreetmap.atlas%3Aatlas&issues=AXTa20XMSXmRf2zUqC-Q&open=AXTa20XMSXmRf2zUqC-Q&pullRequest=686
It might not be an issue with local testing, but if used within Spark it could become one.

matthieun: I am not sure here, but i the object is using that `coordinates` field only at initialization for example, transient might be good enough... Otherwise it will need to be serializable

matthieun (changes requested):
Thanks @Bentleysb, just one code smell that I think needs to be addressed!

src/main/java/org/openstreetmap/atlas/geography/geojson/parser/domain/geometry/MultiPolygon.java
Bentleysb: I ended up finding a number of similar classes with the same issue. My solution to make them serializable is not the prettiest, but there is a bunch of clean up around the Coordinates class to really do it right. I think that clean up makes more sense in a separate PR.

matthieun (approved):
Thanks for the comments! I am ok to do the updates in another PR if you feel this is the best.

PR #689, opened 2020-10-01
Make Relation ID generator deterministic

matthieun:
### Description:

Make sure `EntityIdentifierGenerator` is deterministic when generating IDs for relations that might have a different order of members but ultimately have the same.

### Potential Impact:

More determinism, less meaningless diffs.

### Unit Test Approach:

Updated existing unit test for the id generator

### Test Results:

Tests pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
@@ -3,6 +3,7 @@ package org.openstreetmap.atlas.geography.atlas.builder;
 import java.io.Serializable;
 import java.util.AbstractCollection;
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Iterator;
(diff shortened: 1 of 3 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/builder/RelationBean.java
```diff
@@ -49,6 +52,21 @@ public RelationBeanItem(final RelationBeanItem item)
...
+        public int compareTo(final RelationBeanItem other)
+        {
+            int result = this.getType().compareTo(other.getType());
+            if (result == 0)
+            {
+                result = Long.compare(this.getIdentifier(), other.getIdentifier());
+            }
+            if (result == 0)
+            {
+                result = this.getRole().compareTo(other.getRole());
```
adahn6, line 65: is this even possible?

matthieun: If you compare an item to itself, yes. But I think your question applies more to the same relation, and unfortunately yes too, since OSM does not enforce removing duplicates. This is why we keep lists along, instead of sets!

adahn6: I see, interesting!

adahn6 (approved):

PR #692, opened 2020-10-06
Big node improvement: Service and Link roads

Huyuntj:
### Description:

In this PR, we made some improvement to current big node model, includes,
1. enable service roads to be junction edge candidate
2. exclude link road to be considered as dual carriage way

### Potential Impact:

Will generate more dual carriage way big nodes including service road as junction edge; also will exclude out some FP edges connect to link road being identified as junction edges.

### Unit Test Approach:

Unit test added

### Test Results:

Pass

------


matthieun (approved):
Thank you @Huyuntj for the improvement. The test DNK file seems a bit large for it to be included as a test resource, but I think we can roll with it.

PR #693, opened 2020-10-07
CompleteRelation.changeMemberRole should list old role as explicitly excluded

matthieun:
### Description:

Changing the role of a member should be the same as removing the member and adding it back with a new role. As such, it also needs to add the old role as explicitly excluded, which was an omission here.

### Potential Impact:

There used to be duplicate member roles when doing multi-atlas operations on top of ChangeAtlas objects, since the neighboring Atlas still had the old role present. Now with the explicitly excluded mention, this should not happen anymore.

### Unit Test Approach:

Added one unit test

### Test Results:

Pass

------

### DIFF
--- a/src/test/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteRelationTest.java
+++ b/src/test/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteRelationTest.java
@@ -176,6 +176,24 @@ public class CompleteRelationTest
     }
 
     @Test
+    public void testMemberIsExplicitlyRemovedWhenRoleSwapped()
+    {
+        final String newRole = "newRole";
+        final String originalRole = "originalRole";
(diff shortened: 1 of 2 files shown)


Bentleysb (approved):

src/test/java/org/openstreetmap/atlas/geography/atlas/complete/CompleteRelationTest.java
```diff
@@ -175,6 +175,24 @@ public void testIsCompletelyShallow()
         Assert.assertTrue(superShallow.isShallow());
     }
 
+    @Test
+    public void testMemberIsExplicitlyRemovedWhenRoleSwapped()
```
adahn6, line 179: nice test!

adahn6 (approved):

lucaspcram (approved):
Great catch.

PR #696, opened 2020-10-08
Minor slicing changes

adahn6:
### Description:

This PR makes two minor changes. First, it adds additional logging around the number of slices being made to various entities to further increase transparency around the changes being made. Second, it adds an additional check for overlapping to rely not just on `.covers()` but also on `.intersects() && .intersection().getDimension()>0`. This additional check is added because there are extremely rare cases in which a sliced line should be covered by the original but isn't-- in these cases, the intersection of the two will be a 1-dimensional geometry so this calculation can give a more accurate result. 

### Potential Impact:
Almost none-- the logging is non-impactful and the corner case around sliced geometry is exceedingly rare.

### Unit Test Approach:

No tests here for this case-- predicting this geometry overlaps failure is exceedingly difficult, so it's not easy to capture in a unit test. 
### Test Results:

N/A

------


src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1188,7 +1189,14 @@ private void sliceArea(final Line line)
             return;
         }
 
-        if (!checkSlices(slices.keySet(), slices.values().size(), line))
+        long numSlices = 0;
+        for (final Set<org.locationtech.jts.geom.Polygon> sliceSet : slices.values())
+        {
+            numSlices += sliceSet.size();
+        }
+        logger.warn("Way {} was sliced into {} slices", line.getOsmIdentifier(), numSlices);
```
matthieun, line 1197: It seems this warn statement happens without check on the numSlices? Or is that snippet of code already in a corner-case section of code?

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1318,9 +1361,23 @@ private void sliceLine(final Line line)
...
+                intersectingBoundaryPolygons, line.getOsmIdentifier());
+
+        long numSlices = 0;
+        for (final Set<LineString> sliceSet : slices.values())
+        {
+            numSlices += sliceSet.size();
+        }
+        if (this.isAtlasEdge.test(line))
+        {
+            logger.warn("Edge {} was sliced into {} slices", line.getOsmIdentifier(), numSlices);
```
matthieun, line 1373: Same question/concern here!

matthieun (changes requested):
Thanks @adahn6!
Just a question/concern that might just be a non-issue!

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
adahn6: in this case, maybe `INFO` is more appropriate. the 0 number is still useful to log, as entities with bounding boxes overlapping a boundary but *not* intersecting the boundary will run through this logic only to have no slicing actually happen-- this log allows for seeing those ways directly.

lucaspcram (approved):

PR #698, opened 2020-10-09
Update ConfiguredFilter with RegexTaggableFilter

mm-ciub:
### Description:

Integrated the previously created `RegexTaggableFilter` in the `ConfiguredFilter` class.

### Potential Impact:

The impacted classes are only atlas classes that use the `ConfiguredFilter` which can now pass an extra `RegexTaggableFilter` if wanted. If the new configuration is not passed, the behaviour will be the same as before the changes. 

### Unit Test Approach:

Added a test for the new constructor in `RegexTaggableFilter` to assure the creation from `String` is successful.
Added verifiers in the previously existing `ConfiguredFilterTest` to make sure that the new filter addition does not break anything and behaves as expected, both during filtering and during JSON creation.

### Test Results:

No other tests have been conducted.


matthieun (changes requested):
Thanks for that integration!
Any chance you can also update the readme here? https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/tags/filters/README.md
I see you had already created a section for the regex, it would need a mention on how you create it with a String! Also the ~~ConfiguredTaggableFilter section seems outdated~~ ConfiguredFilter section is not there; if you have a chance to add it even better! Otherwise I will take care of this one at some point.
Thanks!

mm-ciub:
I have updated the documentation. I also added a section for the ConfiguredFilter. I hope the information is sufficient.

jwpgage (approved):

PR #704, opened 2020-10-26
ZipFileWritableResource no longer leaves open file descriptors

lucaspcram:
### Description:
`ZipFileWritableResource` no longer leaves open file descriptors lying around when interacting with `PackedAtlas`. Additionally, it now supports non-default file systems for easier testing.

### Potential Impact:
No more leaking FDs.

### Unit Test Approach:
Added a unit test using jimfs!

### Test Results:
👍

------


src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipFileWritableResource.java
```diff
@@ -44,64 +29,38 @@ public ZipFileWritableResource(final File source)
...
-                throw new IOException("Entry " + name + " does not exist.");
-            }
-        }
-        catch (final IOException e)
+        return new InputStreamResource(() ->
         {
-            throw new CoreException("Cannot get the entry {} from the Zipfile {}.", name,
-                    this.getFileSource().getName(), e);
-        }
+            final ZipInputStream zipStream = getZipInputStream();
```
matthieun, line 60: Are there any performance differences with this approach, i.e. opening a new file and its stream for each entry? Not sure if the `streamToSeekThrough.getNextEntry();` below is expensive or not.

src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipFileWritableResource.java
```diff
@@ -1,16 +1,17 @@
...
 
+import java.io.BufferedInputStream;
 import java.io.IOException;
-import java.io.InputStream;
-import java.util.function.Supplier;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
 import java.util.zip.ZipEntry;
-import java.util.zip.ZipFile;
```
matthieun, line 7: Yay!

matthieun (approved):
Nice! Just a question about performance!

src/main/java/org/openstreetmap/atlas/streaming/resource/zip/ZipFileWritableResource.java
lucaspcram: There is a moderate performance hit. I ran some tests where I rapidly opened a bunch of different atlas fields. Nothing huge. For example, opening and atlas and reading all the edge identifiers takes on average ~50ms longer now.

PR #706, opened 2020-10-29
Catch way-sectioning errors

matthieun:
### Description:

This is to catch way-sectioning errors that might come from corrupted data, before a whole way-sectioning process has to fail. The Lines that cannot be way-sectioned are logged and then dropped.

### Potential Impact:

More way-sectioning jobs can go through without failing. However that increases the risk of dropping large amounts of data in case a future underlying bug is introduced and the way-sectioning process does not fail.

### Unit Test Approach:

Added new test, with specific OSM data.

### Test Results:

Pass

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/sectioning/WaySectionProcessor.java
@@ -939,14 +939,21 @@ public class WaySectionProcessor
         changeSet.getLinesThatBecomeEdges().forEach(lineIdentifier ->
         {
             final Line line = this.rawAtlas.line(lineIdentifier);
-            final List<TemporaryEdge> edges;
-            if (line.isClosed())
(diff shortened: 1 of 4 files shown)


MikeGost:
Miss this code! Where kind of corrupt data was it?

matthieun:
The issue is with this OSM way that was recently deleted! https://www.openstreetmap.org/way/858888726/history
```
LINESTRING (49.9933986 14.9056073, 49.9933775 14.9053719, 49.9929567 14.9054263, 49.9933775 14.9053719, 49.9933986 14.9056073, 49.9933775 14.9053719, 49.9933986 14.9056073)
```
Does not look like it, but it overlaps itself in many places!
image

matthieun:
Looks like it was just replaced by this one: https://www.openstreetmap.org/way/864587512

MikeGost:
Oh interesting! I remember seeing similar junk when testing this logic, glad it was cleaned up!

matthieun:
Fixed code smells. Seems the duplication was pre-existing, so leaving that...

adahn6 (approved):

PR #711, opened 2020-12-07
TaggableFilters overhauled with new TaggableMatcher class

lucaspcram:
### Description:
`TaggableFilter` syntax has been overhauled, with the new functionality packaged in the roughly equivalent `TaggableMatcher` class. `TaggableMatcher` is intended to be a drop-in replacement for `TaggableFilter` wherever it's used.

An example of the syntax overhaul:
```
# Old TaggableFilter syntax
foo->bar||baz->bat&cat->mat
```
```
# New TaggableMatcher syntax
(foo = bar | baz=bat) & cat = mat
```
Regexes are supported like:
```
name = /.*\b[s|S]treet\b.*/
```
See the included README for many more details. [You can view the README here.](https://github.com/lucaspcram/atlas/tree/taggable/src/main/java/org/openstreetmap/atlas/tags/filters/matcher)

The `ConfiguredFilter` class can now specify a `taggableMatcher` field to use the new syntax. There is also now a `ConfiguredTaggableMatcher` class, which is a mirror class to `ConfiguredTaggableFilter`.
```
"junctionRoundaboutFilter":
{
    "predicate":
    {
        "imports": [
            "org.openstreetmap.atlas.geography.atlas.items"
        ],
        "command": "e instanceof Node || e instanceof Edge"
    },
    "taggableMatcher": "junction = roundabout"
},
```

PR also includes an atlas command (`TaggableMatcherPrinter`) for printing the TaggableMatcher trees. This may help debugging. For example, the tree for the above `TaggableMatcher` would generate this output:
```
$ atlas print-matcher '(foo = bar | baz=bat) & cat = mat'
(foo = bar | baz=bat) & cat = mat
                                &
                ┌───────────────┴───────────────┐
                |                               =
        ┌───────┴───────┐               ┌───────┴───────┐
        =               =              cat             mat
    ┌───┴───┐       ┌───┴───┐
   foo     bar     baz     bat
```

### Potential Impact:
None so far. Integration efforts will be slow, as there are many `TaggableFilters` in the wild. For now, it is an explicit opt-in feature. For example, `ConfiguredFilter` can specify a `taggableMatcher` field to use the new syntax.

### Unit Test Approach:
All components tested thoroughly. See all the added unit tests.

### Test Results:
👍

------


src/main/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/Lexer.java
```diff
@@ -0,0 +1,433 @@
...
+     * @param inputLine
+     *            the input line
+     * @return a {@link List} of the processed {@link Token}s
+     */
+    public List<Token> lex(final String inputLine) // NOSONAR
+    {
+        final List<Token> lexedTokens = new ArrayList<>();
+        final LexemeBuffer lexemeBuffer = new LexemeBuffer();
+        final InputBuffer inputBuffer = new InputBuffer(inputLine);
+        while (inputBuffer.peek() != InputBuffer.EOF)
```
matthieun, line 129: Wow, this method!

src/main/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/Parser.java
```diff
@@ -0,0 +1,464 @@
...
+        this.tokenBuffer = new TokenBuffer(tokens, inputLine);
+        this.inputLine = inputLine;
+    }
+
+    public ASTNode parse()
+    {
+        BinaryOperator.clearIdentifierCounter();
+        UnaryOperator.clearIdentifierCounter();
+        Operand.clearIdentifierCounter();
+        return or();
```
matthieun, line 110: Aaah you start with or I see

src/test/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/tree/TreePrinterTest.java
```diff
@@ -0,0 +1,63 @@
...
+{
+    @Test
+    public void test()
+    {
+        final String input1 = "foo=bar | !(baz=bat)";
+        ASTNode root = new Parser(new Lexer().lex(input1), input1).parse();
+        final String tree1 = TreePrinter.print(root);
+        Assert.assertEquals(
+                "                                |                               \n"
+                        + "                ┌───────────────┴───────────────┐               \n"
```
matthieun, line 21: This is a bold test... What if we change the printing slightly in the future? It has to be all re-printed here?

matthieun (commented):
This is really nice, really easy to follow, great work @lucaspcram!
I am curious about why the Lexer and Parser have to be excluded from Checkstyle..? Is that something unavoidable?

lucaspcram:
@matthieun I excluded method name rule specifically, because I wanted my method names to match the token/transition rule names for clarity. But I guess check style won't allow short method names? It was complaining about my methods called `or()` and `xor()`, etc. Which was just annoying so I disabled it.

src/main/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/Lexer.java
lucaspcram: Haha what about it? A lot of `if` statements? ;)

src/main/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/Parser.java
lucaspcram: Yes! If you check the grammar at the top of the class you can see that the OR transition is the first one. So we start recursive descent with that rule.

src/test/java/org/openstreetmap/atlas/tags/filters/matcher/parsing/tree/TreePrinterTest.java
lucaspcram: Ha this is a good point. I actually do want to update this in the future to improve the printing of `!` operators. So I will leave these tests in place for now to make sure when I do fiddle with things, it doesn't break.

lucaspcram:
@mm-ciub I see you added some regex features to the old TaggableFilter in #677 . Does my PR fulfill many of those same needs? If not, what additional features would you want to see in order to convince you to use TaggableMatchers? Thanks :)

mm-ciub:
> @mm-ciub I see you added some regex features to the old TaggableFilter in #677 . Does my PR fulfill many of those same needs? If not, what additional features would you want to see in order to convince you to use TaggableMatchers? Thanks :)

Hey. This looks great and it would be awesome if I could use it. Right now, the `RegexTaggableFilter` has a couple of useful features and I wonder if the new `TaggableMatcher` could support them:
- the`RegexTaggableFilter` takes a list of tag names and a list of regex patterns and tests the patterns for value match for each of them. This helps for cases where one tag has multiple regex or the other way around; can this be done using this matcher or must there be a syntax like `source=/regex1/|source=/regex2/|source_maxspeed=/regex1/` etc.
- The `RegexTaggableFilter` allows a map of excepted values -> each tag has a list of values that will match the regex pattern but are considered valid. Could something like this be added in the `TaggableMatcher`?

lucaspcram:
@mm-ciub Cool! The design behind `TaggableMatcher` is to encourage users to use familiar boolean operations to express their desired match condition. That way it is easy for new developers to quickly understand what a matcher is doing, since most everyone is already familiar with operators like `&`, `|`, etc. So let me see if I understand your requests.

>...for cases where one tag has multiple regex or the other way around; can this be done...

You could do something like:
```
source=(/regex1/ | /regex2/) | source_maxspeed=/regex1/
```
You could also do a regex in both the key and value if that is helpful, so for e.g.
```
/source(_maxspeed)?/=(/regex1/ | /regex2/)
```
The catch here is that this would also match `source_maxspeed=regex2`, which may not be desired.

> ...each tag has a list of values that will match the regex pattern but are considered valid [sic, did you mean invalid?]. Could something like this be added...

You could express this in a boolean-like way. For example, you could exclude "Lake Michigan" from a general lake matcher like so:
```
name = /.*\b[l|L]ake\b.*/ & name != "Lake Michigan"
```
Another example: if you want to match any tag that looks like `speedlimit=$NUMBER`, but exclude speed limits less than 20, you could do this:
```
speedlimit = /[0-9]+/ & speedlimit != /[0-1]?[0-9]/
```

Did these examples clear things up? If there's something I missed or failed to understand, please let me know! If there is some extra behavior/feature that would make `TaggableMatcher` more attractive, let me know because I may be able to implement it 🙂

adahn6 (approved):

PR #712, opened 2020-12-11
Use Jackson instead of Gson for geojson parser

seancoulter:
### Description:

This changes GeoJsonParserGsonImpl to GeoJsonParserJacksonImpl. It was found that Gson is insufficient for deserializing some OSM feature identifiers (which are Longs) from geojsons. Particularly, we ran into numerical truncation problems. The old GeoJsonParserGsonImpl deserializes strings of geojson into Objects using the ObjectTypeAdapter offered by Gson. The root of the truncation problem is that this adapter tokenizes all Numbers in the geojson into doubles: https://github.com/google/gson/blob/master/gson/src/main/java/com/google/gson/internal/bind/ObjectTypeAdapter.java#L79 . So, when we have a large OSM feature identifier in the geojson, its bits get shaved when it's converted to a Double (Doubles have less magnitude bit storage than Longs), then it's eventually converted back to a Long with those lost bits unaccounted for. The result is a deserialized identifier that is not the same as the serialized identifier.

Looks like Gson is aware of this and we might have a fix in the future: https://github.com/google/gson/pull/1290

Jackson does not appear to suffer from this truncation issue. There are other changes in this PR to support the migration to using Jackson. Since Jackson deserializes Numbers to be the same as their true serialized type (which could be an Integer or Double or Long etc) instead of always Doubles, we pass around Numbers in Coordinates.java and only convert them to Doubles when creating a Position, ie when we need them to be Doubles.

I also updated a test to use a large signed Long as an identifier and changed one instance of single quotes -> double quotes since apparently single quotes are not JSON conformant (Gson may have disregarded this before)

### Potential Impact:

All usage of GeoJsonParserGsonImpl in all projects should be renamed to GeoJsonParserJacksonImpl. Inputs and outputs should not change with the exception of certain identifiers being correctly deserialized in the output. Remains to be seen the overall impact of performance.

### Unit Test Approach:

Same unit tests and updated a couple

### Test Results:

Passing

------


src/test/resources/org/openstreetmap/atlas/geography/geojson/parser/featureChangePropertiesExample2.json
```diff
@@ -41,7 +41,7 @@
           "name": "RELATION_MEMBER",
           "type": "ADD",
           "itemType": "NODE",
-          "id": 402306209000000,
+          "id": -1374305525422343588,
```
matthieun, line 44: I am curious as to why this test ID had to change?

matthieun (approved):
This is a great analysis, thanks @seancoulter.
I am also glad that it works with the provided version of Jackson, so we do not have to worry about future dependency conflicts.

src/test/resources/org/openstreetmap/atlas/geography/geojson/parser/featureChangePropertiesExample2.json
seancoulter: This was actually just to test that the parser correctly deserialized this id since it wasn't doing so before the changes. I just picked a feature from existing test data and assigned it a new id to test this out. Only usage of this file appears to be for this test so I wasn't concerned with keeping ids consistent with OSM

lucaspcram (approved):

PR #720, opened 2021-01-05
Adding 3 highway tags that are not included.

reichg:
Simply adding 3 highway tags from [OSM Wiki](https://wiki.openstreetmap.org/wiki/Key:highway)


matthieun (approved):
Looks good to me!

PR #721, opened 2021-01-06
Add equal to MultiPolyLine

chunzc:
### Description:

This PR added isEquals and hashCode to MultiPolyLine class to make it comparable.

### Potential Impact:

N/A

### Unit Test Approach:

Added testEquals in MultiPolyLineTest.

### Test Results:

All tests passed.

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
@@ -89,9 +89,31 @@ public class MultiPolyLine
     }
 
     @Override
+    public boolean equals(final Object other)
+    {
+        if (!(other instanceof MultiPolyLine))
+        {
+            return false;
+        }
+        final MultiPolyLine otherItem = (MultiPolyLine) other;
+        return this.hashCode() == otherItem.hashCode();
+    }
+
+    @Override
     public GeoJsonType getGeoJsonType()
     {
         return GeoJsonType.MULTI_LINESTRING;
+    }
+
+    @Override
+    public int hashCode()
+    {
+        int hash = 0;
+        for (final PolyLine polyLine : this.polyLineList)
+        {
+            hash += polyLine.hashCode();
+        }
+        return hash;
     }
 
     @Override
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
@@ -96,16 +96,7 @@ public class MultiPolyLine
             return false;
         }
         final MultiPolyLine otherItem = (MultiPolyLine) other;
-        final Iterator<PolyLine> iterator = this.iterator();
-        final Iterator<PolyLine> iteratorOther = otherItem.iterator();
-        while (iterator.hasNext() && iteratorOther.hasNext())
-        {
-            if (!iterator.next().equals(iteratorOther.next()))
-            {
-                return false;
-            }
-        }
-        return !iterator.hasNext() && !iteratorOther.hasNext();
+        return this.hashCode() == otherItem.hashCode();
     }
 
     @Override
--- a/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
@@ -96,7 +96,16 @@ public class MultiPolyLine
             return false;
         }
         final MultiPolyLine otherItem = (MultiPolyLine) other;
-        return this.hashCode() == otherItem.hashCode();
+        final Iterator<PolyLine> iterator = this.iterator();
+        final Iterator<PolyLine> iteratorOther = otherItem.iterator();
+        while (iterator.hasNext() && iteratorOther.hasNext())
+        {
+            if (!iterator.next().equals(iteratorOther.next()))
+            {
+                return false;
+            }
+        }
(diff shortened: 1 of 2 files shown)


src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
```diff
@@ -88,12 +88,34 @@ public Rectangle bounds()
...
 
+    @Override
+    public boolean equals(final Object other)
+    {
+        if (!(other instanceof MultiPolyLine))
+        {
+            return false;
+        }
+        final MultiPolyLine otherItem = (MultiPolyLine) other;
+        return this.hashCode() == otherItem.hashCode();
```
adahn6, line 99: pretty concerned about using this as a measure for equality. hash codes can potentially lead to both false negatives and false positives in certain cases

src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
```diff
@@ -88,12 +88,34 @@ public Rectangle bounds()
...
         return GeoJsonType.MULTI_LINESTRING;
     }
 
+    @Override
+    public int hashCode()
+    {
+        int hash = 0;
+        for (final PolyLine polyLine : this.polyLineList)
+        {
+            hash += polyLine.hashCode();
```
adahn6, line 114: this concerns me a lot-- this could easily overflow the integer, leading to unexpected behavior

adahn6 (changes requested):
I think we should probably implement a more robust and deterministic `.equals()` than just hashcode comparisons, this current implementation will probably lead to cornercases of incorrect results. maybe compare underlying geometry and tags?

src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
chunzc: Thanks for the comment, I will use String to record the hash in the list.

src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
chunzc: I've changed hashCode to generate from concatenated string of polyLine's hashCode.

snb3300: @chunzc I think you can still run into integer overflow in the ```hashCode``` method below. Instead of relying on the hashCode you can probably compare the geometry. Maybe ```wkt``` if possible?

chunzc: ok, I will change it back to geometry comparison to apply equals on each polyline via iterator.

src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
```diff
@@ -88,12 +88,43 @@ public Rectangle bounds()
...
+        if (!(other instanceof MultiPolyLine))
+        {
+            return false;
+        }
+        final MultiPolyLine otherItem = (MultiPolyLine) other;
+        final Iterator<PolyLine> iterator = this.iterator();
+        final Iterator<PolyLine> iteratorOther = otherItem.iterator();
+        while (iterator.hasNext() && iteratorOther.hasNext())
+        {
+            if (!iterator.next().equals(iteratorOther.next()))
```
snb3300, line 103: It looks like we are expecting the order of polylines to be the same in both objects. Do we want the equals to be this strict?

chunzc: ok, I can relax that.

adahn6 (dismissed):

adahn6:
thanks for addressing this! looks good now 👍

src/main/java/org/openstreetmap/atlas/geography/MultiPolyLine.java
```diff
@@ -88,12 +90,42 @@ public Rectangle bounds()
...
+    public boolean equals(final Object other)
+    {
+        if (!(other instanceof MultiPolyLine))
+        {
+            return false;
+        }
+        final MultiPolyLine otherItem = (MultiPolyLine) other;
+        final Set<PolyLine> polyLineSet = new HashSet<>(this.polyLineList);
+        final Set<PolyLine> otherPolyLineSet = new HashSet<>(otherItem.getPolyLineList());
+        return otherItem.getPolyLineList().size() == this.getPolyLineList().size()
```
snb3300, line 103: As ```polyLineSet``` & ```otherPolyLineSet``` are HashSet, I think you can directly return ```polyLineSet.equals(otherPolyLineSet)```. equals method for set checks for the size equality.

src/test/java/org/openstreetmap/atlas/geography/MultiPolyLineTest.java
```diff
@@ -138,4 +138,17 @@ public void testCreateMultiPolyLineFromWKT()
         Assert.assertTrue(polyLines.contains(polyLine1));
         Assert.assertTrue(polyLines.contains(polyLine2));
     }
+
+    @Test
+    public void testEquals()
```
snb3300, line 143: Should we also test for a case where two ```MultiPolyLine``` are different?

chunzc: Sure, I will add a case for that.

lucaspcram (approved):
Thanks for these updates!

PR #723, opened 2021-01-08
Adding highway tags and functionality

reichg:
### Description:

These functions help determine if the tag should belong to a node, a way, or both. Hopefully this is allowed here.


reichg:
@seancoulter @Bentleysb 
Should this be done in the Atlas-Checks repo?

Bentleysb:
@reichg, I think this is a good addition here, you just need to make a test class for it [here](https://github.com/osmlab/atlas/tree/dev/src/test/java/org/openstreetmap/atlas/tags) and add unit tests for the new methods.

reichg:
> @reichg, I think this is a good addition here, you just need to make a test class for it [here](https://github.com/osmlab/atlas/tree/dev/src/test/java/org/openstreetmap/atlas/tags) and add unit tests for the new methods.

Let me know what you think of the tests!

src/test/java/org/openstreetmap/atlas/tags/HighwayTagTestCase.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author v-garei
+ */
+public class HighwayTagTestCase
+{
+
+    @Test
+    public void highwayTagTests()
+    {
+        final Taggable emergencyBayTaggable = Taggable.with(HighwayTag.KEY, "emergency_bay");
```
matthieun, line 19: Here for the sake of searchability, I would replace `"emergency_bay"` with `HighwayTag.EMERGENCY_BAY.toString()`... Unless what you meant to test here is that the enum is properly spelled.

matthieun (changes requested):
Thanks for a nice addition! Just one suggestion/question. If not an issue I can approve :)

src/test/java/org/openstreetmap/atlas/tags/HighwayTagTestCase.java
reichg: you right you right!

reichg:
not sure what to do about that build error

matthieun:
We are migrating to another build system (github actions) and there are a few loose ends I am still working on. Sorry for the confusion...

src/test/java/org/openstreetmap/atlas/tags/HighwayTagTestCase.java
```diff
@@ -0,0 +1,44 @@
...
+
+        final TestTaggable falseNodeOnlyTag = new TestTaggable(HighwayTag.KEY,
+                HighwayTag.TRUNK.toString());
+        Assert.assertFalse(HighwayTag.isNodeOnlyTag(falseNodeOnlyTag));
+
+        final TestTaggable falseWayOnlyTag = new TestTaggable(HighwayTag.KEY,
+                HighwayTag.BUS_STOP.toString());
+        Assert.assertFalse(HighwayTag.isWayOnlyTag(falseWayOnlyTag));
+
+        final TestTaggable trueNodeOnlyTag = new TestTaggable(HighwayTag.KEY, "traffic_mirror");
```
matthieun, line 31: Did you want to also change this one and the two below to use the enum?

matthieun (changes requested):

src/test/java/org/openstreetmap/atlas/tags/HighwayTagTestCase.java
reichg: > Did you want to also change this one and the two below to use the enum?

yeah I am dumb, for some reason I was thinking that since these ones were new they wouldn't be accessible yet.. whoops

matthieun (approved):
Thanks for the fixes!

PR #727, opened 2021-01-14
Closeable Resources

tsmock:
### Description:

Add Closeable Resources, for use when a Resource depends upon an AutoCloseable resource that must not be closed prior to the user finishing with the resource.

### Potential Impact:

This largely comes down to memory impact. In https://github.com/osmlab/atlas-generator/pull/167, closing resources significantly freed up space and sped up processing. Unfortunately, during testing, a bug where resources were closed prior to reads was missed. This fixes that issue by creating resources that store the closeable resources internally, which can then be closed automatically when finished.

### Unit Test Approach:

The added unit tests use JUnit5 to check assertions (I find that the `assertThrows`/`assertDoesNotThrow` syntax to be much easier to use for checking for assertions).

The addition of JUnit5 support should not affect any existing tests (there is a vintage runner for JUnit4 and JUnit3 tests).

The unit tests check various conditions where the `ResourceCloseable`s are closed (so, `try-with-resources` or manually), and checks for what happens when a required resource is unexpectedly closed).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -38,7 +38,6 @@ configurations
 dependencies
 {
     compile packages.slf4j.api
-    compile packages.junit
     compile packages.opencsv
     compile packages.gson
     compile packages.http
@@ -61,6 +60,8 @@ dependencies
     compile packages.jackson.core
     compile packages.jackson.databind
     compile packages.jackson.dataformat
+    // Support JUnit 3/4 tests
+    compile packages.junit.junit4
     compile packages.protobuf_java
     compile packages.protoc
     compile packages.artifact
@@ -78,9 +79,10 @@ dependencies
     shaded project.configurations.getByName('compile')
     shaded packages.slf4j.log4j12
     shaded packages.log4j
+
 }
 
-task shaded(type: Jar){
+task shaded(type: Jar) {
     baseName = project.name
     classifier = 'shaded'
     from
--- a/src/main/java/org/openstreetmap/atlas/streaming/resource/ResourceCloseable.java
+++ b/src/main/java/org/openstreetmap/atlas/streaming/resource/ResourceCloseable.java
@@ -1,0 +1,52 @@
+package org.openstreetmap.atlas.streaming.resource;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Objects;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import org.openstreetmap.atlas.exception.CoreException;
+
+/**
+ * Use when the returned resource has some items that cannot be closed prior to finishing with this
+ * resource.
+ *
+ * @author Taylor Smock
+ */
+public interface ResourceCloseable extends Resource, AutoCloseable
+{
+    @Override
+    default void close() throws Exception
+    {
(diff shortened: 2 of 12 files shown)


build.gradle
```diff
@@ -78,9 +77,29 @@ dependencies
     shaded project.configurations.getByName('compile')
     shaded packages.slf4j.log4j12
     shaded packages.log4j
+
+    // Support Junit 5 tests
+    testImplementation packages.junit.api
+    testRuntimeOnly packages.junit.engine
+    // Support JUnit 3/4 tests
+    compile packages.junit.junit4
+    testRuntimeOnly packages.junit.vintage
```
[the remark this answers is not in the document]
tsmock: I'll check that -- I was replacing the `compile packages.junit` earlier in the stanza. I *think* that needs to be kept (as `compile packages.junit.junit4` since some of the non-test classes uses JUnit4 in some of the downstream projects, and I presume that some non-test classes do the same in `atlas`. I'll check.).

build.gradle
```diff
@@ -78,9 +77,29 @@ dependencies
...
+ * Artifact related items
+ */
+task javadocJar(type: Jar) {
+    classifier = 'javadoc'
+    from javadoc
+}
+
+task sourcesJar(type: Jar) {
+    classifier = 'sources'
+    from sourceSets.main.allSource
```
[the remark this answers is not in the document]
tsmock: Yep.

src/main/java/org/openstreetmap/atlas/streaming/resource/ResourceCloseable.java
```diff
@@ -0,0 +1,48 @@
...
+                closeable.close();
+            }
+            catch (final Exception exception)
+            {
+                exceptions.add(exception);
+            }
+        }
+        if (!exceptions.isEmpty())
+        {
+            throw exceptions.iterator().next();
```
matthieun, line 35: It would be nice here to have a more detailed message, with the number of exceptions, and the cause set as the first one?
```java
throw new CoreException("{} exceptions thrown while closing {}", 
    exceptions.size(), this.getName(), exceptions.iterator().next());
```

build.gradle
```diff
@@ -78,9 +77,12 @@ dependencies
     shaded project.configurations.getByName('compile')
     shaded packages.slf4j.log4j12
     shaded packages.log4j
+
+    // Support JUnit 3/4 tests
+    compile packages.junit.junit4
```
matthieun, line 82: Can this one be with all the other compile up top? Just a nit.

matthieun (changes requested):
Thanks! Just a few optional details left!

src/main/java/org/openstreetmap/atlas/streaming/resource/ResourceCloseable.java
tsmock: This can be done. I really wanted to merge them all together, initially, but I ended up deciding that it might be a good idea to just throw the first once, since later exceptions might be thrown due to the first exception.

build.gradle
tsmock: Sure.

matthieun (approved):
Thanks!

PR #728, opened 2021-01-19
EntityIdentifierGenerator now has some config options

lucaspcram:
### Description:
EntityIdentifierGenerator now has some config options that can be accessed with the `EntityIdentifierGenerator.Configuration` class.

For example, users can specify that they only want ID generation to account for geometry:
```java
long id = new EntityIdentifierGenerator.Configuration().useGeometry()
                .getGenerator().generateIdentifier(point);
```

Or users can opt to generate based on everything **except for tags**:
```java
long id = new EntityIdentifierGenerator.Configuration().useDefaults().excludeTags()
                .getGenerator().generateIdentifier(point);
```

This assertion will pass:
```java
// The old style of generating an ID that accounts for all feature attributes
long id1 = new EntityIdentifierGenerator().generateIdentifier(point);

// The exact same thing, expressed with the Configuration object
long id2 = new EntityIdentifierGenerator.Configuration().useDefaults()
                .getGenerator().generateIdentifier(point);

assert id1 == id2;
```

Trying to use an empty `EntityIdentifierGenerator.Configuration` will yield the following error:
```java
// Config is empty, no attributes were specified
new EntityIdentifierGenerator.Configuration().getGenerator().generateIdentifier(point);

org.openstreetmap.atlas.exception.CoreException: EntityIdentifierGenerator.Configuration was empty! Please set at least one of geometry, tags, or relation members.
```

If you try to use a `Configuration` set for attributes that do not exist in the given features, this will also generate an error:
```java
EntityIdentifierGenerator nonRelationInvariantGenerator = new EntityIdentifierGenerator.Configuration()
            .useRelationMembers().getGenerator();

// Works fine on relation type entities
nonRelationInvariantGenerator.generateIdentifier(relation);

// Config is non-relation invariant, so will throw an error for non-relation type entities
nonRelationInvariantGenerator.generateIdentifier(point);

org.openstreetmap.atlas.exception.CoreException: EntityIdentifierGenerator.Configuration was non-relation invariant! Please set at least one of geometry or tags to generate IDs for non-relation type entities.
```

### Potential Impact:
None. New features are opt-in. Old code will work as-is.

### Unit Test Approach:
See updated unit tests.

### Test Results:
👍

------


adahn6:
ahh! this is super cool! 🔥

src/main/java/org/openstreetmap/atlas/utilities/identifiers/EntityIdentifierGenerator.java
```diff
@@ -26,9 +26,138 @@
  */
 public class EntityIdentifierGenerator
 {
+    /**
+     * A config class for {@link EntityIdentifierGenerator} to set various configuration parameters.
```
adahn6, line 30: making this an inner class is a great idea 👍

adahn6 (approved):
code looks good to me. out of curiosity, any idea how this could affect possible hashing collisions?

matthieun (approved):
Very good!

Huyuntj (approved):
Changes looks good !

lucaspcram:
@adahn6 Provided the configuration is left default, it shouldn't affect much. Obviously though, you will end up with two different IDs for a single feature if you change the attributes. I think the collision probability is still relatively low. We're basically taking the top half of an MD5 hash.

PR #690, opened 2020-10-02
Use JTS Polygonizer to re-construct Polygons

matthieun:
### Description:

Use JTS Polygonizer to re-construct Polygons; this allows the construction of Polygons touching in one single point, which the legacy algorithm did not allow for.

In case the Polygonizer does not construct a polygon, we attempt a second pass with the legacy algorithm.

### Potential Impact:

More polygons are reconstructed. For example, the legacy algorithm can't build Austria's boundary because of that self touching point here: https://www.openstreetmap.org/node/2064631022. With the new algorithm, the boundary will have two outers which touch in that point.

### Unit Test Approach:

- Updated existing unit tests to account for JTS option
  - Added Polygon and Multipolygon `isSimilar()` using JTS's [HausdorffSimilarityMeasure](https://locationtech.github.io/jts/javadoc/org/locationtech/jts/algorithm/match/HausdorffSimilarityMeasure.html) to allow for the inconsistencies in which the new algorithm starts the polygons; comparing WKTs in tests does not cut it anymore.
- Added specific unit tests for the enw capabilities

### Test Results:

Pass

------

In doubt: [Contributing Guidelines](CONTRIBUTING.md)

PR #733, opened 2021-02-05
Normalize Polygon geometry for consistency

snb3300:
### Description:

```MultiplePolyLineToPolygonsConverter``` uses JTS ```Polygonizer``` to create Polygons from a list of Polylines. Normalize the result Polygon(s) for consistent geometry.

Based on https://github.com/osmlab/atlas/pull/690 Use Polygonizer when creating geometry for a ```ComplexWaterBody``` to include water bodies with outers intersecting at one point as they are valid

### Unit Test Approach:

All existing unit tests should pass

------


lucaspcram (approved):
Great update! Thanks for this.

PR #734, opened 2021-02-11
Minor slicing fix for certain areas

adahn6:
### Description:

In rare cases, an Area can get sliced into resulting multipolygon geometry; this corner case arises when an Area completely contains a boundary, leaving one or more holes after slicing. Since Areas only support simply polygon geometry, the code path rejects this new sliced geometry and attempt to convert the Area to a linear Line geometry and slice that. However, in this case, using `Area.asPolygon()` in the new CompleteLine accidentally preserves the polygonal geometry. So instead, a fix here casts it to a JTS Polygon, then converts the exterior ring to a PolyLine, which guarantees a linear representation when sliced again.
 
### Potential Impact:

Extremely limited, this is an unusual corner case.

### Test Results:

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
@@ -1319,7 +1319,9 @@ public class RawAtlasSlicer
             final Set<Long> relationIds = new HashSet<>();
             area.relations().forEach(relation -> relationIds.add(relation.getIdentifier()));
             final CompleteLine lineFromArea = new CompleteLine(area.getIdentifier(),
-                    area.asPolygon(), area.getTags(), relationIds);
+                    JTS_POLYLINE_CONVERTER.backwardConvert(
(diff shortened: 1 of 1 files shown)


lucaspcram (approved):

src/main/java/org/openstreetmap/atlas/geography/atlas/raw/slicing/RawAtlasSlicer.java
```diff
@@ -1319,7 +1319,9 @@ private void sliceArea(final Area area)
             final Set<Long> relationIds = new HashSet<>();
             area.relations().forEach(relation -> relationIds.add(relation.getIdentifier()));
             final CompleteLine lineFromArea = new CompleteLine(area.getIdentifier(),
-                    area.asPolygon(), area.getTags(), relationIds);
+                    JTS_POLYLINE_CONVERTER.backwardConvert(
```
MikeGost, line 1322: Is it worth adding a test case for this edge case?

adahn6: probably 💀for now I'm prioritizing a fix since this is a one-in-a-million corner case, but I'll see if I can capture the behavior in a sufficient unit test

PR #736, opened 2021-02-12
Fix for certain tags being dropped during sectioning

adahn6:
### Description:

In the rare case of invalid geometry during sectioning, a misunderstanding of the FeatureChange `.withAddedTag()` API led to `synthetic_invalid_way_section` tag being added but all other tags being dropped. This issue has been fixed by properly instantiating the CompleteLine object it intends to add the tag to with the full TagMap before invoking `.withAddedTag()`. 

### Potential Impact:

Limited other than not dropping tags for this object.

### Unit Test Approach:
One added to capture the expected results

### Test Results:
The original tags are no longer dropped

------


AlexHsieh22 (dismissed):

matthieun (changes requested):
Nice find! Please add a quick unit test to make sure tags are there?

matthieun (approved):
Thanks a lot for adding the test!

PR #738, opened 2021-02-18
Allow expand indefinitely for relation exploration

matthieun:
### Description:

When toggling the aggressive relation expansion (to explore neighboring shards in search for relations with the same id) the dynamic atlas expander was ignoring it when the indefinite expansion is true, thinking it would be redundant. But in some cases, when the `entitiesToConsiderForExpansion` predicate is definded, or when some sparse areas have relations with far flung members (which would not be explored by regular indefinite expansion) this behavior was unexpected. 

### Potential Impact:

In more cases with neighboring shards be explored for relations.

### Unit Test Approach:

Added one unit test

### Test Results:

Relied on unit tests and other simple local testing. 

------


isabelleh (dismissed):
Nice!

adahn6 (approved):

PR #740, opened 2021-02-23
Minor fix to sectioning logic

adahn6:
### Description:

In certain cases for Edges ending in odd geometry (specifically, repeated points at the end of the line), sectioning was dropping geometry. This fix ensures that is no longer the case.

### Potential Impact:

Extremely limited, these geometries are rare-- but they now section as expected.

### Unit Test Approach:

A previous unit test covered this case, but thought the behavior should be to fail sectioning the line. This unit test has been updated to reflect the now correct behavior.

### Test Results:

Unit test passes.

------


lucaspcram (approved):
Great test!

PR #743, opened 2021-02-25
Keep information for OSC Change files (specifically, keep all OSM Nodes)

tsmock:
### Description:

Add an option (`-keepAll` set in osmlab/atlas-generator) to be used when generating an Atlas file for QC/QA purposes. This prevents the removal of non-tagged nodes.

### Potential Impact:

In order to create geometry changes for MapRoulette, osmlab/atlas-checks needs to know the OSM ids of the nodes of the way.

The atlas files _will_ be larger.

### Unit Test Approach:

N/A

### Test Results:

This does increase file sizes (files generated with https://github.com/osmlab/atlas-generator/pull/179 ).

| File | Original Size | New Size |
|-----|-------------|-----------|
| USA_10-202-390.atlas | 2 MB | 6.2 MB |
| USA_10-202-391.atlas | 1 MB | 3.0 MB |
| USA_10-203-390.atlas | 2.1 MB | 6.5 MB |
| USA_10-203-390.atlas | 5.5 MB | 14 MB |


adahn6:
overall this looks fine to me. the code smells would be good to cleanup, though, and it would be good to have some tests to cover this option

jwpgage (approved):

PR #748, opened 2021-03-16
First round of refactors to use new AtlasShellToolsCommandTemplate

lucaspcram:
### Description:
Some refactors to make use of the new template feature. This PR is not the complete refactor. A followup will address the now-deprecated template classes and fully convert the rest of the commands. For now, this PR should serve as a model for any downstream commands which would like to refactor to clean up their class hierarchies.

### Potential Impact:
Should be none, other than some template classes now deprecated.

### Unit Test Approach:
Added a new template test.

### Test Results:
Test suite 👍

------


src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/SubAtlasCommand.java
```diff
@@ -137,26 +120,26 @@ protected void processAtlas(final Atlas atlas, final String atlasFileName,
 
         if (this.optionAndArgumentDelegate.hasOption(SLICE_FIRST_OPTION_LONG))
         {
-            if (this.polygon.isPresent())
+            if (this.polygon != null)
```
matthieun, line 123: That seems odd, is it expected?

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/templates/MultipleOutputCommand.java
```diff
@@ -16,7 +16,9 @@
  * registers a '--output' option, the argument to which will be used as an output directory.
  *
  * @author lcram
+ * @deprecated
```
matthieun, line 19: Maybe say here what to use instead?

config/checkstyle/suppressions.xml
```diff
@@ -13,6 +13,7 @@
     <suppress checks="AvoidEscapedUnicodeCharacters" files="src/main/java/org/openstreetmap/atlas/utilities/command/terminal/TTYAttribute.java" />
     <suppress checks="ParameterName" files="src/main/java/org/openstreetmap/atlas/utilities/function/TernaryFunction.java" />
     <suppress checks="ParameterName" files="src/main/java/org/openstreetmap/atlas/utilities/function/TernaryOperator.java" />
+    <suppress checks="ParameterName" files="src/main/java/org/openstreetmap/atlas/utilities/function/TernaryConsumer.java" />
```
matthieun, line 16: As I was commenting about checkstyle and the parameter names, this appeared hahaha

matthieun: Maybe we should just call them `one` `two` and `three` and not have to do that every time?

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryConsumer.java
```diff
@@ -0,0 +1,58 @@
...
+     *             if {@code after} is null
+     */
+    default TernaryConsumer<T, U, V> andThen(
+            final TernaryConsumer<? super T, ? super U, ? super V> after)
+    {
+        Objects.requireNonNull(after);
+        return (l, c, r) ->
+        {
+            accept(l, c, r);
+            after.accept(l, c, r);
```
matthieun, line 55: How far until there is a stack overflow? 😺

matthieun (changes requested):
Nice work! Just a few minor comments and questions!

src/main/java/org/openstreetmap/atlas/utilities/function/TernaryConsumer.java
lucaspcram: Far enough that the Java standard library doesn't mind implementing it this way either :P

You can look at their `BiConsumer` for an example.

lucaspcram:
Nice comments thanks @matthieun I will take a look and address them soon!

src/main/java/org/openstreetmap/atlas/utilities/command/subcommands/SubAtlasCommand.java
lucaspcram: It is. Before, I was storing this instance variable as an `Optional` which is bad style. I changed it to a plain polygon, so now it needs a null check.

config/checkstyle/suppressions.xml
matthieun: Wouls still be nice to get rid of the suppressions, though..

matthieun (approved):

PR #749, opened 2021-03-23
boundary-itemizer command to group output per country

matthieun:
### Description:

`boundary-itemizer` is an atlas command which takes a country boundary file and writes it to a `wkt` and a `geojson` folder. In each folder, it used to write each outer polygon in a separate file.
With this update, it writes all geometries for a country in a single file, which makes it easier to use.

### Potential Impact:

This is an atlas command, so the impact is easier debugging

### Unit Test Approach:

Updated existing unit test

### Test Results:

Tested on large country boundary files

------


lucaspcram (approved):
This is the most beautiful sonar report ever.

lucaspcram (changes requested):
You can upgrade this by making use of the new `CountryBoundaryMapTemplate` to clean up and standardize the boundary reading code. See `AnyToGeoJsonCommand` for an example. You can read the map like this:

```java
        CountryBoundaryMap map = null;
        final Optional<CountryBoundaryMap> mapOptional = CountryBoundaryMapTemplate
                .getCountryBoundaryMap(this);
        if (mapOptional.isEmpty())
        {
            this.outputDelegate.printlnErrorMessage("failed to load country boundary");
            return 1;
        }
        map = mapOptional.get();
```

lucaspcram:
```java
    @Override
    public void registerManualPageSections()
    {
        .
        .
        registerManualPageSectionsFromTemplate(new CountryBoundaryMapTemplate());
        .
    }

    @Override
    public void registerOptionsAndArguments()
    {
        .
        registerOptionsAndArgumentsFromTemplate(new CountryBoundaryMapTemplate());
        .
        super.registerOptionsAndArguments();
    }
```

lucaspcram (approved):

PR #751, opened 2021-04-06
atlas/geography/Segment: Initialize list with size 2

tsmock:
This significantly decreases the cost for creating segements (a good
    chunk of the cost is from growing the ArrayList)

Signed-off-by: Taylor Smock <tsmock@fb.com>

### Description:

Initialize an arraylist in `atlas/geography/Segment#asList` with the expected number of objects.

### Potential Impact:

Cheaper segment initialization

### Unit Test Approach:

Already covered

### Test Results:
Original:
image

After patch: (run not yet finished)


matthieun (approved):
And 2 is not even a magic number! Thanks @tsmock!

PR #752, opened 2021-04-13
Allocation Perf fixes

tsmock:
### Description:

Various changes to improve memory allocation and performance from async profiler (allocation).

### Potential Impact:

Better performance.

### Unit Test Approach:

N/A -- note that there is a failing test from PackedTagStore: Avoid some HashMap grow calls (https://github.com/osmlab/atlas/commit/62f3b69ba480154fce970bb1b5d9a7ddd8800d62). It appears to just be an ordering issue. I'm on the fence as to whether or not I should just update the test, or not.

### Test Results:

------


matthieun (commented):
This all looks good to me, thanks for the improvements!
I am curious about this test failure though. What ordering is wrong here?

tsmock:
> This all looks good to me, thanks for the improvements!
> I am curious about this test failure though. What ordering is wrong here?

OK. In `BareAtlasTest#testSaveAsLineDelimitedGeoJsonFeatures`, the difference is:
`[...same text...]{"fixme":"please","highway":"trunk",[...same text...]`
`[...same text...]{"highway":"trunk","fixme":"please",[...same text...]`
So, changing how the `HashMap` was implemented changed how keys were ordered in the output.

matthieun:
I mean, as far as I know we should not care about the order... So it seems ok for me to just update the test.

tsmock:
Fair enough. I guess if it breaks stuff downstream, we can just say "It was undefined behavior". TBH, it will probably be tests that are affected, and nothing else.

I'll go ahead and make the change, and squash everything into one commit.

tsmock:
@matthieun : For https://github.com/osmlab/atlas/blob/dev/src/main/java/org/openstreetmap/atlas/utilities/collections/Iterables.java#L93, do you remember why that method exists, instead of reusing `Arrays.asList`? About the only advantage I can see for it is that it _doesn't_ use the originating array as a backing array.

tsmock:
Keep alive ping

adahn6:
this looks good to me

adahn6 (approved):

PR #754, opened 2021-04-23
Add consolidate logic

adahn6:
### Description:

In certain cases, sliced relations may only generate unintentional small pieces-- this issue can be due to imprecise coordinates, unusual feature definition, or simply very close alignment of borders and features. This PR adds in logic to allow flexible definitions for relations that should be "consolidated" to a single slice, which effectively trims the relation's geometry to the boundary for one country and no others. There are two criteria for consolidation. First is the tagging criteria, which is now stored in the same relation slicing filter as an additional filter; it defaults to `type->boundary` relations as these frequently follow boundaries used for slicing very closely and are most likely to generate these unintentional pieces. Second is a percentage threshold, which will only allow a relation to be "consolidated" if its largest piece is greater than 80% of the original geometry's size. Between these two conditions, consolidation should be relatively rare as well as configurable to prevent or enable specific other cases as desired.

### Potential Impact:

These relations are relatively rare, but will still impact anyone using the default filters. If this behavior isn't desired, the filter can be set to disallow all relations. atlas-generator will need to be updated as well.

### Unit Test Approach:

Two unit tests, checking that tagged relations are consolidated or not based on sizing criteria. As all other relations in unit tests won't meet the tagging criteria, those unit test assertions being unchanged covers the cases of relations not meeting the tagging criteria. 
### Test Results:

💯

------


isabelleh (approved):
What kind of changes do you expect with the shift from isMultiPolygon to isGeometric that I see in a few places?

adahn6:
@isabelleh good question, I don't have a specific reason per se. but we have a lot of relations that we expect geometries for that aren't `type->multipolygon`, so I was hoping avoid the clash in expectations behind an API named `isMultipolygon` which kind of implies that tag would be present with that value. I think I rewrote all of the internal Atlas uses for this method, so anything downstream that's using other APIs like the relation to multipolygon converters won't notice a difference. but I guess anything downstream that uses this will need to update to preserve expected behavior.

lucaspcram (approved):

PR #761, opened 2021-06-30
Do not remove nodes that are synthetic nodes

atiannicelli:
### Description:

This is a fix to not remove nodes that are marked as synthetic nodes when slicing

### Potential Impact:

This will remove fewer Atlas nodes from the atlas file generation and allow the generator to complete a full global execution.

### Unit Test Approach:

I was able to reproduce the issue locally and once this change was patched the crash was not seen. 

### Test Results:

Going to perform a full global execution now... but this should have no adverse affect.


jwpgage:
Thanks for the change Alex. However the build is failing due to spotless violations. Can you run ./gradlew spotlessApply and push?

atiannicelli:
@jwpgage - I should be thanking you for the change!! Thanks!

jwpgage (approved):

adahn6 (approved):

PR #764, opened 2021-08-05
DNM: Enable relation geometries

adahn6:
### Description:

This PR adds in a powerful new feature for certain Relations, in which MultiPolygon geometries are stored and replayed rather than constructed from member geometries on demand. In order to support this, a number of important changes have been made:

- PackedAtlasesSerializer (MUST BE PROTOBUF FORMAT) will write Relation MultiPolygon geometries to a new underlying file array
- PackedAtlas will now deserialize those geometries as needed, as well as maintaining a cached array of MultiPolygons that have been deserialized for faster replay
- Relations have a new method, `.asMultiPolygon()` that will return an Optional with their MultiPolygon geometry
- CompleteRelation is the primary way of adding geometry to a Relation, by using the new `.withMultiPolygonGeometry()` method
- Additionally, should a Relation have an existing stored geometry and updates are made to either its member list of the geometry of underlying members, CompleteRelation and ChangeRelation will calculate updates to ensure these changes are reflected in the stored geometry!
- Slicing has been updated to better reflect this, as it now updates the stored geometries for Relations rather than calculating synthetic members for later reconstruction. Note that as a caveat, a sliced Relation will have extra geometry that is not reflected by any member.
- ComplexEntity finders will now use the stored geometry

Note that across the code, the expectation has been made that *wherever* a Relation geometry could be used, the new API should be used and *not* assembled on demand. While a convenience method `.asMultiPolygon(true)` will assemble the geometry from members if needed, this method is NOT intended for serious use. Rather, the goal should be to ensure any and all Relations that need to have geometries are correctly generated and stored *BEFORE* any process attempt to use them. As an example, slicing an Atlas now assumes these geometries are stored-- it does not assemble. Similarly, ComplexEntity finders will assume the geometries are stored. This assumption is not made lightly, but is instead because of the *SIGNIFICANT* performance improvements from doing so. Assembling from members should now be considered a last resort done in exceptional circumstances, and immediate effort should be made to migrate all processes that might need it to using stored geometries instead.

For behavior examples, there are new unit tests in ChangeAtlasTest demonstrating stored geometry modification, AtlasDiffTest showing diffs between relation geometries, PackedAtlasSerializerTest showing the serialization, various ComplexEntity tests showing geometry finding, and RelationMultiPolygonGeometryTest showing general use cases.

### Potential Impact:

A lot for atlas-generator.... a followup PR will introduce the ability to create and store relation geometries during atlas generation. Otherwise, a lot of forward facing APIs are just using this behind the scene and involve no direct code changes.

### Unit Test Approach:

Described above!

### Test Results:

Lots of builds. Lots. Of. Builds. Things look good right now, but subject to change.

------


dependencies.gradle
```diff
@@ -8,7 +8,7 @@ project.ext.versions = [
     opencsv: '2.3',
     gson: '2.2.4',
     http: '4.5.1',
-    jts: '1.18.0',
+    jts: '1.18.2-SNAPSHOT',
```
matthieun, line 11: Will this be updated to a release when ready to merge?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeBuilder.java
```diff
@@ -53,16 +53,14 @@ public synchronized ChangeBuilder add(final FeatureChange featureChange)
     }
 
     /**
-     * Iteratively {@link #add(FeatureChange)} all the FeatureChanges.
-     *
+     * @see #addAll(Stream)
```
matthieun, line 56: It seems this is just re-ordering.. strange!

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
```diff
@@ -64,6 +83,97 @@ public RelationMemberList allKnownOsmMembers()
                         .collect(Collectors.toList());
     }
 
+    @Override
+    public Optional<MultiPolygon> asMultiPolygon()
+    {
+        if (!this.isGeometric())
+        {
+            return Optional.ofNullable(null);
```
matthieun, line 91:
```suggestion
            return Optional.empty();
```

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
```diff
@@ -64,6 +83,97 @@ public RelationMemberList allKnownOsmMembers()
...
+                    final Set<LineString> removed = removedMembers();
+                    final Set<LineString> added = addedMembers();
+
+                    if (removed.isEmpty() && added.isEmpty())
+                    {
+                        return sourceJtsGeometry;
+                    }
+
+                    Geometry updatedGeometry = convertMultiPolygonToLineCollection(sourceGeom);
+
```
matthieun, line 122: It'd be nice to have short comments along this method!

src/main/java/org/openstreetmap/atlas/geography/atlas/change/ChangeRelation.java
```diff
@@ -130,6 +263,23 @@ public Long osmRelationIdentifier()
         return ChangeEntity.getAttributeOrBackup(this.source, this.override, memberExtractor, name);
     }
 
+    private GeometryCollection convertMultiPolygonToLineCollection(final MultiPolygon multipolygon)
```
matthieun, line 266: Maybe move this to a new JTS converter?

src/main/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeValidator.java
```diff
@@ -50,6 +58,193 @@ protected void validateChangeNotEmpty()
         }
     }
 
+    protected void validateGeometricRelationsUpdated()
+    {
+        this.change.changesFor(ItemType.EDGE)
```
matthieun, line 63: Woooooo maybe break this one in more methods?

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -78,85 +85,89 @@
     protected static final String FIELD_SERIALIZER = "serializer";
     protected static final String FIELD_SAVE_SERIALIZATION_FORMAT = "saveSerializationFormat";
     protected static final String FIELD_LOAD_SERIALIZATION_FORMAT = "loadSerializationFormat";
+    protected static final String FIELD_CONTAINS_ENHANCED_RELATION_GEOMETRY = "containsEnhancedRelationGeometry";
     protected static final String FIELD_META_DATA = "metaData";
     protected static final String FIELD_DICTIONARY = "dictionary";
-    protected static final Object FIELD_DICTIONARY_LOCK = new Object();
+    private final transient Object fieldDictionaryLock = new Object();
```
matthieun, line 91: I am curious about the reasoning behind those changes!

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
```diff
@@ -227,6 +239,8 @@
     private final LongToLongMultiMap relationIndexToRelationIndices;
     private final LongToLongMultiMap relationOsmIdentifierToRelationIdentifiers;
     private final LongArray relationOsmIdentifiers;
+    private ByteArrayOfArrays relationGeometries;
```
matthieun, line 242: I would call this one `relationIndexToRelationGeometry`

matthieun: If that is what it is actually...

matthieun: So to make this consistent, I would make this some `MultipolygonArray` (which would need to be created) on the model of `PolygonArray` for areas.
Then I would get rid of `builtRelationGeometries` below, and use the array directly in accessor methods like
```
protected MultiPolygon relationGeometry(final long index)
```
below.

matthieun (changes requested):
This makes you think about the complexity behind a mature project like that!
Good testing!
Just a few nits here and there, and also a suggestion on how to store the data in packed atlas to keep it consistent. I am curious to see if you actually chose a `Map` for performance, or if it is by default.
Thanks @adahn6 and @lucaspcram!

matthieun:
> Lots. Of. Builds.

🤣

dependencies.gradle
adahn6: No. It's a significant step back in performance (as in, on a scale of over 400x slower). It doesn't change enough for this snapshot to be a concern.

src/main/java/org/openstreetmap/atlas/geography/atlas/change/validators/ChangeValidator.java
adahn6: yeah this is on the todo list lol

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
adahn6: when static objects are used as locks, any threads that try to deserialize use the same monitors even if they're deserializing different Atlases. that is ... not the intended behavior of these monitors, and has serious performance impact.

src/main/java/org/openstreetmap/atlas/geography/atlas/packed/PackedAtlas.java
adahn6: Hmm, I actually think this is a bad idea. Currently, we're getting a lot of mileage out of the fact that a WKT for a multipolygon is pretty small once compressed, and it has all the associated mappings of outers to inners already built in. We can't use the PrimitivePointsArray like PolygonArray does easily, because to do that we'd have to do the work to serialize holes ourselves, which would mean an array of outers which are themselves arrays as well, so it would be an array of arrays of arrays..... when in this case a single 2d array works quite well. I'm willing to hear the argument that it's necessary, but performance and size are quite capable as is so I'm not convinced upfront.

lucaspcram: I am inclined to agree with Sam - from a serialization perspective it's much clearer and simpler to just store the WKT. The array loads really quickly and it also means we can build the polygon on demand. Otherwise we'd have to spend a bunch of time designing some kind of proto object and dealing with all that nonsense.

dependencies.gradle
adahn6: looks like JTS is doing a release in a few weeks, so this won't be a huge deal either way. it'll either release in time for this PR to get updated, or we'll release with snapshot and upgrade ASAP shortly after.

lucaspcram (approved):
Looks good so far! Still going through to see if I have any more feedback.

PR #767, opened 2021-09-09
Exclude long straight route as junction route

Huyuntj:
### Description:

 Current merging logic merges small big nodes into large ones  and we found that for some cases, after merging, merged junction route could be a very long and straight which for most cases are invalid. In this PR, we exclude them as valid junction route.
Also we support configurable parameter map to define "long" and "straight" per countries.

### Potential Impact:

This will change what big node finds results

### Unit Test Approach:

Unit Test

### Test Results:

Passed

------


src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/bignode/BigNodeFinder.java
```diff
@@ -440,6 +462,53 @@ protected boolean hasJunctionEdgeTags(final Edge edge)
         return true;
     }
 
+    // Check if merged junction candidate is a very long straight route which is normally not a
+    // valid junction route
+    protected boolean isStraightLongRoute(final Route mergedCandidate)
+    {
+        if (mergedCandidate.size() < 2)
```
lucaspcram, line 469: Maybe a variable name or comment explanation for 2 here?

lucaspcram (changes requested):
Looks great! Nice tests. Just one suggestion.

lucaspcram (approved):
Looks great!

src/main/java/org/openstreetmap/atlas/geography/atlas/items/complex/bignode/BigNodeFinder.java
```diff
@@ -440,6 +461,55 @@ protected boolean hasJunctionEdgeTags(final Edge edge)
...
+                        && nextEdge.getTag(NameTag.KEY).isPresent()
+                        && !currentEdge.getTag(NameTag.KEY).get()
+                                .equalsIgnoreCase(nextEdge.getTag(NameTag.KEY).get()))
+                {
+                    return false;
+                }
+                currentEdge = nextEdge;
+            }
+            logger.info("Invalid Merged Candidate Route length is {} with WKT {}",
+                    mergedCandidate.length(), mergedCandidate.asPolyLine().toWkt());
```
lucaspcram, line 504:
```suggestion
        if (logger.isInfoEnabled())
        {
            logger.info("Invalid Merged Candidate Route length is {} with WKT {}",
                    mergedCandidate.length(), mergedCandidate.asPolyLine().toWkt());
        }
```

lucaspcram: This might help performance. Prevents computations from occurring unnecessarily when the log level is set higher.

lucaspcram (changes requested):
One more!

lucaspcram (approved):

PR #777, opened 2021-11-01
Generation: Keep tags throughout the generation process

tsmock:
### Description:

Keep atlas tags throughout the generation process.
* Also, added tag ("keepAll") so that end users know if all nodes are
  present in an atlas file.

### Potential Impact:

This will allow downstream users to know how an atlas file was generated, if necessary.
A specific example is the OrphanNodeCheck in atlas-checks, which looks for nodes without tags that are not part of relations. When keeping all nodes in the atlas file, this check breaks due to the following reasons:
1. It was looking for nodes that were not dropped during generation, had no tags, and were not part of a relation.
2. With the -keepAll flag, this check flags most nodes (for example, nodes that are part of ways or may be part of ways but have no tags).

### Unit Test Approach:

Add test that ensures that specific stages of pbf -> atlas generation does not drop tags.

### Test Results:

N/A


adahn6 (dismissed):
lgtm

adahn6 (approved):

PR #784, opened 2022-03-16
Update PackedAtlas to load all fields into memory before serialization

adahn6:
### Description:

This PR is a straightforward change to the PackedAtlas serialization implementation that guarantees that all fields have been loaded into memory before the Java serialization happens-- otherwise, we risk serializing an incomplete version of the object.

### Potential Impact:

Limited, this is a minor change as Java serialization is uncommon for Atlas.

### Test Results:

------


lucaspcram (approved):
Good catch on this one. This could trip up code which expects Java serialization of atlas to work normally.

PR #798, opened 2023-10-26
Code refactored for atlas start and end route paths

raju-gatla:
### Description:
Generating invalid routes for complex bigNodes
image

### Solution
Process if we have not yet visited the path/edge for start and end in any direction (It would be really weired to revisit an edge/path in a BigNode, both in forward and reverse directions.

### Potential Impact:
Should able to generate valid start to end routes

### Unit Test Approach:
Added unit tests

### Test Results:
Ran locally using tests which is generating expected routes

------

### DIFF
--- a/src/main/java/org/openstreetmap/atlas/geography/atlas/routing/AllPathsRouter.java
+++ b/src/main/java/org/openstreetmap/atlas/geography/atlas/routing/AllPathsRouter.java
@@ -193,7 +193,7 @@ public final class AllPathsRouter
         }
         // Add start edge to the path
         path.push(start);
-        //This will avoid adding same edge both in forward and reverse direction
+        // This will avoid adding same edge both in forward and reverse direction
         onPath.add(start.end().getIdentifier());
 
         if (start.equals(end))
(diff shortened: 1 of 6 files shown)


src/main/java/org/openstreetmap/atlas/geography/atlas/routing/AllPathsRouter.java
```diff
@@ -193,7 +193,7 @@ private static void allRoutes(final Edge start, final Edge end, final Stack<Edge
         }
         // Add start edge to the path
         path.push(start);
-        onPath.add(start.getIdentifier());
```
varaprasad6700, line 196: Can you add a comment on this method saying that this won't be giving routes which will have a set of same positive and negative edge

varaprasad6700, line 196: Can you add a comment on this method saying that this won't be giving routes which will have a set of same positive and negative edge

varaprasad6700 (changes requested):

src/main/java/org/openstreetmap/atlas/geography/atlas/routing/AllPathsRouter.java
raju-gatla: Added comment

varaprasad6700 (approved):
LGTM

saveerreddy (approved):
Every thing looks good to me. Thank You

Huyuntj:
Changes look good to me. @matthieun pls merge it.

Huyuntj (approved):

